Issue #19, opened 2020-05-16
addtional scrape config not updated for vmagent


f41gh7:
AdditionalScrape cm not updated after value change.

 Solution: add watching + ownership for it to VmAgent.

PR #21, opened 2020-05-24
added watch and ownership for vmagent additional scrape config

f41gh7:
fixes to bug #19 
 Now changes to additional scrape config secret will trigger reconcile loop for vmagent.


pkg/controller/factory/vmagent.go
```diff
@@ -485,3 +488,39 @@ func selectorLabelsVmAgent(cr *victoriametricsv1beta1.VmAgent) map[string]string
 	return labels
 
 }
+
+//add ownership - it needs for object changing tracking
+func addAddtionalScrapeConfigOwnership(cr *victoriametricsv1beta1.VmAgent, rclient client.Client, l logr.Logger) error {
+	if cr.Spec.AdditionalScrapeConfigs != nil {
```
tenmozes: one of the go codestyle paradigm is to "return fast", so please change to 

`if cr.Spec.AdditionalScrapeConfigs != nil {
  return nil
}
`

pkg/controller/factory/vmagent.go
```diff
@@ -485,3 +488,39 @@ func selectorLabelsVmAgent(cr *victoriametricsv1beta1.VmAgent) map[string]string
...
+		secret := &corev1.Secret{}
+		err := rclient.Get(context.TODO(), types.NamespacedName{Namespace: cr.Namespace, Name: cr.Spec.AdditionalScrapeConfigs.Name}, secret)
```
tenmozes: don't need to use TODO context it just doesn't work, please pass it to function input or use context.Background

pkg/controller/factory/vmagent.go
```diff
@@ -485,3 +488,39 @@ func selectorLabelsVmAgent(cr *victoriametricsv1beta1.VmAgent) map[string]string
...
+		if err != nil {
+			l.Error(err, "cannot find secret with scrape config", "secret", cr.Spec.AdditionalScrapeConfigs.Name)
```
tenmozes: "secret <name> not found"

pkg/controller/factory/vmagent.go
f41gh7: changed with return fast.

pkg/controller/factory/vmagent.go
f41gh7: added background context.

pkg/controller/factory/vmagent.go
f41gh7: changed

PR #24, opened 2020-06-02
add basic vmcluster with vmstorage

tenmozes:


pkg/apis/victoriametrics/v1beta1/vmcluster_types.go
```diff
@@ -0,0 +1,354 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []VmCluster `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&VmCluster{}, &VmClusterList{})
+}
+
+type VmSelect struct {
+	Enabled bool `json:"enabled"`
```
f41gh7: Enabled is not a common pattern for operators.  Usually, it implicitly adds default values.  I think it`s better to replace with 
```
VmSelect  *VmSelect  `json:"vmselect,omitempty"`
```

and check for nil at reconcile.

pkg/apis/victoriametrics/v1beta1/vmcluster_types.go
```diff
@@ -0,0 +1,354 @@
...
+	StorageDefaultVolumeName = "vmstorage-volume"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// VmClusterSpec defines the desired state of VmCluster
+type VmClusterSpec struct {
+	VmSelect  VmSelect  `json:"vmselect"`
+	VmInsert  VmInsert  `json:"vminsert"`
```
f41gh7: can you add the tag omitempty for all non-required fields?

pkg/apis/victoriametrics/v1beta1/vmcluster_types.go
```diff
@@ -0,0 +1,354 @@
...
+	Name string `json:"name"`
+	// +optional
+	Image Image `json:"image"`
+	// +optional
+	PriorityClassName string `json:"priorityClassName"`
+	// +optional
+	FullnameOverride string `json:"fullnameOverride"`
+	// +optional
+	ExtraArgs map[string]string `json:"extraArgs"`
+	// +optional
```
f41gh7, line 203: + extraVars
[]v1.EnvVar
https://github.com/VictoriaMetrics/operator/blob/master/pkg/apis/victoriametrics/v1beta1/vmagent_types.go#L196

pkg/apis/victoriametrics/v1beta1/vmcluster_types.go
```diff
@@ -0,0 +1,354 @@
...
+	Annotations map[string]string `json:"annotations"`
+	// +optional
+	Tolerations v1.Toleration `json:"tolerations"`
+	// +optional
+	NodeSelector v1.NodeSelector `json:"nodeSelector"`
+	// +optional
+	Affinity v1.Affinity `json:"affinity"`
+	// +optional
+	PodAnnotations map[string]string `json:"podAnnotations"`
```
f41gh7: it can be replaced with podMetadata
https://github.com/VictoriaMetrics/operator/blob/master/pkg/apis/victoriametrics/v1beta1/vmagent_types.go#L17

pkg/apis/victoriametrics/v1beta1/vmcluster_types.go
```diff
@@ -0,0 +1,354 @@
...
+	ReplicaCount   int               `json:"replicaCount"`
```
f41gh7: + for unify,  other objects have Replicas field *int32

pkg/apis/victoriametrics/v1beta1/vmcluster_types.go
```diff
@@ -0,0 +1,354 @@
...
+	// +optional
+	SecurityContext v1.PodSecurityContext `json:"securityContext"`
+	// +optional
+	CacheMountPath string `json:"cacheMountPath"`
+	// +optional
+	Service VmSelectService `json:"service"`
+	// +optional
+	StatefulSet VmSelectSts `json:"statefulSet"`
+	// +optional
+	PersistentVolume VmSelectPV `json:"persistentVolume"`
```
f41gh7: Should we use storage spec for it ?  It provides some nice options with empty dir.

https://github.com/VictoriaMetrics/operator/blob/master/pkg/apis/victoriametrics/v1beta1/alertmanager_types.go#L85

pkg/controller/vmcluster/storage_controller.go
```diff
@@ -0,0 +1,289 @@
...
+	if c.cluster.Status.VmStorage.Status != status {
+		c.cluster.Status.VmStorage.Status = status
+		log.Info("update storage status", "status", status)
+		if err := c.client.Status().Update(ctx, c.cluster); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+	if status == victoriametricsv1beta1.VmStorageStatusExpanding {
+		log.Info("cluster still expanding requeue request")
+		return reconcile.Result{Requeue: true}, nil
```
f41gh7, line 39: Requeue after can be added for some delay, is it useful? 
 		return reconcile.Result{Requeue: true, RequeueAfter: time.Second *10 }, nil

deploy/examples/vmcluster_basic.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: victoriametrics.com/v1beta1
+kind: VmCluster
+metadata:
+  name: example-vmcluster
+spec:
+  vmselect:
+    enabled: true
+    name: vmselect
+    image:
+      repository: victoriametrics/vmselect
```
f41gh7, line 10: + for default values, I suggest using predefined values at conf package
https://github.com/VictoriaMetrics/operator/blob/master/conf/config.go#L18 
and inject it for missing values at deploy/sts generation part
https://github.com/VictoriaMetrics/operator/blob/master/pkg/controller/factory/vmsingle.go#L138

pkg/controller/vmcluster/storage_controller.go
```diff
@@ -0,0 +1,289 @@
...
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{
+				{
+					Name:       "http",
+					Protocol:   corev1.ProtocolTCP,
+					Port:       storageSpec.Service.ServicePort,
+					TargetPort: intstr.FromString("http"),
+				},
+				{
+					Name:       "vmselect",
```
f41gh7, line 258: It`s storage service, does it requires vmselect and vminsert targets?

pkg/controller/vmcluster/storage_controller.go
```diff
@@ -0,0 +1,289 @@
...
+		TCPSocket: &corev1.TCPSocketAction{
+			Port: intstr.FromString("http"),
+		},
+	}
+	args := []string{
+		fmt.Sprintf("--retentionPeriod=%d", storageSpec.RetentionPeriod),
+		fmt.Sprintf("--storageDataPath=%s", storageSpec.PersistentVolume.MountPath),
+	}
+	for arg, value := range storageSpec.ExtraArgs {
+		args = append(args, fmt.Sprintf("--%s:%s", arg, value))
```
f41gh7, line 117: should it be
```
"--%s=%s", arg, value)
```
?

pkg/apis/victoriametrics/v1beta1/vmcluster_types.go
f41gh7: changed at vmAlert and vmAgent to replicaCount *int32

Issue #4, opened 2020-05-10
tests for operator


f41gh7:
Currently, there is no tests, need to add it. 

1) first of all -  rules/monitors config generation.

2) unit tests for reconcile logic

https://sdk.operatorframework.io/docs/golang/unit-testing/

3) e2e tests 
https://sdk.operatorframework.io/docs/golang/e2e-tests/

PR #29, opened 2020-06-07
added e2e tests, fixed panic, updated docs and rbac 

f41gh7:
#4 
1) fixed panic at vmSingle - changed get annotations method
2) added e2e tests - now it doesn't include at ci, need to fix the problem with Github actions.
3) RBAC moved to separate files:  role, role_binding and service_account files.

### DIFF
--- a/e2e/main_test.go
+++ b/e2e/main_test.go
@@ -1,14 +1,14 @@
 package e2e
 
 import (
-	"github.com/VictoriaMetrics/operator/pkg/apis"
-	operator "github.com/VictoriaMetrics/operator/pkg/apis/victoriametrics/v1beta1"
-	framework "github.com/operator-framework/operator-sdk/pkg/test"
-	"k8s.io/apimachinery/pkg/runtime"
 	"testing"
 	"time"
 
+	"github.com/VictoriaMetrics/operator/pkg/apis"
+	operator "github.com/VictoriaMetrics/operator/pkg/apis/victoriametrics/v1beta1"
 	f "github.com/operator-framework/operator-sdk/pkg/test"
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yaml
```diff
@@ -46,3 +48,22 @@ jobs:
           tag_with_ref: true
           dockerfile: cmd/manager/Dockerfile
 
+# TODO fix it, for some reason it fails with timeout
```
tenmozes, line 51: please remove commented code

e2e/main_test.go
```diff
@@ -0,0 +1,64 @@
+package e2e
+
+import (
+	"github.com/VictoriaMetrics/operator/pkg/apis"
```
tenmozes: please sort imports

e2e/main_test.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+
+func TestVmApps(t *testing.T) {
+
+	err := addToSchemeCrds(t)
+	if err != nil {
+		t.Fatalf("failted to add custom resource to scheme: %v", err)
+	}
+	t.Run("vm", func(t *testing.T) {
```
tenmozes: why do we need to wrap every run into parent "vm" run ?

e2e/vmalert_test.go
```diff
@@ -0,0 +1,82 @@
+package e2e
+
+import (
+	goctx "context"
+	"fmt"
+	operator "github.com/VictoriaMetrics/operator/pkg/apis/victoriametrics/v1beta1"
+	//mon "github.com/VictoriaMetrics/operator/pkg/apis/monitoring/v1"
```
tenmozes: please remove

.github/workflows/main.yaml
f41gh7: done

e2e/main_test.go
f41gh7: applied goimports

e2e/main_test.go
f41gh7: nope, thx, I`ve missed it. It`s at framework's example with parallel test exec.

I've removed wrap.

e2e/vmalert_test.go
f41gh7: done

Issue #26, opened 2020-06-06
load tlsConfig for serviceMonitors  isn't working


f41gh7:
Prometheus-operator provides ability to load tls config for target scrape from secrets:

https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/operator.go#L1519
```yaml
(c *Operator) loadTLSAssets(mons map[string]*monitoringv1.ServiceMonitor) (map[string]TLSAsset, error) ```

 We support the same api objects:

 https://github.com/VictoriaMetrics/operator/blob/master/pkg/apis/monitoring/v1/servicemonitor_types.go#L129

```yaml
type TLSConfig struct {
  ...
	// Stuct containing the CA cert to use for the targets.
	CA SecretOrConfigMap `json:"ca,omitempty"`
	Cert SecretOrConfigMap `json:"cert,omitempty"`

	// Secret containing the client key file for the targets.
	KeySecret *v1.SecretKeySelector `json:"keySecret,omitempty"`
...
}```

  It would be nice to have the same behaviour.

PR #31, opened 2020-06-10
Added tlsAssets load #26

f41gh7:
fixed problem with TLSConfig load from secret or configmap for serviceMonitors,

Added tests,
Changed vmAgent createOrUpdate behaviour,
fixed small typos

### DIFF
--- a/pkg/controller/factory/servicemons.go
+++ b/pkg/controller/factory/servicemons.go
@@ -410,13 +410,13 @@ func getCredFromSecret(
 	sel v1.SecretKeySelector,
 	cacheKey string,
 	cache map[string]*v1.Secret,
-) (_ string, err error) {
+) (string, error) {
 	var s *v1.Secret
 	var ok bool
 
 	if s, ok = cache[cacheKey]; !ok {
 		s = &v1.Secret{}
-		if err = rclient.Get(ctx, types.NamespacedName{Namespace: ns, Name: sel.Name}, s); err != nil {
+		if err := rclient.Get(ctx, types.NamespacedName{Namespace: ns, Name: sel.Name}, s); err != nil {
 			return "", fmt.Errorf("unable to fetch key from secret%s: %w", sel.Name, err)
 		}
 		cache[cacheKey] = s
@@ -431,7 +431,7 @@ func getCredFromConfigMap(
 	sel v1.ConfigMapKeySelector,
 	cacheKey string,
 	cache map[string]*v1.ConfigMap,
-) (_ string, err error) {
+) (string, error) {
 	var s *v1.ConfigMap
 	var ok bool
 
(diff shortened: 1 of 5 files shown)


pkg/controller/factory/servicemons.go
```diff
@@ -397,45 +396,70 @@ func loadBasicAuthSecret(basicAuth *monitoringv1.BasicAuth, s *v1.SecretList) (B
...
+}
+
+func getCredFromConfigMap(
+	ctx context.Context,
+	rclient client.Client,
+	ns string,
+	sel v1.ConfigMapKeySelector,
+	cacheKey string,
+	cache map[string]*v1.ConfigMap,
+) (_ string, err error) {
```
tenmozes: https://github.com/golang/go/wiki/CodeReviewComments#naked-returns could you please remove naked return

pkg/controller/factory/servicemons.go
```diff
@@ -397,45 +396,70 @@ func loadBasicAuthSecret(basicAuth *monitoringv1.BasicAuth, s *v1.SecretList) (B
...
+	var s *v1.ConfigMap
+	var ok bool
+
+	if s, ok = cache[cacheKey]; !ok {
```
tenmozes, line 438: looks like possitive path is faster can you change the logic and if key is in a cache return it

pkg/controller/factory/vmagent.go
```diff
@@ -453,3 +464,133 @@ func addAddtionalScrapeConfigOwnership(cr *victoriametricsv1beta1.VmAgent, rclie
...
+				continue
+			}
+
+			prefix := mon.Namespace + "/"
+			secretSelectors := map[string]*corev1.SecretKeySelector{}
+			configMapSelectors := map[string]*corev1.ConfigMapKeySelector{}
+			if ep.TLSConfig.CA != (monitoringv1.SecretOrConfigMap{}) {
+				switch {
+				case ep.TLSConfig.CA.Secret != nil:
+					secretSelectors[prefix+ep.TLSConfig.CA.Secret.Name+"/"+ep.TLSConfig.CA.Secret.Key] = ep.TLSConfig.CA.Secret
```
tenmozes: could you please intorduce a method - buildSecterName or smth similar?

pkg/controller/factory/servicemons.go
f41gh7: done

pkg/controller/factory/servicemons.go
f41gh7: Nope, you have to extract data from it anyway:
```yaml
if a, ok := s.Data[sel.Key]; ok {
   return a, nil
}
```
In a positive path, it would be required to do one more time.

pkg/controller/factory/vmagent.go
f41gh7: added builder method

Issue #37, opened 2020-06-15
The CustomResourceDefinition is invalid for 1.17 and 1.18 kubernetes versions


f41gh7:
There are two problems:
1) with 1.16 and 1.17 clusters
```bash
 The CustomResourceDefinition "vmagents.victoriametrics.com" is invalid:
* spec.validation.openAPIV3Schema.properties[spec].properties[containers].items.type: Invalid value: "object": must be a scalar or atomic type as item of a list with x-kubernetes-list-type=set
* spec.validation.openAPIV3Schema.properties[spec].properties[volumeMounts].items.type: Invalid value: "object": must be a scalar or atomic type as item of a list with x-kubernetes-list-type=set
* spec.validation.openAPIV3Schema.properties[spec].properties[remoteWrite].items.type: Invalid value: "object": must be a scalar or atomic type as item of a list with x-kubernetes-list-type=set
* spec.validation.openAPIV3Schema.properties[spec].properties[initContainers].items.type: Invalid value: "object": must be a scalar or atomic type as item of a list with x-kubernetes-list-type=set
* spec.validation.openAPIV3Schema.properties[spec].properties[imagePullSecrets].items.type: Invalid value: "object": must be a scalar or atomic type as item of a list with x-kubernetes-list-type=set
```
  listType=set must be removed to fix it.

2) with 1.18 cluster version
```bash
The CustomResourceDefinition "appservices.app.example.com" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
```

Kubernetes upstream problem.

 https://github.com/operator-framework/operator-sdk/issues/3235

 need to find out minimal version of Kubernetes, that operator supports.

Issue #36, opened 2020-06-15
RBAC: possibly missing "replicasets" resource in vm-operator ClusterRole


clmssz:
When starting the operator : 

```
vm-operator-85f8d8b599-xp5s8 vm-operator {"level":"info","ts":1592233312.4326746,"logger":"cmd","msg":"Could not create metrics Service","error":"failed to initialize service object for metrics: replicasets.apps \"vm-operator-85f8d8b599\" is forbidden: User \"system:serviceaccount:monitoring:vm-operator\" cannot get resource \"replicasets\" in API group \"apps\" in the namespace \"monitoring\""}
```

Fixed by adding `replicasets` in the ClusterRole definition of vm-operator (not sure if it was an oversight or voluntary to not include them in the first place)

Issue #35, opened 2020-06-15
Alertmanager CRD overlaps with the Prometheus Operator one


clmssz:
Hi there,

Not sure if the files in master are the most up to date ones, since you mention an `install` directory in the docs and it's not there yet, but in the `deploy/crds` for vmalertmanagers, singular and plural terms mismatch - If the goal is to use the coreos CRD for alertmanager, better remove it from CRDs here as it will conflict with an already deployed prometheus operator

```
  names:
    kind: Alertmanager
    listKind: AlertmanagerList
    plural: vmalertmanagers
    shortNames:
    - vma
    singular: alertmanager
```

Issue #30, opened 2020-06-07
improve e2e tests 


f41gh7:
There is method for checking deployment: 
```yaml
err = e2eutil.WaitForOperatorDeployment(t, f.KubeClient, namespace, "vm-operator", 1, retryInterval, timeout)```

And i sugest to add  few additional methods: 
1) WaitForConfigmap() - optional check keys name
2) WaitForStatefulSet  - validate status ready
3) WaitForSecret() - optional check keys name

PR #39, opened 2020-06-15
update recent bugs, fix kubernetes crd

f41gh7:
removed list-Type=set - fixed kubernetes crd - #37 
improved e2e tests tools - added new method WaitForSts WaitForConfigMap, WaitForSecret and WaitForService  - #30
added e2e test for alertmanager - #30
added replicaset RBAC for operator service - #36 
changed Alertmanager plular name to vmalertmanager - #35

### DIFF
--- a/e2e/utils.go
+++ b/e2e/utils.go
@@ -1,6 +1,8 @@
 package e2e
 
 import (
+	"errors"
+	"fmt"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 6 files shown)


cmd/manager/main.go
```diff
@@ -130,11 +130,17 @@ func main() {
 	if err != nil {
 		log.Info("Could not create metrics Service", "error", err.Error())
 	}
+	//retrieve operator namespace
```
tenmozes, line 133: please add an option to run sdk locally, as i know there is a method to define where code is running inside cluster or outside

e2e/utils.go
```diff
@@ -0,0 +1,99 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"testing"
+	"time"
+)
+
+func WaitForSts(t *testing.T, kubeclient kubernetes.Interface, namespace, name string, replicas int,
+	retryInterval, timeout time.Duration) error {
+	err := wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		sts, err := kubeclient.AppsV1().StatefulSets(namespace).Get(name, metav1.GetOptions{})
```
tenmozes: you can reduce complexity via
- move wait code to one function and accept function to get the item

f41gh7: added WaitForEntity

PR #44, opened 2020-07-13
elegant solution for k8s 1.18

k1rk:
only till https://github.com/kubernetes-sigs/controller-tools/issues/444 be resolved


Makefile
```diff
@@ -44,10 +45,19 @@ install-develop-tools: install-golint
...
 
-gen:
+gen-crd:
 	$(OPERATOR_BIN) generate crds --crd-version=v1beta1
 	$(OPERATOR_BIN) generate k8s
 
+fix118:
+	docker run --rm -v "${PWD}":/workdir mikefarah/yq /bin/sh -c " \
+		$(YAML_MERGE) deploy/crds/victoriametrics.com_vmagents_crd.yaml scripts/patch1.18.yaml && \
+		$(YAML_MERGE) deploy/crds/victoriametrics.com_vmalertmanagers_crd.yaml scripts/patch1.18.yaml && \
```
f41gh7: yq d -i deploy/crds/victoriametrics.com_vmagents_crd.yaml spec.validation.openAPIV3Schema.properties.spec.properties..containers.items.properties

@k1rk  what do you think? we can simply  remove properties for this crd

k1rk: added

k1rk: or actually removed :D

f41gh7: ah, i mean to apply it as general solution for all containers and initContainers defenitions :)

k1rk: 👌  done

PR #50, opened 2020-07-29
TLSConfig for remote targets #47

f41gh7:
added TLSConfig to VMAgent and VMAlert,
added unit and e2e test for new features
changed NotifierURL to Notifier with Spec
add serviceAccount

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -22,4 +22,3 @@ spec:
       - name: manager
         args:
         - "--metrics-addr=127.0.0.1:8080"
-        - "--enable-leader-election"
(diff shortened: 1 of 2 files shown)


controllers/factory/servicemons_build.go
```diff
@@ -668,21 +668,12 @@ func addTLStoYaml(cfg yaml.MapSlice, namespace string, tls *victoriametricsv1bet
 		if tls.CAFile != "" {
 			tlsConfig = append(tlsConfig, yaml.MapItem{Key: "ca_file", Value: tls.CAFile})
 		}
-		if tls.CA.Secret != nil {
```
f41gh7, line 671: bug, fix it

PR #60, opened 2020-08-17
Upgraded sdk and go version, changed folders structure.

f41gh7:


Dockerfile.build
```diff
@@ -0,0 +1,27 @@
+# Build the manager binary
+FROM golang:1.13 as builder
```
tenmozes: please update to 1.15

controllers/factory/alertmanager.go
```diff
@@ -7,7 +7,7 @@ import (
 	"path"
 
 	victoriametricsv1beta1 "github.com/VictoriaMetrics/operator/api/v1beta1"
-	"github.com/VictoriaMetrics/operator/conf"
+	"github.com/VictoriaMetrics/operator/internal/conf"
```
tenmozes: pleas rename conf to config

Issue #59, opened 2020-08-05
implement blackbox crd


k1rk:
implement blackbox crd, as in prom-operator
https://github.com/prometheus-operator/prometheus-operator/pull/2832

PR #63, opened 2020-08-18
Blackbox crd implementation #59

f41gh7:
- added new crd
- fixed minor issues with config generation
- added labels and annotations inherit for prometheus api converter


tenmozes (approved):
thank you for contribution

PR #73, opened 2020-08-25
Docs add kustomize instalation

### LINKED ISSUE #70: Add kustomization.yml example to the docs
There is support to avoid manual downloading of manifest simply by using Kustomize resources. All we need is to add following example to the documentation to allow others to quickly start deploying and modifying their operator.

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
- namespace.yml
- github.com/VictoriaMetrics/operator/config/default?ref=v0.2.0

namespace: monitoring
images:
- name: victoriametrics/operator
  newTag: v0.2.0
```  

Please let me know what do you think and I can send patch right away.

As bonus point it would be great if we could bake in version into git revision so one can skip image patching, but I would expect that as separate step as it would probably require some changes in release process.

elmariofredo:
Closes #70 

Signed-off-by: Mario Vejlupek <mario@vejlupek.cz>

### DIFF
--- a/docs/quick-start.MD
+++ b/docs/quick-start.MD
@@ -3,7 +3,7 @@
 
 Operator serves to make running Victoria Metrics applications on top of Kubernetes as easy as possible while preserving Kubernetes-native configuration options.
 
-## Installing operator
+## Installing by Manifest
 
 Obtain release from releases page:
 [https://github.com/VictoriaMetrics/operator/releases](https://github.com/VictoriaMetrics/operator/releases)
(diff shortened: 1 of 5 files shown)


docs/quick-start.MD
```diff
@@ -39,6 +39,69 @@ kubectl get pods
...
+    control-plane: vm-operator
+  name: vm-operator
+spec:
+  template:
+    spec:
+      containers:
+      - name: manager
+        env:
+        - name: WATCH_NAMESPACE
+          value: "app-prod"
```
f41gh7: There is a problem with WATCH_NAMESPACE, currently operator was tested only with WATCH_NAMESPACE="" aka all namespaces.

 So, i think, its better to introduce some other parameter, for instance:
```- name: VM_CLUSTERDOMAINNAME
      value: cluster.local
```

docs/quick-start.MD
```diff
@@ -39,6 +39,69 @@ kubectl get pods
 
 ```
 
+
+## Installing by Kustomize
+
+You can install operator using [Kustomize](https://kubernetes-sigs.github.io/kustomize/) by pointing to the remote kustomization file.
+
+```yaml
```
f41gh7, line 47: what do you think, is it better to use
cat << EOF >> file_name
...content
EOF
 
 Or just place desired filename to comment?

f41gh7 (changes requested):

docs/quick-start.MD
devopsmariocom: sure no problem

f41gh7 (approved):

f41gh7:
Thanks, looks awesome!

PR #79, opened 2020-08-30
Fix alertmanager example indention

elmariofredo:
See Alermanager docs for reference https://prometheus.io/docs/alerting/latest/configuration/#configuration-file


f41gh7:
thanks!

Issue #71, opened 2020-08-24
Add ARM support


elmariofredo:
I understand that it's early and not priority just want to raise my hand that my raspberry pi 4 64bit cluster is forced to use prometheus-operator. So please if it's not that big of a issue let's make it happen 🙏

f41gh7:
Hello, 

  Wow, pi cluster, it  sounds interesting)

  I guess, its very simple and i will add cross compilation into next release.

elmariofredo:
That would be fantastic 😍 let me know if you will need any testing or other support. Thanks!

f41gh7:
i've added cross compilation, you can test operator with docker image: `victoriametrics/operator:cross-compilation-arm64`.
 I've tested it at aws arm, it was able to start.

 Also, you have to have change values for it, apply patch: 
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    control-plane: vm-operator
  name: vm-operator
spec:
  template:
    spec:
      containers:
      - name: manager
        env:
        - name: VM_VMALERTDEFAULT_VERSION
          value: v1.40.0-arm64
        - name: VM_VMAGENTDEFAULT_VERSION
          value: v1.40.0-arm64
        - name: VM_VMSINGLEDEFAULT_VERSION
          value: v1.40.0-arm64
        - name: VM_VMCLUSTERDEFAULT_VMSELECTDEFAULT_VERSION
          value: v1.40.0-cluster-arm64
        - name: VM_VMCLUSTERDEFAULT_VMSTORAGEDEFAULT_VERSION
          value: v1.40.0-cluster-arm64
        - name: VM_VMCLUSTERDEFAULT_VMSELECTDEFAULT_VERSION
          value: v1.40.0-cluster-arm64
        - name: VM_VMALERTMANAGER_ALERTMANAGERVERSION
          value: v0.20.0-arm64
        - name: VM_VMALERTMANAGER_CONFIGRELOADERIMAGE
          value: jimmidyson/configmap-reload:v0.4.0-arm64
        - name: VM_VMALERTDEFAULT_CONFIGRELOADIMAGE
          value: jimmidyson/configmap-reload:v0.4.0-arm64
        - name: VM_VMAGENTDEFAULT_CONFIGRELOADIMAGE
          value: quay.io/coreos/prometheus-config-reloader:v0.41.1-arm64
 
```

 Let me know, if something go wrong.

elmariofredo:
[...]

PR #82, opened 2020-09-09
Add docker manifest

f41gh7:
It needs for multiarch builds.And it allows to run docker image without explicit platform specification.


tenmozes (approved):

f41gh7:
closes #71

PR #87, opened 2020-09-20
Create vmservicescrape-metrics-server.yml

elmariofredo:
Add example metric-server scrape

### DIFF
--- a/config/examples/vmservicescrape-metrics-server.yml
+++ b/config/examples/vmservicescrape-metrics-server.yml
@@ -1,3 +1,7 @@
+# Scrape Metrics Server
+# Cluster-wide aggregator of resource usage data. 
+# https://github.com/kubernetes-sigs/metrics-server
+---
 apiVersion: operator.victoriametrics.com/v1beta1
 kind: VMServiceScrape
 metadata:
@@ -11,7 +15,7 @@ spec:
     matchLabels:
       kubernetes.io/name: Metrics-server
   endpoints:
-  - port: https
+  - port: 443
     scheme: https
     tlsConfig:
       insecureSkipVerify: true
--- a/config/examples/vmservicescrape-metrics-server.yml
+++ b/config/examples/vmservicescrape-metrics-server.yml
@@ -6,10 +6,10 @@ apiVersion: operator.victoriametrics.com/v1beta1
 kind: VMServiceScrape
 metadata:
   name: metrics-server
-  namespace: kube-system
+  namespace: monitoring-system
 spec:
   namespaceSelector:
-    matchName:
+    matchNames:
     - kube-system
   selector:
     matchLabels:
--- a/config/examples/vmservicescrape-metrics-server.yml
+++ b/config/examples/vmservicescrape-metrics-server.yml
@@ -17,3 +17,33 @@ spec:
       insecureSkipVerify: true
       caFile: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
     bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: view-metrics
+rules:
+- apiGroups:
+    - metrics.k8s.io
+  resources:
+    - pods
+    - nodes
+  verbs:
+    - get
+    - list
+    - watch
+- nonResourceURLs: ["/metrics"]
+  verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: view-metrics
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: view-metrics
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: User
+    name: system:serviceaccount:monitoring-system:vmagent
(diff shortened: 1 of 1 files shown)


config/examples/vmservicescrape-metrics-server.yml
```diff
@@ -0,0 +1,49 @@
...
+  namespace: kube-system
+spec:
+  namespaceSelector:
+    matchName:
+    - kube-system
+  selector:
+    matchLabels:
+      kubernetes.io/name: Metrics-server
+  endpoints:
+  - port: https
```
f41gh7: I've tried latest metrics-server release, and there is no endpoint port name at their configuration. So i had to replace port with `targetPort: 443`. 

 Did you faced the same problem? 0.3.7 release of metrics-server.

 `kubectl -n kube-system get endpoints metrics-server -o yaml` returned:
```bash
apiVersion: v1
kind: Endpoints
metadata:
  creationTimestamp: "2020-09-12T23:31:26Z"
  labels:
    addonmanager.kubernetes.io/mode: Reconcile
    kubernetes.io/cluster-service: "true"
    kubernetes.io/name: Metrics-server
  name: metrics-server
  namespace: kube-system
  resourceVersion: "4011249"
  selfLink: /api/v1/namespaces/kube-system/endpoints/metrics-server
  uid: 3794d5c6-6d55-4118-8b6b-fa5ba4ea6066
subsets:
- addresses:
  - ip: 10.0.0.5
    nodeName: gke-cluster-1-pool-1-preem-d6699a17-jxmr
    targetRef:
      kind: Pod
      name: metrics-server-v0.3.6-547dc87f5f-4568r
      namespace: kube-system
      resourceVersion: "4000294"
      uid: acb418b1-5d4d-4619-9108-b4817aab6105
  ports:
  - port: 443
    protocol: TCP
```

 name for port is missing, probably they changed behaviour or i missed something.

f41gh7 (commented):
Thanks a lot!

  May i ask you to add link to metrics-server repository? It's useful for people, who not familiar with it. I guess, it can be placed to the comment line.

config/examples/vmservicescrape-metrics-server.yml
devopsmariocom: yeah this depends on distribution usually when you setup k8s with some tool like k3s, kubeadm is will deploy it for you. I will change it to 433 to have it explicit

devopsmariocom: Also I have created https://github.com/kubernetes-sigs/metrics-server/pull/585 so let's see we can revert back to https when this is fixed.

config/examples/vmservicescrape-metrics-server.yml
```diff
@@ -0,0 +1,53 @@
...
+---
+apiVersion: operator.victoriametrics.com/v1beta1
+kind: VMServiceScrape
+metadata:
+  name: metrics-server
+  namespace: kube-system
+spec:
+  namespaceSelector:
+    matchName:
+    - kube-system
```
devopsmariocom, line 13: @f41gh7 I'm not sure about this as changing this value doesn't seems to have any effect to resulting configuration, it's defined in CRD but is it really doing anything?

f41gh7: Well, there is typo - it must be `matchNames` instead of `matchName` :)

And in this case, yes, it does nothing. Because  namespace selector  was added implicitly, if selectors is not defined.
 There is complicated logic around it:
1) spec.ignoreNamespaceSelectors = true - crd namespace will be added.
2) if spec.namespaceSelector.any = true - empty selector, match all namespaces
3) if len(matchNames) == 0 - crd namespace will be added
4) otherwise use user defined list of namespaces.

devopsmariocom: Ah silly mistake 😄 thanks! I have fix it and place VMServiceScrape to the `monitoring-system`, so others have guides how to target other namespaces. It would be great if for each object there was example file with all options with inline comments, ideally automatically generated from CRD definition as browsing thru definition itself isn't much fun 😉.

Also this leads me to question about debug mode, because I had to come up with technique ( `kubectl exec -it -n monitoring-system vmagent-vmagent-6d4ffc5697-mzxr9 -c config-reloader -- gunzip /etc/vmagent/config/vmagent.yaml.gz -c` ) to figure out what is generated in order to make this manifests to work and it would be much faster if I could turn vmagent into debug mode where it would print it's config on init and every time it's changed, wdyt?

elmariofredo:
> Thanks a lot!
> 
> May i ask you to add link to metrics-server repository? It's useful for people, who not familiar with it. I guess, it can be placed to the comment line.

Link is there, also https is now 443 instead, let me know if you want to change anything else.

config/examples/vmservicescrape-metrics-server.yml
f41gh7: Yep, i faced the same problem. I thought about code-generator for such examples, but its tricky to write it, there are fields with two options (like intOrStr or Configmap or Secret and so on). It will take some time to write it, probably take a look this week.
 
   Unfortunately no, operator-sdk logger has only Info and Error methods, it possible to replace it with some other logger, but i dont think, that its an option for now. I suggest to use following command:

 ```bash
 kubectl -n monitoring-system get secrets vmagent-CRD_NAME -o 'go-template={{index .data "vmagent.yaml.gz"}}' | base64 -d | gunzip
```
 with it you can use yq for filtering https://github.com/mikefarah/yq

PR #95, opened 2020-10-28
fixed  vmselect args

zhiyin009:
- Fixed field in vmagent example
- Fixed vmselect args logic

### DIFF
--- a/controllers/factory/vmcluster.go
+++ b/controllers/factory/vmcluster.go
@@ -527,8 +527,8 @@ func makePodSpecForVMSelect(cr *v1beta1.VMCluster, c *config.BaseOperatorConf) (
 	}
 	if cr.Spec.ReplicationFactor != nil {
 		var dedupIsSet bool
-		for _, arg := range cr.Spec.VMSelect.ExtraArgs {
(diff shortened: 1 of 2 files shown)


controllers/factory/vmcluster.go
```diff
@@ -527,8 +527,8 @@ func makePodSpecForVMSelect(cr *v1beta1.VMCluster, c *config.BaseOperatorConf) (
 	}
 	if cr.Spec.ReplicationFactor != nil {
 		var dedupIsSet bool
-		for _, arg := range cr.Spec.VMSelect.ExtraArgs {
-			if strings.Contains(arg, "-dedup.minScrapeInterval") {
+		for arg, _ := range cr.Spec.VMSelect.ExtraArgs {
```
f41gh7:
```suggestion
		for arg := range cr.Spec.VMSelect.ExtraArgs {
```
 I suggest to change signature, in this case value can be omitted.

f41gh7:
Thanks! Looks good.

f41gh7:
Thanks for contribution!

zhiyin009:
😊

PR #100, opened 2020-11-05
fixes: deployment update repeatedly

zhiyin009:
# Version
k8s: 1.18.2
victoria-operator: v0.3.0

# Describe the bug
If create crd VMAlert, VMAgent, VMcluster with multiple elements in spec.extraArgs, deployment will be updated repeatedly by operator，again and again....
Though without applying new version, order of args will be changed after reconcile, cause of vm-operator generated args by looping unordered map extraArgs.

# To Reproduce
```yaml
kind: VMSelect
...
spec:
  extraArgs:
    evaluationInternal: 10s
    datasource.lookback: 1m
```

# Fixes
I fixed it by sorting the args array before return.

### DIFF
--- a/controllers/factory/vmagent.go
+++ b/controllers/factory/vmagent.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #105, opened 2020-11-28
In PrometheusConverter in created objects include ownerReferences

teqwve:
I've noticed that removing a `PrometheusRule` doesn't make the corresponding `VMRule` deleted immediately as I've expected. As it seemed easy to code (just adding some `ownerReferences`) I've decided to have some fun and add it myself. For consistency I've updated other converters as well. Apart from e2e tests I've deployed it on some 'real' cluster and tested PrometheusRule behaviour - worked as expected.

During tests I've encountered that issue (at least I think it's an issue) with typo in prober discovery so I've fixed it as well.

Can such a feature be added? I haven't found any related information in documentation or discussion in issues (although I haven't searched a lot).

I'm not sure if this should be enabled by default. From my point of view it should, but I imagine it might break someone else's deployment (someone might, even unintentionally, rely on the fact that converted objects persist after original ones are removed). I can add some `EnablePrometheusConverterOwnerReferences` (false by default) setting to the configuration and pass it down to converters.

### DIFF
--- a/controllers/converter/apis.go
+++ b/controllers/converter/apis.go
@@ -7,6 +7,7 @@ import (
 	"github.com/VictoriaMetrics/operator/controllers/factory"
 	v1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
@@ -45,6 +46,16 @@ func ConvertPromRule(prom *v1.PrometheusRule) *v1beta1vm.VMRule {
 			Name:        prom.Name,
 			Labels:      prom.Labels,
 			Annotations: prom.Annotations,
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion:         v1.SchemeGroupVersion.String(),
+					Kind:               v1.PrometheusRuleKind,
+					Name:               prom.Name,
+					UID:                prom.UID,
+					Controller:         pointer.BoolPtr(true),
+					BlockOwnerDeletion: pointer.BoolPtr(true),
+				},
+			},
 		},
 		Spec: v1beta1vm.VMRuleSpec{
 			Groups: ruleGroups,
@@ -60,6 +71,16 @@ func ConvertServiceMonitor(serviceMon *v1.ServiceMonitor) *v1beta1vm.VMServiceSc
 			Namespace:   serviceMon.Namespace,
 			Annotations: serviceMon.Annotations,
 			Labels:      serviceMon.Labels,
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion:         v1.SchemeGroupVersion.String(),
+					Kind:               v1.ServiceMonitorsKind,
+					Name:               serviceMon.Name,
+					UID:                serviceMon.UID,
+					Controller:         pointer.BoolPtr(true),
+					BlockOwnerDeletion: pointer.BoolPtr(true),
+				},
+			},
 		},
 		Spec: v1beta1vm.VMServiceScrapeSpec{
 			JobLabel:        serviceMon.Spec.JobLabel,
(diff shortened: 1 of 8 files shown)


f41gh7 (changes requested):
Thanks for submitting PR and spotting a typo with converter. 

  It looks good to me. Tests become nicer.

 There is an issue with objects update process. When operator starts, it updates all converted objects, but not all fields are copied from "old" objects. OwnerReferences are missing. It should be placed at:
https://github.com/VictoriaMetrics/operator/blob/master/controllers/vmprometheusconverter_controller.go#L248
https://github.com/VictoriaMetrics/operator/blob/master/controllers/vmprometheusconverter_controller.go#L299
https://github.com/VictoriaMetrics/operator/blob/master/controllers/vmprometheusconverter_controller.go#L349
https://github.com/VictoriaMetrics/operator/blob/master/controllers/vmprometheusconverter_controller.go#L444

with something like: 
```
existingVMProbe.OwnerReferences = vmProbe.OwnerReferences
```

> I'm not sure if this should be enabled by default. From my point of view it should, but I imagine it might break someone else's deployment (someone might, even unintentionally, rely on the fact that converted objects persist after original ones are removed). I can add some `EnablePrometheusConverterOwnerReferences` (false by default) setting to the configuration and pass it down to converters.

 I think that `EnablePrometheusConverterOwnerReferences` with default false - best solution for this problem. Can you add it? 
 Original intention was to provide simple and safe way for migration from one operator to another.

teqwve:
Thanks for the reply!

> There is an issue with objects update process. When operator starts, it updates all converted objects, but not all fields are copied from "old" objects. OwnerReferences are missing. It should be placed at:
> https://github.com/VictoriaMetrics/operator/blob/master/controllers/vmprometheusconverter_controller.go#L248
> https://github.com/VictoriaMetrics/operator/blob/master/controllers/vmprometheusconverter_controller.go#L299
> https://github.com/VictoriaMetrics/operator/blob/master/controllers/vmprometheusconverter_controller.go#L349
> https://github.com/VictoriaMetrics/operator/blob/master/controllers/vmprometheusconverter_controller.go#L444
> 
> with something like:
> 
> ```
> existingVMProbe.OwnerReferences = vmProbe.OwnerReferences
> ```

Sorry, I should have read the code more carefully. I've fixed it and tested manually that redeploying the operator with changed setting updates every rule converted before (and generally that updates actually update OwnerReferences).

I've decided to not add tests for this (redeploying with different settings) because it would require me to restart manager during tests, reset it's configuration after tests and so on... I could just change the global config (since it's the same process and there are pointers everywhere), but it would make e2e tests quite tightly coupled with the actual implementation.

> > I'm not sure if this should be enabled by default. From my point of view it should, but I imagine it might break someone else's deployment (someone might, even unintentionally, rely on the fact that converted objects persist after original ones are removed). I can add some `EnablePrometheusConverterOwnerReferences` (false by default) setting to the configuration and pass it down to converters.
> 
> I think that `EnablePrometheusConverterOwnerReferences` with default false - best solution for this problem. Can you add it?
> Original intention was to provide simple and safe way for migration from one operator to another.

Added.

teqwve:
The build is failing because it's a PR from a fork and github hides secrets :/

I don't know if you want to handle this somewhow, but maybe adding something like:
```
      - name: build crosscompile and push to remote registry
        env:
          TAG:   ${{ env.GITHUB_REF_SLUG }}
        run: |
          if [ -n '${{secrets.REPO_KEY}}' ]; then
            echo ${{secrets.REPO_KEY}} | docker login --username ${{secrets.REPO_USER}} --password-stdin
            make publish-via-docker
          else
            make build-operator-crosscompile docker-operator-crosscompile  # Just test if docker can be built
          fi
``` 
would make some sense?

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

f41gh7:
> The build is failing because it's a PR from a fork and github hides secrets :/
> 
> I don't know if it's an issue for you, or if you want to handle this somewhow, but maybe adding something like:
> 
> ```
>       - name: build crosscompile and push to remote registry
>         env:
>           TAG:   ${{ env.GITHUB_REF_SLUG }}
>         run: |
>           if [ -n '${{secrets.REPO_KEY}}' ]; then
>             echo ${{secrets.REPO_KEY}} | docker login --username ${{secrets.REPO_USER}} --password-stdin
>             make publish-via-docker
>           else
>             make build-operator-crosscompile docker-operator-crosscompile  # Just test if docker can be built
>           fi
> ```
> 
> would make some sense?

 Yes, it makes sense. I'll fix lit a bit later.

teqwve:
@f41gh7 thanks for devoting your time to this PR and generally thanks for you work!

Issue #118, opened 2020-12-07
[feature request] VMNodeScrape


ymmt2005:
It would be great if we can use a custom resource to scrape all the nodes of a Kubernetes cluster.

cf. https://prometheus.io/docs/prometheus/latest/configuration/configuration/#node

f41gh7:
Hello, can you add a bit more context, which fields is needed/wanted. A draft example of custom resources would help a lot.

ymmt2005:
Hi, thank you for maintaining the great product!

In our context, some programs including kubelet and kube-proxy runs on every node as a systemd service.
Since they are not Pods, we need to scrape node ports.  An example would look like:

```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMNodeScrape
metadata:
  name: example-node-scrape
spec:
  port: 8080
  scheme: http
  selector:
    matchLabels:
     kubernetes.io/os: linux
```

We don't need a list of endpoints.  Single port/scheme per VMNodeScrape is enough.

f41gh7:
PoC implementation will be added with PR https://github.com/VictoriaMetrics/operator/pull/123

There are few examples:
1) cadvisor CRD:
https://github.com/VictoriaMetrics/operator/pull/123/files#diff-0ad69e724866cf58aaf0e22de457cfeb493f413da649dd0e0716c40f6d1ad39e
2) full values CRD: 
https://github.com/VictoriaMetrics/operator/pull/123/files#diff-f196e546c105a8ea9520412a10d5252b5d2d2bcaf6b0c3c677a93e93f7ad2490

 If something is missing let me know.

umezawatakeshi:
Hello. I'm interested in this issue. (Actually, ymmt-san is my fellow worker.)

May I comment directly on the PR?

f41gh7:
> Hello. I'm interested in this issue. (Actually, ymmt-san is my fellow worker.)
> 
> May I comment directly on the PR?

 Hello, y, for sure, feel free to add review.

f41gh7:
feature added to the latest v0.6.0 release!

ymmt2005:
Thank you for all the efforts!  We love VictoriaMetrics and its operator.

PR #123, opened 2020-12-10
Adds node scrape

f41gh7:


api/v1beta1/vmagent_types.go
```diff
@@ -202,6 +202,11 @@ type VMAgentSpec struct {
 	// +optional
 	ProbeNamespaceSelector *metav1.LabelSelector `json:"probeNamespaceSelector,omitempty"`
 
+	// NodeScrapeSelector defines VMNodeScrape to be selected for scraping.
+	// if neither PodScrapeNamespaceSelector, nor ProbeSelector nor PodScrapeSelector, nor NodeScrapeSelector are specified,
+	// configuration is unmanaged.
+	// +optional
+	NodeScrapeSelector *metav1.LabelSelector `json:"nodeScrapeSelector,omitempty"`
```
umezawatakeshi, line 208: It would be nice if `nodeScrapeNamespaceSelector` is also added. `VMNodeScrapeSpec` doesn't need `namespaceSelector` because Nodes themselves are not namespaced. But VMNodeScrapes are namespaced like VMServiceScrapes, so `VMAgentSpec` needs `nodeScrapeNamespaceSelector`

config/rbac/role.yaml
```diff
@@ -5,331 +5,346 @@ kind: ClusterRole
...
+  verbs:
+  - '*'
+- apiGroups:
+  - operator.victoriametrics.com
+  resources:
+  - vmnodescrapes/status
+  verbs:
+  - get
+  - patch
+  - update
```
umezawatakeshi, line 350: This YAML document is broken due to wrong indentation.

api/v1beta1/vmnodescrape_types.go
```diff
@@ -0,0 +1,112 @@
...
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// VMNodeScrapeSpec defines discovery for targets placed on kubernetes nodes,
+// usually its node-exporters and other host services.
+// InternalIP is used as __address__ for scraping.
+type VMNodeScrapeSpec struct {
```
umezawatakeshi, line 13: Just a question: Why did you design `VMNodeScrapeSpec` as a "flat" struct? `VMServiceScrapeSpec` has `selector` and `sampleLimit` etc as its direct children while `port` and `relabelConfigs` etc are inside `endpoints` slice (array). Are there any special reason?

ymmt2005: @umezawatakeshi It's because I requested so https://github.com/VictoriaMetrics/operator/issues/118#issuecomment-739917475

api/v1beta1/vmagent_types.go
```diff
@@ -202,6 +202,11 @@ type VMAgentSpec struct {
 	// +optional
 	ProbeNamespaceSelector *metav1.LabelSelector `json:"probeNamespaceSelector,omitempty"`
 
+	// NodeScrapeSelector defines VMNodeScrape to be selected for scraping.
+	// if neither PodScrapeNamespaceSelector, nor ProbeSelector nor PodScrapeSelector, nor NodeScrapeSelector are specified,
```
umezawatakeshi: This line mentions NodesScrapeSelector but similar lines above are not updated to follow this line.

api/v1beta1/vmnodescrape_types.go
umezawatakeshi: @ymmt2005 Oh, understood.

api/v1beta1/vmagent_types.go
f41gh7: I thought about it, but maybe its too complicated for user? You have to care about additional selector and its correctness. 
As far as i know, prometheus made NamespaceSelectors deprecated. 

 Any way, its not hard to add, but its hard to remove later)

umezawatakeshi: There are already namespace selectors for other scraping CRs (`VMServiceScrape` etc.), so I don't think complexity will be a problem. It is preferable to have it for consistency.

I read the [API documentation for the Prometheus operator](https://github.com/prometheus-operator/prometheus-operator/blob/v0.44.1/Documentation/api.md), but I couldn't find any mention of deprecating namespace selectors. Instead, it says `serviceMonitorSelector` and `podMonitorSelector` are deprecated. (It makes me wonder).

f41gh7: Indeed, thanks for pointing, i've thought, that namespaceSelectors are deprecated.. 

 Added implementation for nodeScrapeNamespaceSelector.

k1rk (approved):
LGTM, except for mistake in `config/rbac/role.yaml` which is mentioned by @umezawatakeshi

PR #138, opened 2020-12-29
Removes null crd fields

f41gh7:


f41gh7:
related issue https://github.com/VictoriaMetrics/helm-charts/pull/104

PR #168, opened 2021-02-22
Add deletecollection verb to the editor roles

kfyharukz:
The current editor role manifests don't have verb `deletecollection`.
It seems there's no reason not to grant, so can I do it?


f41gh7:
seems reasonable. Thanks!

PR #184, opened 2021-03-09
fix namespace and serviceaccount in config/examples yaml

TinySong:


config/examples/vmagent.yaml
```diff
@@ -12,6 +12,7 @@ spec:
   staticScrapeSelector: {}
   staticScrapeNamespaceSelector: {}
   replicaCount: 1
+  serviceAccountName: vmagent
```
f41gh7, line 15: lets leave it empty. In this case operator will create needed role itself. It must simply deployment for vmagent.

config/examples/vmagent_rbac.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: vmagent
-  namespace: monitoring-system
+  namespace: default
```
f41gh7, line 5: Lets leave it as is.

 Namespace can be changed with kustomize or with sed replacement, as mentioned at docs.

 For kubectl apply manual changing is needed any way.

Issue #163, opened 2021-02-18
Option override Kubernetes service type or Kubernetes ingress for VM/select/insert 


alexandrst88:
Hi! Ability to override Kubernetes Service Type, like LoadBalancer or NodePort.
```
type: ClusterIP
```

I've checked docs, and wasn't able to find how to override VM/select/insert service type. It's must have to expose vm cluster, if you want to have one dedicated metrics system setup.

https://github.com/VictoriaMetrics/operator/blob/771605e81f9b2a9b04808b8b66ff59b3ffebb980/controllers/factory/vmcluster.go#L726. Seems like it's hardcoded, @f41gh7, I could open PR for this one, if you are ok with?

f41gh7:
Hello, for now, it can be created manually as additional service and type is hardcoded. 

 PR is welcome!  But please keep in mind, that changing value of Type requires service recreation. 

 Probably, its good idea to create separate definition at `spec`, how to expose object to outer world - with additional service, ingress, etc.

TinySong:
any progress ? if not, I will fix this

f41gh7:
Unfortunately, i cannot do it in next few weeks. Feel free to fix:)

f41gh7:
will close it after release.

TinySong:
is there a new bug to be founded for the issue reopened?

f41gh7:
No, everything is ok. We are usually closing issues with features after release.

f41gh7:
Was included to the 0.11.0 release. 

 Example usage:

```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
metadata:
  name: example-vmcluster-persistent
spec:
  retentionPeriod: "4"
  replicationFactor: 2
  vmstorage:
    replicaCount: 1
  vmselect:
    replicaCount: 1
    serviceSpec:
         metadata:
               name: vmselect-vip-access
         spec:
            type: ClusterIP
  vminsert:
    replicaCount: 1
    serviceSpec:
       metadata:
           annotations:
               service.beta.kubernetes.io/aws-load-balancer-subnets: subnet-xxxx, mySubnet
       spec:
           type: LoadBalancer
```

 This configuration produces two additional services:
- service with name `vmselect-vip-access` and clusterIP
- with with name `vminsert-example-vmcluster-persistent-additional-service`

```bash
vminsert-example-vmcluster-persistent                      ClusterIP      10.217.4.64    <none>                                 8480/TCP                     6s
[...]

PR #185, opened 2021-03-10
add service spec to support override service Spec

TinySong:
fix https://github.com/VictoriaMetrics/operator/issues/163

This feature fix will involve the following components. current just finish vmalert，and vmagent。
add service spec in echo CRD to support override the service spec by user
- [X] vmalert
- [X] vmagent
- [X] vminsert
- [X] vmselect
- [X] vmalertmanager
- [X] vmstorage (no need)
- [X] apidoc
 
there may be some questions:
1. Is this implementation feasible ? 
2. in cluster mode, why vminsert component is deployment resource, and vmselect is statefulset ?
3. if vmalertmanager and vmstorage want to  supports custom service types or other field, need create new  no headless service.


f41gh7:
My notes:
 1) need to handle case when user changes service type from clusterIP for NodePort for example. In this case old service must be deleted, than new one created.

2) I suggest  keep the current service as is and create separate service with user defined options - service name, labels and annotations. It allows to control loadbalancer behavior only for this service for instance. 

 About questions:
1. implementation looks good, i can suggest to make generic function, that will set default options for service if its not defined for current service, something like: `func setDefaultService(current, default *v1.ServiceSpec) `
2. vminsert is stateless, vmselect may have pvc for cache.
3. separate service for every CRD object should solve this issue.

TinySong:
all component code done.

schmurfy:
While waiting for something like this to be available how can I set the service type for vmsingle to LoadBalancer with fixed address ?
I tried to create another service of that type pointing to the same selectors but it does not seem to find the target, I am not sure if k8s allows having multiple service pointing to the same pods on the same port.

api/v1beta1/vmagent_types.go
```diff
@@ -237,6 +237,9 @@ type VMAgentSpec struct {
 	// ExtraEnvs that will be added to VMAgent pod
 	// +optional
 	ExtraEnvs []v1.EnvVar `json:"extraEnvs,omitempty"`
+	// ServiceSpec that will be added to vmagent service spec
+	// +optional
+	ServiceSpec v1.ServiceSpec `json:"serviceSpec,omitempty"`
```
f41gh7: Can you add custom type for serviceSpec:

```go
 type ServiceSpec struct {
 Name string
 Spec: v1.ServiceSpec
  Annotations: map[string]string
  Labels: map[string]string
```

 and use it as 
``` go
ServiceSpec *ServiceSpec
```
 for all data structures.

f41gh7:
> While waiting for something like this to be available how can I set the service type for vmsingle to LoadBalancer with fixed address ?
> I tried to create another service of that type pointing to the same selectors but it does not seem to find the target, I am not sure if k8s allows having multiple service pointing to the same pods on the same port.

 It should be ok to create additional services, check labelSelector values, probably you have issue with it.

f41gh7:
@TinySong  sorry for merge conflict, can you resolve it.

api/v1beta1/vmagent_types.go
bobsongplus: The struct looks more reasonable, I will fix it, thanks.

TinySong:
@f41gh7  confliction fixed

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #191, opened 2021-03-20
Add a way to configure simple topologySpreadConstraints and PodDisruptionBudget


umezawatakeshi:
To deploy HA monitoring cluster, it is important to configure appropriate topologySpreadConstraints (tSC) and PodDisruptionBudget (PDB). If the users want to configure simple tSCs and PDBs just covers the Pods created from the operator's custom resources, the users have to write same label selectors as the operator creates.

**Suggestion:**

- When labelSelector in tSC is not specified (= nil), the operator fills the selector with .spec.selector in the resulting Deployment or StatefulSet. Specifying nil for the labelSelector is meaningless, so I think that it does not cause a problem to use nil as a special purpose.
- Add podDisruptionBudget parameter and the operator creates PDB from it.

For example, the operator creates the following Deployment and PDB from the following VMAgent. (note that many fields which are not concerned by this suggestion are omitted)

```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAgent
metadata:
  name: example
spec:
  replicaCount: 3
  topologySpreadConstraints:
  - maxSkew: 1
    topologyKey: kubernetes.io/hostname
    whenUnsatisfiable: DoNotSchedule
    labelSelector:
      matchLabels:
        app.kubernetes.io/name: vmagent
  - maxSkew: 1
    topologyKey: kubernetes.io/os
    whenUnsatisfiable: ScheduleAnyway
    # labelSelector is not specified
  podDisruptionBudget:
    minAvailable: 2
```

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: vmagent-example
  namespace: default
spec:
  replicas: 3
  selector:
    matchLabels:
      app.kubernetes.io/component: monitoring
      app.kubernetes.io/instance: example
      app.kubernetes.io/name: vmagent
      managed-by: vm-operator
  template:
    spec:
      topologySpreadConstraints:
      - labelSelector: # specified labelSelector is just copied
          matchLabels:
            app.kubernetes.io/name: vmagent
        maxSkew: 1
        topologyKey: kubernetes.io/hostname
        whenUnsatisfiable: DoNotSchedule
      - labelSelector: # labelSelector is not specified -> copied from .spec.selector
          matchLabels:
            app.kubernetes.io/component: monitoring
            app.kubernetes.io/instance: example
            app.kubernetes.io/name: vmagent
            managed-by: vm-operator
        maxSkew: 1
[...]

PR #199, opened 2021-03-28
Add a easy way to create simple topologySpreadConstraints and PodDisruptionBudget

umezawatakeshi:
implements #191 

Signed-off-by: UMEZAWA Takeshi <umezawa@utsoft.org>


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

umezawatakeshi:
By the way, there is much duplicate code not only in this PR but also in the originally existing code. I think it is bug-prone and should be solved. However, I'm not sure that it should be done **now** because it could lead to a breaking change in Golang-level API, depending on how to solve it.

f41gh7:
> By the way, there is much duplicate code not only in this PR but also in the originally existing code. I think it is bug-prone and should be solved. However, I'm not sure that it should be done **now** because it could lead to a breaking change in Golang-level API, depending on how to solve it.

 Yep, i've refactored it a bit.  For now, i prefer to make small patches with improvements - remove duplicate code and unify some parts.

  All breaking changes can be done before v1.0 release, if its needed.

Issue #205, opened 2021-03-30
cannot update or create vmcluster


lujiajing1126:
vm-operator 0.12.1
vm cluster: 1.57.1-cluster

We intend to upgrade from `1.57.0-cluster` to `1.57.1-cluster`, but not all components updated successfully.

The logs of the operators are pasted as following,

```
{"level":"info","ts":1617112677.317875,"logger":"controller_vmcluster","msg":"Reconciling VMCluster","Request.Namespace":"monitoring-system","Request.Name":"hera-vmcluster"}
{"level":"info","ts":1617112677.3252242,"logger":"factory","msg":"creating psp for vmcluster"}
{"level":"info","ts":1617112677.397108,"logger":"factory","msg":"create or update vmstorage for cluster","controller":"vmstorage","cluster":"hera-vmcluster"}
{"level":"info","ts":1617112677.3972619,"logger":"factory","msg":"vmstorage was found, updating it","controller":"vmstorage","cluster":"hera-vmcluster"}
{"level":"info","ts":1617112677.4064224,"logger":"factory","msg":"vmstorage sts was reconciled","controller":"vmstorage","cluster":"hera-vmcluster"}
{"level":"info","ts":1617112679.4066384,"logger":"factory","msg":"sts update is needed","sts":"vmstorage-hera-vmcluster","currentVersion":"vmstorage-hera-vmcluster-677cbbdd8","desiredVersion":"vmstorage-hera-vmcluster-58757b567b"}
{"level":"info","ts":1617112679.4067063,"logger":"factory","msg":"checking if update needed","controller":"sts.rollingupdate","desiredVersion":"vmstorage-hera-vmcluster-58757b567b"}
{"level":"info","ts":1617112679.4068286,"logger":"factory","msg":"update isn't needed","controller":"sts.rollingupdate","desiredVersion":"vmstorage-hera-vmcluster-58757b567b"}
{"level":"info","ts":1617112679.4178104,"logger":"factory","msg":"check pods availability"}
{"level":"info","ts":1617112679.4179344,"logger":"factory","msg":"pods available","count":2,"spec-count":2}
{"level":"info","ts":1617112679.4179585,"logger":"factory","msg":"create or update vmselect for cluster","controller":"vmselect","cluster":"hera-vmcluster"}
[...]

PR #211, opened 2021-04-02
Check storage resize first

lujiajing1126:
Signed-off-by: Megrez Lu <lujiajing1126@gmail.com>

According to the last commit 306b1014bb0db97a7ed298bbe12cb22637978412, we will diff pvc objects.

But the following resize check would be unreachable.

Relate to issue #205

### DIFF
--- a/controllers/factory/expansion.go
+++ b/controllers/factory/expansion.go
@@ -66,6 +66,12 @@ func wasCreatedSTS(ctx context.Context, rclient client.Client, pvcName string, n
 		if actualPVC == nil || newPVC == nil {
 			return false
 		}
+
+		if i := newPVC.Spec.Resources.Requests.Storage().Cmp(*actualPVC.Spec.Resources.Requests.Storage()); i != 0 {
+			log.Info("must re-recreate sts, its pvc claim was changed", "size-diff", i)
+			return true
+		}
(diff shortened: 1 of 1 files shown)


controllers/factory/expansion.go
```diff
@@ -66,19 +66,20 @@ func wasCreatedSTS(ctx context.Context, rclient client.Client, pvcName string, n
 		if actualPVC == nil || newPVC == nil {
 			return false
 		}
+
+		if i := newPVC.Spec.Resources.Requests.Storage().Cmp(*actualPVC.Spec.Resources.Requests.Storage()); i != 0 {
+			log.Info("must re-recreate sts, its pvc claim was changed", "size-diff", i)
```
f41gh7, line 71: I've realized, that  `Storage().Cmp()` - returns not a diff size, but indicator int - is it the same or not. 

 Will fix it later.

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

controllers/factory/expansion.go
lujiajing1126: Thanks for pointing it out! Maybe I could send another fix later.

PR #212, opened 2021-04-02
Fix size diff result

lujiajing1126:
Signed-off-by: Megrez Lu <lujiajing1126@gmail.com>

As mentioned in the #211 , the result of `Quantity.Cmp(Quantity) -> int` is not the actual diff size.

I am not sure if this modification is the most elegant way. Pls have a check.


f41gh7:
Looks good!

f41gh7:
Thanks!

PR #267, opened 2021-06-21
modify the implementation of VMPodScrape.spec.selector

fatsheep9146:
hi, I want to make a suggestion to modify the implementation of VMPodScrape.spec.selector from relabel_configs to kubernetesSDConfig.selector.

In our cases, if we use relabels_config, then it will list all pods from specified namespace,s even all namespaces, this will make a big pressure for apiserver if the k8s cluster has many nodes and pods.

### DIFF
--- a/controllers/factory/scrapes_build.go
+++ b/controllers/factory/scrapes_build.go
@@ -294,7 +294,7 @@ func generatePodScrapeConfig(
 	cfg = honorTimestamps(cfg, ep.HonorTimestamps, overrideHonorTimestamps)
 
 	selectedNamespaces := getNamespacesFromNamespaceSelector(&m.Spec.NamespaceSelector, m.Namespace, ignoreNamespaceSelectors)
-	cfg = append(cfg, generateK8SSDConfig(selectedNamespaces, apiserverConfig, basicAuthSecrets, kubernetesSDRolePod))
+	cfg = append(cfg, generatePodK8SSDConfig(selectedNamespaces, m.Spec.Selector, apiserverConfig, basicAuthSecrets, kubernetesSDRolePod))
 
 	if ep.Interval != "" {
 		cfg = append(cfg, yaml.MapItem{Key: "scrape_interval", Value: ep.Interval})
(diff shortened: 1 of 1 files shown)


f41gh7:
Did you test those changes? 

 Probably, its an issue with `vmagent` ? Can you try to upgrade `vmagent` to the 1.61.1 version? 

  Also, can you share value of vmagent metric `vm_promscrape_discovery_kubernetes_group_watchers` ?

fatsheep9146:
Sorry for my delay. 

> Did you test those changes?

Yes, I have test those changes, and I can supplement with some unit tests later

> Also, can you share value of vmagent metric vm_promscrape_discovery_kubernetes_group_watchers?

This metric indicates the number of groupWatcher, a groupWatcher denotes a client which watchs the kubernetes objects for a given kubernetes scrape config. For example, if you specify a job likes this

```
- job_name: test
  honor_labels: false
  kubernetes_sd_configs:
  - role: pod
    namespaces:
      names:
      - kube-system
    selectors:
    - role: pod
      label: k=v
```

then it will create a new group watcher to watch for the pods object which match the label selector k=v and in namespace kube-system.

> Probably, its an issue with vmagent ? Can you try to upgrade vmagent to the 1.61.1 version?

This is not vmagent's issue, vmagent supports adding the selectors  when you specify a scrape config of kubernetes.SDConfig

```
- job_name: test
  honor_labels: false
  kubernetes_sd_configs:
  - role: pod
    namespaces:
      names:
      - kube-system
    selectors:
    - role: pod
      label: k=v
```

And when specified like this, vmagent will create a watch client which the watch URL is 

```
https://xxx/api/v1/namespaces/kube-system/pods?labe
lSelector=k%3Dv
```

the response of the watch request will be the pods match the selector and namespace

f41gh7:
Yep, i know,  thank you for sharing details.

  My assumption was, that creating to many groupWatchers increases load for  kubernetes api server. Seems like i's wrong.

f41gh7:
WDYT, should it be also added to the `VMServiceScrape` ?

f41gh7 (approved):
LGTM

f41gh7:
Fyi, default implementation was changed and operator now do not generate `selectors` by default for `VMPodscrape`.

 This behavior could be changed with `vmagent.spec.enableKubernetesAPISelectors: true` flag. 

Recently, we discovered, that creating too many urlWatchers increases load on Kubernetes API server. And it's better to re-use it and apply client side filtering.  See related issue and commit for details.

PR #341, opened 2021-10-04
changes default behavior for Selector and NamespaceSelector

f41gh7:
now it will match by default any exist objects if both selectors is undefined for given CRD resource.
updates docs section


k1rk (approved):
👍

PR #356, opened 2021-10-18
Default to DNS search paths if ClusterDomainName isn't set explicitly

flokli:
`.svc.<cluster domain>` usually is in the DNS search path of Pods
already, so make use of that in the default case, instead of hardcoding
cluster.local everywhere.

Users with broken DNS search paths can still explicitly configure the
cluster domain via VM_CLUSTERDOMAINNAME, in that case, VM will use the
FQDN.


f41gh7:
Thanks for contribution!

Issue #344, opened 2021-10-07
name of volumeMount in vmstorage StatefulSet does not follow name of volumeClaimTemplate


umezawatakeshi:
For example, when I give a VMCluster like:

```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
metadata:
  name: example-cluster
  namespace: monitoring
spec:
  vmstorage:
    storage:
      volumeClaimTemplate:
        metadata:
          name: foo-db    # THIS!!
        spec:
          storageClassName: ceph-hdd-block
          resources:
            requests:
              storage: 3Gi

  vmselect:
    replicaCount: 1

  vminsert:
    replicaCount: 1
```

... the resulting StatefulSet for vmstorage becomes:

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: vmstorage-example-cluster
  namespace: monitoring
spec:
  template:
    spec:
      containers:
        - args:
          (snip)
          volumeMounts:
            - mountPath: vmstorage-data
              name: vmstorage-db    # XXX
  volumeClaimTemplates:
    - apiVersion: v1
      kind: PersistentVolumeClaim
      metadata:
        name: foo-db    # The name given in VMCluster.spec.vmstorage.storage.volumeClaimTemplate.metadata.name
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 3Gi
        storageClassName: ceph-hdd-block
        volumeMode: Filesystem
      status:
        phase: Pending
```

... and StatefulSet controller cannot create Pods due to volume name mismatch.

According to https://github.com/VictoriaMetrics/operator/blob/v0.19.1/controllers/factory/alertmanager.go#L571 , it seems that the operator allows renaming volumeClaimTemplate. However, `.spec.template.spec.containers[0].volumeMounts[0].name` of the resulting StatefulSet does not follow the volumeClaimTemplate's name. I think it is a problem.

VMSelect and VMAlertmanager, which also create StatefulSets, may have the same problem but I don't have checked them.

f41gh7:
Thanks for reporting! There was incorrect logic for this field, it was fixed at commit 3f89ec83f6c1dc2138d9adf515997864d91407cb. related PR attached to the issue.

 Problem occured only with `vmstorage` and `vmselect`.

 Fix will be included into the next release.

umezawatakeshi:
[...]

PR #361, opened 2021-10-26
adds new logic for statefulset pod re-creation

f41gh7:
now it must properly handle cases for pods rolling update interuption
when statefulset was re-created

In this case, currentRevision will be written to the statefulset.status
And updateRevision will be taken from first updated pod and written to the statefulset.status.updateRevesion

it should fix an issue https://github.com/VictoriaMetrics/operator/issues/344

### DIFF
--- a/controllers/factory/scrapes_test.go
+++ b/controllers/factory/scrapes_test.go
@@ -625,6 +625,7 @@ scrape_configs:
       names:
       - default
   metrics_path: /metrics
+  bearer_token: some-bearer
   relabel_configs:
   - action: keep
     source_labels:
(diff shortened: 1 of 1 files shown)


controllers/factory/k8stools/expansion.go
```diff
@@ -67,7 +88,7 @@ func wasCreatedSTS(ctx context.Context, rclient client.Client, pvcName string, n
 		if actualPVC == nil && newPVC == nil {
 			return false
 		}
-		// actual name for pvc was changed or it was added
+		// one of pvc is not nil
```
tenmozes, line 91: it better to move some varialbe with meaningful lname eg

hasNotNilPVC := actualPVC == nil && newPVC != nil) || (actualPVC != nil && newPVC == ni
if hasNotNilPVC {return true}

PR #377, opened 2021-11-11
config/examples/vmalert-full.yaml: fix yaml

flokli:
Applying the example will fail:

```
Error from server (notifier.url at idx: 1 cannot be empty): error when creating "vmalert-full.yaml": admission webhook "vmalert.victoriametrics.com" denied the request: notifier.url at idx: 1 cannot be empty
```

This probably shouldn't create two notifiers, one without an url, but
one notifier with all the attributes.


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #378, opened 2021-11-11
VMAlert CRD: improve Remove{Read,Write} description

flokli:
Please review on a commit-by-commit basis. It seems `make manifests` hasn't been run, which explains the big unrelated diff.

```
    Make it more clear RemoteRead is for restoring vmalert state across
    restarts, and RemoteWrite is used for storing rule results and vmalert
    state.
```


flokli:
Let's wait for the results of https://github.com/VictoriaMetrics/VictoriaMetrics/issues/1810 before merging this.

f41gh7 (approved):
LGTM

f41gh7:
Main reason to keep both flags - using vmalert with vmcluster. In this case, remoteWrite.url must to point to the url of vminsert and remoteRead.url to the url of vmselect. 

 Thanks for improvements to the documentation!

PR #380, opened 2021-11-14
Help controller-gen with recursive Route type

addreas:
Ended up not being able to apply the CRDs and that seems to be related to the recursive Route in VMAlertmanagerConfigs. I get the error `The CustomResourceDefinition "vmalertmanagerconfigs.operator.victoriametrics.com" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[route].properties[routes].items.type: Required value: must not be empty for specified array items`. 

Before the CRD were re-generated in a recent commit there was a `x-kubernetes-preserve-unknown-fields: true` inside the `items` object, but I couldn't figure out how to annotate things to imitate that. 

Without `kubebuilder:validation:Schemaless` the empty `items: {}` still gave the same error. I have not tested to see if that ends up parsing things differently when creating configurations, only that the CRD itself gets created without error.

### DIFF
--- a/api/v1beta1/vmalertmanagerconfig_types.go
+++ b/api/v1beta1/vmalertmanagerconfig_types.go
@@ -142,6 +142,8 @@ type Route struct {
 	// +optional
 	Continue bool `json:"continue,omitempty"`
 	// Child routes.
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:pruning:PreserveUnknownFields
 	Routes []*Route `json:"routes,omitempty"`
 	// MuteTimeIntervals for alerts
 	// +optional
--- a/config/crd/bases/operator.victoriametrics.com_vmalertmanagerconfigs.yaml
+++ b/config/crd/bases/operator.victoriametrics.com_vmalertmanagerconfigs.yaml
@@ -2519,8 +2519,7 @@ spec:
                     type: string
                   routes:
                     description: Child routes.
-                    items: {}
-                    type: array
+                    x-kubernetes-preserve-unknown-fields: true
                 required:
                 - receiver
                 type: object
--- a/config/crd/legacy/operator.victoriametrics.com_vmalertmanagerconfigs.yaml
+++ b/config/crd/legacy/operator.victoriametrics.com_vmalertmanagerconfigs.yaml
@@ -2476,8 +2476,7 @@ spec:
                   type: string
                 routes:
                   description: Child routes.
-                  items: {}
-                  type: array
+                  x-kubernetes-preserve-unknown-fields: true
               required:
               - receiver
               type: object


flokli:
@addreas this might have been https://github.com/VictoriaMetrics/operator/pull/378/commits/12e29c0bc5df892cd367ae3e0e81e15fe31ce279. Is `make manifests` now providing the CRDs correctly?

flokli:
I checked out this PR locally, and can confirm running `make manifests` doesn't create any further diff in `config/crd/*` - we probably should have CI ensuring the committed CRDs always match what would be generated.

@addreas I noticed that before https://github.com/VictoriaMetrics/operator/commit/12e29c0bc5df892cd367ae3e0e81e15fe31ce279 the `x-kubernetes-preserve-unknown-fields` was at the `routes` level, now it's at `routes.items`. Is that a problem?

```diff
--- a/config/crd/bases/operator.victoriametrics.com_vmalertmanagerconfigs.yaml
+++ b/config/crd/bases/operator.victoriametrics.com_vmalertmanagerconfigs.yaml
@@ -1,8 +1,11 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.6.2
+  creationTimestamp: null
   name: vmalertmanagerconfigs.operator.victoriametrics.com
 spec:
   group: operator.victoriametrics.com
@@ -2516,9 +2519,7 @@ spec:
                     type: string
                   routes:
                     description: Child routes.
-                    items:
-                      x-kubernetes-preserve-unknown-fields: true
-                    type: array
+                    x-kubernetes-preserve-unknown-fields: true
                 required:
                 - receiver
                 type: object
@@ -2536,3 +2537,9 @@ spec:
     storage: true
     subresources:
       status: {}
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
diff --git a/config/crd/bases/operator.v
```

I also noticed `x-kubernetes-preserve-unknown-fields` was present in some other fields (can be seen by cherry-picking this PR and then running `git diff 12e29c0bc5df892cd367ae3e0e81e15fe31ce279~1..HEAD`)

I don't have enough context on why `x-kubernetes-preserve-unknown-fields` existed in first place.

addreas:
@flokli I'm not actually sure if it would cause a problem. This ended up being the next best thing that I could come up with using just kubebuilder markers. Might have missed something though.

Since the "bail out" on the type checking happens earlier than it did before my assumption it that this would allow a larger set of values through the schema validation. What I'm not sure about is if this somehow changes something else that I haven thought of that then causes the unmarshaling to fail inside the controller somehow.

I was thinking about testing this out more properly by creating some nested routes and checking if they make it all the way through. Haven't gotten around to that yet, though.

addreas:
This problem has in practice disappeared in master, though solved in the Makefile instead. That works well enough for me, so I'll close this.

PR #411, opened 2022-01-19
Add prefix to `job_name` to prevent name collision

tommy351:
Add `serviceScrape`, `podScrape`, `nodeScrape`, `staticScrape` and `probe` prefixes to `job_name` to prevent name collision. This pull request is similar to https://github.com/prometheus-operator/prometheus-operator/pull/3913 .


f41gh7:
Thanks! It's major change, let's schedule it for 0.23.0.

f41gh7:
Thanks!

PR #421, opened 2022-02-08
Process remote write secret even if BasicAuth is not set

artifactori:
Remote write secret is ignored when basic auth is not set. There is possibility of using other auth method without using basic auth method, e.g. bearer token.

This PR ensures that remote write secret is processed even if basic auth is not set.

### DIFF
--- a/controllers/factory/vmagent_test.go
+++ b/controllers/factory/vmagent_test.go
@@ -764,6 +764,33 @@ func TestBuildRemoteWrites(t *testing.T) {
 			},
 			want: []string{"-remoteWrite.oauth2.clientID=,some-id", "-remoteWrite.oauth2.clientSecret=,some-secret", "-remoteWrite.oauth2.scopes=,scope-1", "-remoteWrite.oauth2.tokenUrl=,http://some-url", "-remoteWrite.url=localhost:8429,localhost:8431", "-remoteWrite.sendTimeout=10s,15s"},
 		},
+		{
+			name: "test bearer token",
+			args: args{
+				ssCache: &scrapesSecretsCache{
+					bearerTokens: map[string]string{
+						"remoteWriteSpec/localhost:8431": "token",
+					},
+				},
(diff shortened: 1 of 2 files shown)


controllers/factory/vmagent_test.go
```diff
@@ -713,7 +735,7 @@ func TestBuildRemoteWrites(t *testing.T) {
 			name: "test oauth2",
```
artifactori, line 735: This test case is pretty similar with the one handled on this pull request: Using remote-write secret without basic auth. (In this case, OAuth2 is being used.)

At first, I wonder why this test case works, but I found out that secret is passed from `ssCache`, not from the `loadScrapeSecrets` in **scrapes.go**.  In the normal operation, secrets are passed from `loadScrapeSecrets`, and without this patch, `loadScrapeSecrets` won't read the secret if `basicAuth` isn't set.

controllers/factory/vmagent_test.go
```diff
@@ -113,6 +113,28 @@ func TestCreateOrUpdateVMAgent(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "fail if bearer token secret is missing, without basic auth",
```
artifactori, line 117: I can't see any example of using `want` in this method, so since there's `wantErr`, I used negation test here.

I don't think negation test is a good idea as this test is surrounded with non-negation one, but at least this works for now, so might need some help in this part.

artifactori:
I'm using one commit before the last commit in `master` branch as base of this pull request. The last commit has failed CI, so I'm thinking to use the one with working CI instead.

f41gh7 (approved):
LGTM

controllers/factory/vmagent_test.go
f41gh7: Thanks for pointing, will take a look on it.

f41gh7:
Thanks for contribution!

PR #435, opened 2022-02-20
add selector to the generate k8s sd config

f41gh7:
it must increase parallelisiation of discovery at vmagent

vmagent creates k8s api connection per given watch url - role + namespace + selector

it must allow to speed up initial syncronization for vmagent config


f41gh7:
It's not bring any performance tips, no point to use it.

Issue #450, opened 2022-03-12
allowed override vmagent volume


wenerme:
want to prevent vmagent data lost by mount

```yaml
  volumes:
  - name: persistent-queue-data
    hostPath:
      path: /data/var/vmagent/persistent-queue-data
```

operator should merge volume by name.

https://github.com/VictoriaMetrics/operator/blob/10a53e58c8dcb388e7dda62db836513fb0618e90/controllers/factory/vmagent.go#L246-L251

f41gh7:
I think, that it can be achieved with statefulset instead of deployment. Deployment in case of persistent volume claim mount must has upgradeStrategy: ReCreate and replicas: 1.

 Related issue https://github.com/VictoriaMetrics/operator/issues/219


 I hope, soon it will be added.

wenerme:
Some provider seems do not need to recreate, like openebs zfs, allowed multi mount.

hagen1778:
@f41gh7 can we have docs updated accordingly?

f41gh7:
Added an example https://github.com/VictoriaMetrics/operator/blob/master/config/examples/vmagent_stateful_with_sharding.yaml

Issue #448, opened 2022-03-10
enable override of httpListenAddr flag


lllamnyp:
According to

https://github.com/VictoriaMetrics/operator/blob/10a53e58c8dcb388e7dda62db836513fb0618e90/controllers/factory/vmcluster.go#L520

the command line argument `-httpListenAddr=:%s` is hardcoded for the vmselect pods. This makes it impossible, for example, to force listening only on the loopback interface by launching vmselect with `-httpListenAddr=127.0.0.1:<port>`, which is useful when using a TLS-terminating or authn/authz proxy launched as a sidecar container.

A possible fix to this may be to replace lines 519--521 with the following:

```go
	args := []string{}
	if _, ok := cr.Spec.VMSelect.ExtraArgs["httpListenAddr"]; !ok {
		args = append(args, fmt.Sprintf("-httpListenAddr=:%s", cr.Spec.VMSelect.Port))
	}
```

i.e. if the user explicitly specified `httpListenArgs` among the map of `extraArgs`, there should be no need to set the default value.

f41gh7:
Changes was included to the v0.24.0 release. Feel free to re-open the issue if it doesn't solve it.

PR #452, opened 2022-03-16
controllers/factor: allow to override default args

f41gh7:
it must allow to controller application by user needs and redifene arguments with user values
https://github.com/VictoriaMetrics/operator/issues/450
https://github.com/VictoriaMetrics/operator/issues/448


arctan90:
Hello @f41gh7 
I  picked up this PR today, and found a  🐛  maybe :P 
vmagent.go line349 , the 'break' should be braced in the brace.
```
	var containsMaxDiskUsage = false
	for arg := range cr.Spec.ExtraArgs {
	      if arg == "remoteWrite.maxDiskUsagePerURL" {
	          containsMaxDiskUsage = true
	    }
	    break
	}
	if !containsMaxDiskUsage {
	    // limit to 1GB
	    args = append(args, "-remoteWrite.maxDiskUsagePerURL=1073741824")
	}
```

Issue #454, opened 2022-03-17
Can't customize VMServiceScrape of vmcluster


artifactori:
I can't customize the `VMServiceScrape` of `VMCluster` to add extra labels on the metrics.

I've checked the [API docs](https://docs.victoriametrics.com/operator/api.html#vmclusterspec) and [the `VMServiceScrape` creation code from `VMCluster`](https://github.com/VictoriaMetrics/operator/blob/06dada488d629d4d321985e80d14ee04e099bdfd/controllers/factory/scrapes.go#L772-L821) and seems there's no way to modify it for now.

f41gh7:
Changes was included to the v0.24.0 release. Feel free to re-open the issue if it doesn't solve it.

 Thanks for contribution!

artifactori:
It does solve the problem, so I don't need to reopen the issue. Thank you very much, @f41gh7!

PR #455, opened 2022-03-17
Allow customization of VMServiceScrape in various objects

artifactori:
Hi @f41gh7,
This is my 2nd pull request: kinda big and includes API changes, so I hope that it can be reviewed well before having any next actions.

This pull request solves #454 by adding `serviceScrapeSpec` on various places, so we can have more flexibility with the generated `VMServiceScrape` objects.

Thank you.

### DIFF
--- a/controllers/factory/scrapes.go
+++ b/controllers/factory/scrapes.go
@@ -806,7 +806,7 @@ func CreateVMServiceScrapeFromService(ctx context.Context, rclient client.Client
 		Spec: *serviceScrapeSpec,
 	}
 	scrapeSvc.Spec.Selector = metav1.LabelSelector{MatchLabels: service.Spec.Selector}
-	scrapeSvc.Spec.Endpoints = endPoints
+	scrapeSvc.Spec.Endpoints = append(endPoints, scrapeSvc.Spec.Endpoints...)
 
 	err := rclient.Get(ctx, types.NamespacedName{Namespace: service.Namespace, Name: service.Name}, &existVSS)
 	if err != nil {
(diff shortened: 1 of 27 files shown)


controllers/factory/scrapes.go
```diff
@@ -798,11 +803,11 @@ func CreateVMServiceScrapeFromService(ctx context.Context, rclient client.Client
...
 			Annotations:     service.Annotations,
 		},
-		Spec: victoriametricsv1beta1.VMServiceScrapeSpec{
-			Selector:  metav1.LabelSelector{MatchLabels: service.Spec.Selector},
-			Endpoints: endPoints,
-		},
+		Spec: *serviceScrapeSpec,
 	}
+	scrapeSvc.Spec.Selector = metav1.LabelSelector{MatchLabels: service.Spec.Selector}
+	scrapeSvc.Spec.Endpoints = append(endPoints, scrapeSvc.Spec.Endpoints...)
```
f41gh7, line 809: WDYT about patching generated endpoints?

f41gh7 (approved):
LGTM

controllers/factory/scrapes.go
artifactori: Could you please elaborate more on this? How can I patch generated endpoints?

artifactori: Could you please elaborate more on this? @f41gh7 

I assume that you want to patch any endpoints generated by the operator. Since endpoints are array, not map, which isn't straightforward to merge, could you please elaborate how should 2 sets of endpoints be merged?

f41gh7: I'll try to add it as follow-up.

I think, it's possible to match array item by `endpoint.Port` and override generated values (port and path). Usually generated endPoints contains only 1 item, so it should to be easy to match and update. It must cover the case, when you want to add some settings for application endpoint.

f41gh7:
Thanks for contribution!

PR #473, opened 2022-05-03
Adding Name field to VMUser

### LINKED ISSUE #472: Name field in VMUserSpec
The name field passed for each user to vmauth is used in the label `username` for metrics `vmauth_user_requests_total` which will be useful to identify the source of the requests.

The documentation of victoriametrics states as below.
`vmauth` exports `vmauth_user_requests_total` metric with `username` label. The `username` label value equals to `username` field value set in the `-auth.config` file. It is possible to override or hide the value in the label by specifying `name` field. For example, the following config will result in `vmauth_user_requests_total{username="foobar"}` instead of `vmauth_user_requests_total{username="secret_user"}`

pavan541cs:
This PR is to fix #472

### DIFF
--- a/api/v1beta1/vmuser_types.go
+++ b/api/v1beta1/vmuser_types.go
@@ -12,6 +12,9 @@ import (
 
 // VMUserSpec defines the desired state of VMUser
 type VMUserSpec struct {
+	// Name of the VMUser object.
+	// +optional
+	Name *string `json:"name,omitempty"`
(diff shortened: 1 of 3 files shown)


pavan541cs:
@f41gh7 Please review this PR and comment on the same.

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

pavan541cs:
What is the plan for the next release? Any tentative dates?

PR #488, opened 2022-05-31
Fix alertmanager metrics endpoint when routePrefix is configured

blesswinsamuel:
VMAlertmanager CR allows us to configure routePrefix field. If this is used, metrics is not accessible at `/metrics` (it is only accessible at `<routePrefix>/metrics`. Currently, the metrics endpoint that comes up in the generated VMServiceScrape of alertmanager is `/metrics` (even if routePrefix is configured). This PR fixes this by adding the route prefix from routePrefix field of vmalertmanager CR, so that the metrics endpoint points to the right path.

### DIFF
--- a/api/v1beta1/additional.go
+++ b/api/v1beta1/additional.go
@@ -16,13 +16,12 @@ import (
 )
 
 const (
-	vmPathPrefixFlagName             = "http.pathPrefix"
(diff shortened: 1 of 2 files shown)


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #489, opened 2022-06-05
api/v1beta1/VMUser: adds tokenRef

f41gh7:
it allows to fetch value for Bearer Token authorization at VMAuth from pre-created secret.
It works with the same rules as passwordRef


pavan541cs:
@f41gh7 Can you create the latest release. We are waiting for this feature to go live

Issue #498, opened 2022-07-05
Add global setting to override container registry


tamcore:
Currently it's required to set 

- VM_VMALERTDEFAULT_IMAGE
- VM_VMAUTHDEFAULT_IMAGE
- VM_VMAGENTDEFAULT_IMAGE
- VM_VMBACKUP_IMAGE
- VM_VMSINGLEDEFAULT_IMAGE
- VM_VMALERTMANAGER_ALERTMANAGERDEFAULTBASEIMAGE
- VM_VMCLUSTERDEFAULT_VMSELECTDEFAULT_IMAGE
- VM_VMCLUSTERDEFAULT_VMSTORAGEDEFAULT_IMAGE
- VM_VMCLUSTERDEFAULT_VMINSERTDEFAULT_IMAGE
- VM_VMALERTDEFAULT_CONFIGRELOADIMAGE

individually to point to a registry cache. It would be nice to have a simple setting like `VM_CONTAINERREGISTRY` or something to override the current default of `docker.io`.

f41gh7:
Added to the v0.26.0 release, thanks for contribution!

PR #501, opened 2022-07-08
Add global setting to override container registry

tamcore:
This adresses #498. Tested by deploying the compiled operator with
```
      env:
      - name: VM_CONTAINERREGISTRY
        value: harbor.local/docker.io
```
and then deploying a VMSingle instance copy&paste from https://docs.victoriametrics.com/operator/quick-start.html, resulting in
```
philipp@ ~ $ k get deployment vmsingle-example-vmsingle-persisted -o yaml | grep image:
        image: harbor.local/docker.io/victoriametrics/victoria-metrics:v1.77.2
philipp@ ~ $
```

### DIFF
--- a/controllers/factory/alertmanager.go
+++ b/controllers/factory/alertmanager.go
@@ -203,7 +203,7 @@ func makeStatefulSetSpec(cr *victoriametricsv1beta1.VMAlertmanager, c *config.Ba
 
 	cr = cr.DeepCopy()
 
-	image := fmt.Sprintf("%s:%s", cr.Spec.Image.Repository, cr.Spec.Image.Tag)
+	image := fmt.Sprintf("%s/%s:%s", c.ContainerRegistry, cr.Spec.Image.Repository, cr.Spec.Image.Tag)
 
 	amArgs := []string{
 		fmt.Sprintf("--config.file=%s", alertmanagerConfFile),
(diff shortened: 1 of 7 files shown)


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

oxyno-zeta:
Hello,

This looks like a breaking change. 
We just spend 1 hour to understand and found why we had "custom-registry/vmstorage:XXX" to "docker.io/custom-registry/vmstorage:XXX"

Moreover, putting this in front of all images like this includes all config reloaders without any notice.

Maybe you should add a dedicated note into the release to inform people about this change.

Regards,

Oxyno-zeta

f41gh7:
> Hello,
> 
> This looks like a breaking change no ? We just spend 1 hour to understand and found why we had this change: "custom-registry/vmstorage:XXX" to "docker.io/custom-registry/vmstorage:XXX" in pods.
> 
> Moreover, putting this in front of all images like this includes all config reloaders without any notice.
> 
> Maybe you should add a dedicated note into the release to inform people about this change.
> 
> Regards,
> 
> Oxyno-zeta

 Sorry,  it's not desired behavior and breaking change was not expected. Soon will fix it at 0.26.2 release.

PR #536, opened 2022-10-02
Fix crash on Kubernetes 1.25

miketth:
On Kubernetes 1.25, the operator gets into a crash loop, because it tries to get policyv1beta1.PodDisruptionBudget, but it is removed in 1.25. This PR fixes that issue.

Since other resources don't have `Owns(&policyv1beta1.PodDisruptionBudget{})`, I opted to remove it. In case that is not the direction we want to go about it, feel free to change it.


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #544, opened 2022-10-27
{api/vmsingle,api/vmcluster}: add support of `vmbackupmanager` restore on pod start

zekker6:


api/v1beta1/vmcluster_types.go
```diff
@@ -729,6 +729,23 @@ type VMBackup struct {
...
 	// +optional
 	VolumeMounts []v1.VolumeMount `json:"volumeMounts,omitempty"`
+
+	// Restore Allows to enable restore options for pod
+	// Read more: https://docs.victoriametrics.com/vmbackupmanager.html#restore-commands
+	// +optional
+	Restore *VMRestore `json:"restore,omitempty"`
+}
+
+type VMRestore struct {
```
f41gh7, line 739: WDYT about adding boolean field -  `EnableRestoreCommands` to the `type VMBackup struct`?

 It should simplify logic a bit.

zekker6: This level of nesting was added in order to support the following future use-cases (without making breaking changes):
- ability to add separate extraArgs/extraEnvs for restore init container
- ability to run restore from point defined in k8s manifest (for example to create cluster and restore it from certain backup on start)
so that it will be possible to use something like:
```
restore:
  onStart:
    fromBackup: "some/backup/ref"
```

zekker6: If this seems to be redundant at this point please let me know - I'd be happy to simplify structs nesting here :)

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #551, opened 2022-11-07
controllers/factory: fix resizing of PVC for vmsingle

### LINKED ISSUE #409: Editing VMSingle storage size doesn't update the existing PVC
Editing the VMSingle storage size doesn't seem to change an existing PVC.  Is that intentional?  If so, in order to increase storage we would edit the PVC directly.  If that was not the intention, bug?

zekker6:
Fixes #409


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #552, opened 2022-11-08
controllers/factory: implement simpler sts revision tracking(#547)

### LINKED ISSUE #547: Update field currentRevision in StatefulSets created by operator
Hello!

Due to operator not updating `currentRevision` field in accordance with `updateRevision` after STS upgrades, kube-state-metrics's `kube_statefulset_status_replicas_current` metric is showing incorrect value.
Please consider to add this to operator's logic.

These 2 issues has more details:
https://github.com/kubernetes/kube-state-metrics/issues/1324
https://github.com/Opster/opensearch-k8s-operator/issues/310

Thanks!

zekker6:
fixes #547

### DIFF
--- a/controllers/factory/k8stools/sts.go
+++ b/controllers/factory/k8stools/sts.go
@@ -186,29 +186,7 @@ func performRollingUpdateOnSts(ctx context.Context, wasRecreated bool, rclient c
 		if err != nil {
 			return err
 		}
-		err = waitForPodReady(ctx, rclient, ns, pod.Name, c, func(pod *corev1.Pod) error {
-			// its special hack
-			// we check first pod revision label after re-creation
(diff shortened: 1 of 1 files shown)


f41gh7:
This hack is needed for case, when you delete Statefulset for re-creation ( when storage changed for instance). Kubernetes controller doesn't update Statefulset status fields in this case. It happens, when all pods are deleted and created again. If there was some issue with pod re-creation (timeout or any other), pods are re-recreated again and it falls into infinite loop.

f41gh7:
Thanks for implementation, was fixed at https://github.com/VictoriaMetrics/operator/commit/fc4abe6da99d73145130624745b46dafe4314b6e

Issue #554, opened 2022-11-15
duplicated 'alertmanager.yaml' fields when creating VMAlertmanagerConfig with nested Route


yamatcha:
I found duplicated alertmanager.yaml fields when creating vmalertmanagerconfig with nested Route.
Here is the example of vmalertmanagerconfig.

```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAlertmanagerConfig
metadata:
  name: example-email-web
spec:
  route:
    receiver: email
    group_interval: 1m
    routes:
      - receiver: email
        matchers:
          - severity = "critical"
      - continue: true
        receiver: slack
  receivers:
    - name: email
      email_configs:
        - to: some-email@example.com
          from: alerting@example.com
          smarthost: example.com:25
          text: ALARM
```

Then alertmanager.yaml is as shown below.
(2) is unexpectedly duplicated

```
$ kubectl get secret vmalertmanager-example-email-web-config -ojson | jq -r '.data."alertmanager.yaml"' | base64 -d
route:
  receiver: default-example-email-web-email
  routes:
  - routes:
    - matchers: (1)
      - severity = "critical"
      receiver: default-example-email-web-email
      continue: false
    - matchers: (2)
      - severity = "critical"
      receiver: default-example-email-web-email
      continue: false
    matchers:
    - namespace = "default"
    group_interval: 1m
    receiver: default-example-email-web-email
    continue: true
receivers:
- name: default-example-email-web-email
  email_configs:
  - from: alerting@example.com
    text: ALARM
    to: some-email@example.com
    smarthost: example.com:25
templates: []
```

Environment:
- VictoriaMetrics/Operator: Version 0.29.0

# Additional information

I supporse that it's due to a bug in `parseNestedRoutes`.
 
https://github.com/VictoriaMetrics/operator/blob/fc4abe6da99d73145130624745b46dafe4314b6e/api/v1beta1/vmalertmanagerconfig_types.go#L158-L173

This function parses `src.RawRotes` and add the corresponding route to `src.Routes`.
If this function is called twice, one route is added twice.

yamatcha:
I tried to fix #556

f41gh7:
Hello, I suppose, it's related with regression after Unmarshall changes, Validate call mutates CR.

 this change must fix an issue.

```
func (amc *VMAlertmanagerConfig) Validate() error {
	validateSpec := amc.DeepCopy()
[...]

PR #556, opened 2022-11-15
api/vmalertmanagerconfig: fix parseNestedRoutes

yamatcha:
Implements #554
The side effect of `parseNestedRoutes`  cause the problem, so I eliminate it and reassigned Route after parsing.

Signed-off-by: yamatcha <soju-yamashita@cybozu.co.jp>


f41gh7:
Was fixed at https://github.com/VictoriaMetrics/operator/commit/4a7e87368209e98de21cfb64d88eabf2da87f71f

 Thanks for example PR!

PR #576, opened 2023-01-01
Upgrade fsnotify to 1.6.0

yanggangtony:
Signed-off-by: yanggang <gang.yang@daocloud.io>

As [fsnotify v1.6.0](https://github.com/fsnotify/fsnotify/releases/tag/v1.6.0) released..
Follow it and improve the code .
Like 
```
if event.Op&Write == Write && !(event.Op&Remove == Remove) {
  }

--->

  if event.Has(Write) && !event.Has(Remove) {
  }
```

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -1,6 +1,6 @@
 module github.com/VictoriaMetrics/operator/api
 
-go 1.17
+go 1.18
 
 require (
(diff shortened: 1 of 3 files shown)


f41gh7:
Thanks for contribution!

PR #584, opened 2023-01-17
controllers/factory: do not set `-dedup.minScrapeInterval=1ms`

hagen1778:
Prevent setting `-dedup.minScrapeInterval=1ms` if
`replicationFactor` is 1. Even 1ms has an impact on cluster's performance and behavior. 
So it is better to avoid using it at no need.


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

Issue #592, opened 2023-02-08
alertmanager templates changed not trigger config reload 


okzheng:
I have configured a directory for storing Alertmanager templates. It was created as a ConfigMap mounted into `/etc/vm/configs/<configmap-name>`.  When I only modified these templates, the sidecar configmap-reload was not triggered to work.
I see the args of configmap-reload, it only watches a fixed directory `/etc/alertmanager/config`. I could not add another directory.
What could I do for triggering automatic Alertmanager reload when I change these templates?
Could vm-operator support a custom configmap-reload configuration?
Thanks.

okzheng:
Thanks~

f41gh7:
Since v0.31.0  `VMAlertmanager` supports templates reference. Need to add docs for it.

PR #596, opened 2023-02-27
Add support of vmalertmanager.spec.templates and autoreload dirs for templates and configmaps (#592)

Amper:
Issue #592:
- support of vmalertmanager.spec.templates field with autoreload support without pods restart
- support of autoreload for vmalertmanager.spec.configmaps in alertmanager without pods restart


controllers/factory/alertmanager.go
```diff
@@ -415,27 +449,25 @@ func makeStatefulSetSpec(cr *victoriametricsv1beta1.VMAlertmanager, c *config.Ba
...
+		Image: fmt.Sprintf("%s", formatContainerImage(c.ContainerRegistry, c.VMAlertManager.ConfigReloaderImage)),
+		Args: []string{
+			fmt.Sprintf("-webhook-url=%s", localReloadURL),
+			fmt.Sprintf("-volume-dir=%s", alertmanagerConfDir),
 		},
+		VolumeMounts:             crVolumeMounts,
+		Resources:                resources,
+		TerminationMessagePolicy: v1.TerminationMessageFallbackToLogsOnError,
+	}
+	if len(cr.Spec.ConfigMaps) > 0 {
```
f41gh7: does config-reloader recursively watches dir changes? probably we have to add TemplatesDir with cm name?
 
 As far as I know it uses fs.notify

```Are subdirectories watched too?
No, you must add watches for any directory you want to watch (a recursive watcher is on the roadmap: https://github.com/fsnotify/fsnotify/issues/18).
```

api/victoriametrics/v1beta1/additional.go
```diff
@@ -364,3 +364,11 @@ func (ds *DiscoverySelector) AsListOptions() (*client.ListOptions, error) {
 		LabelSelector: s,
 	}, nil
 }
+
+// ConfigMapKeyReference refers to a key in a ConfigMap.
```
f41gh7, line 368: It's better to use default kubernetes type `ConfigMapKeySelector`

Amper: @f41gh7 in the first version (where the files were formed by separate entries in Volumes[].Items and volumeMounts[]) i just used it, but after changes i became confused by the extra Optional field in this type, which is can't be used in any way

controllers/factory/alertmanager.go
Amper: It worked fine during testing, but it may have happened by accident, so I'll check carefully.

Amper: For some reason everything works for me, but you're right, it shouldn't work (it can depend on the environment), so i fixed

f41gh7 (approved):
LGTM

f41gh7:
Thanks!

Issue #508, opened 2022-08-08
Add podAntiAffinity when vmagent in shard mode


okzheng:
We have three k8s pod for deploying vm components. Then we config a vmagent cr with 2 shards, every shard with 2 replicas. 
When I add podAntiAffinity config for vmagent in shard mode, I could only add some labels to all sharding, and I could not add label shardNum. The result is all vmagent pods from a vmagent cr will be anti affinity.
It is unnecessary avoid placing vmagents from different shard on a node that match same conditions. We just need to avoid scheduling vmagents pod from a same shard to a node.
So if the vm-operator could add label shardNum to podAntiAffinity automatically or provide a flag to control the action, It would fit in more places.
May be have more reasonable resolution?

f41gh7:
Hello, operator adds `shard-num` label to the podSpec:
```
spec:
  template:
      metadata:
         labels:
            shard-num: "1"
```

 but it isn't enough to create proper antiAffinity rule. Need to think, how to implement it properly. Maybe some boolean flag, that will replace value for `shard-num` label with correct number at affinity section.

f41gh7:
Since v0.31.0  `VMAgent` supports placeholder for %SHARD_NUM%. It could be useful for affinity and pdb generation.

 Need to add docs for it.

PR #598, opened 2023-02-28
Add support "%SHARD_NUM%" placeholder for vmagent sts/deployment (#508)

Amper:


controllers/factory/k8stools/placeholders.go
```diff
@@ -0,0 +1,34 @@
...
+	"strings"
+)
+
+func RenderPlaceholders[T any](resource *T, placeholders map[string]string) (*T, error) {
+	data, err := json.Marshal(resource)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal resource for filling placeholders: %w", err)
+	}
+
+	if len(placeholders) == 0 || len(data) == 0 {
```
f41gh7: It could be moved into fast path, e.g.:

func Render...(arg [])
 if len(arg) == {
     return
  }
 In this case, we doesn't call Marshal and save some cpu and memory.

 Think, that check len(data) == 0 is not needed or we can check resource for nil at fast path

controllers/factory/k8stools/placeholders.go
```diff
@@ -0,0 +1,34 @@
...
+		return resource, nil
+	}
+
+	strData := string(data)
+	for placeholderName, value := range placeholders {
+		placeholder := strings.ToUpper(fmt.Sprintf("%%%s%%", placeholderName))
```
f41gh7: I don't think that we need such magic. Instead we can validate place holder placeholderName:
- it must starts and ends with %.

controllers/factory/k8stools/placeholders.go
```diff
@@ -0,0 +1,34 @@
+package k8stools
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+func RenderPlaceholders[T any](resource *T, placeholders map[string]string) (*T, error) {
```
f41gh7, line 12: Add some comment, for example

```
// RenderPlaceholders replaces placeholders at resource with given values
// placeholder must be in %NAME% format
// resource must be json serializable 
RenderPlaceholders()

```

controllers/factory/vmagent.go
```diff
@@ -142,13 +151,23 @@ func CreateOrUpdateVMAgent(ctx context.Context, cr *victoriametricsv1beta1.VMAge
 			}
 		}
 	} else {
+		// To save compatibility in the single-shard version still need to fill in placeholders
+		placeholders := map[string]string{"shard-num": "0"}
```
f41gh7: Lets create global var

 vmagentShardPlaceHolder = `%SHARD_NUM%` and refer it at docs.

 E.g. placeholders := map[string]string{ vmagentShardPlaceHolder: "0"}

f41gh7 (commented):
Overall looks good.

 I think it's fine to use placeholder only at vmagent. Since it's not needed anywhere else atm.

controllers/factory/k8stools/placeholders.go
Amper: fixed

controllers/factory/k8stools/placeholders.go
Amper: done

controllers/factory/k8stools/placeholders.go
Amper: fixed

controllers/factory/vmagent.go
Amper: done

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #613, opened 2023-03-28
Remove single quotes from remote write headers

axelsccp:
The operator adds single quotes to the headers inside remote write, resulting that the headers is identified as a whole string, not key:value, and when using this header to send to remote write server, the tenant that matches this header is not found.


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #623, opened 2023-04-07
VMAlertmanagerConfig: EmailConfig.require_tls always set to true


mas9612:
EmailConfig.require_tls always set to true, regardless of EmailConfig.require_tls value in a VMAlertmanagerConfig manifest.
https://docs.victoriametrics.com/operator/api.html#emailconfig

It seems that it is not checked in the buildEmail func.
https://github.com/VictoriaMetrics/operator/blob/933987a2a5fc3d6fe76b59b15f2e8c324d92ee22/controllers/factory/alertmanager/config.go#L799-L844

Therefore, global.smtp_require_tls value is always used, which is set to `true` by default.
https://prometheus.io/docs/alerting/latest/configuration/

Haleygo:
Looks like vm don't want us to set `require_tls` to false because [Go does not support unencrypted connections to remote SMTP endpoints](https://pkg.go.dev/net/smtp?utm_source=godoc#PlainAuth).
But the doc doesn't describe it clearly.
And if user need to set `require_tls` to false to test something, it will be a bug.

PR #624, opened 2023-04-10
alertmanagerConfig: support set require_tls to false

Haleygo:
fix https://github.com/VictoriaMetrics/operator/issues/623


Haleygo:
@f41gh7 PTAL :)

controllers/factory/alertmanager/config.go
```diff
@@ -798,13 +796,18 @@ func (cb *configBuilder) buildPagerDuty(pd operatorv1beta1.PagerDutyConfig) erro
 
 func (cb *configBuilder) buildEmail(email operatorv1beta1.EmailConfig) error {
 	var temp yaml.MapSlice
-	if email.TLSConfig != nil {
+	if email.RequireTLS != nil {
+		temp = append(temp, yaml.MapItem{Key: "require_tls", Value: *email.RequireTLS})
+	}
+	// not set tls_config if require_tls is false
+	if email.RequireTLS == nil || *email.RequireTLS {
```
f41gh7, line 808: It looks that this part could panic in cases:
 *email.RequireTLS == true and email.TLSConfig == nil.

 Probably, it's better to add sanity check
```go
if email.RequireTLS == nil ||  *email.RequireTLS {
   if email.TLSConfig == nil {
             return fmt.Errorf("incorrect email configuration, tls is required, but no config provided at spec")
   }
}
```

controllers/factory/alertmanager/config.go
```diff
@@ -574,7 +574,10 @@ func (cb *configBuilder) buildWebhook(wh operatorv1beta1.WebhookConfig) error {
 	}
 
 	temp = append(temp, yaml.MapItem{Key: "url", Value: url})
-	temp = append(temp, yaml.MapItem{Key: "max_alerts", Value: wh.MaxAlerts})
+
+	if wh.MaxAlerts != 0 {
```
Haleygo, line 578: only show maxalert in config when not 0

f41gh7 (approved):
LGTM

PR #625, opened 2023-04-10
vmalertmanagerConfig: fix webhook config maxAlerts not work

Haleygo:


f41gh7:
Thanks for contribution!

PR #640, opened 2023-05-08
api/podscrape: adds FilterRunning option as prometheus does

f41gh7:
it prevents scrapping metrics from completed pods, usually a cronJob pods. It's enabled by default.
updates generated-client


Amper (approved):
LGTM

Issue #641, opened 2023-05-09
how to use vmoperator without cluster scope permissions


SaschaP95:
Hello together

I'm working in a bigger company with a K8s as a Service platform. We order a K8s cluster and get a managed one.
This works fine but has limited permissions and dont allow everything.
I my case its only allowed to have more then get, list and watch permission on cluster scope level and in clusterroles.
Therefore i rewrite the vm-operator deployment to fulfill the RBAC concept.

I have clusterroles with get, list and watch permissions and a role with create, delete, etc. permission in each namespace.
I also configured the _WATCH_NAMESPACE_ env flag in the deployment.

Now the problem starts. I can deploy thats no problem but I get some errors in the vm-operator logs.

[...]

PR #647, opened 2023-05-18
operator: allows to properly run operator with single namespace

f41gh7:
disables cluster wide kubernetes API call with WATCH_NAMESPACE set for specific namespace. It allows to securely use operator without need to assign additional permissions for it. https://github.com/VictoriaMetrics/operator/issues/641

### DIFF
--- a/controllers/factory/selectors.go
+++ b/controllers/factory/selectors.go
@@ -22,7 +22,8 @@ func getNSWithSelector(ctx context.Context, rclient client.Client, nsSelector, o
 	// for each namespace apply list with  selector
 	// combine result
 	switch {
-	case nsSelector == nil:
+	// in single namespace mode, return object ns
+	case nsSelector == nil || watchNS != "":
 		namespaces = append(namespaces, objNS)
 	default:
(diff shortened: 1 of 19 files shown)


controllers/factory/vmuser.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
+	"github.com/VictoriaMetrics/operator/internal/config"
```
Amper: incorrect import order

controllers/factory/selectors.go
```diff
@@ -65,13 +66,7 @@ func visitObjectsWithSelector(ctx context.Context, rclient client.Client, ns []s
 		return nil
 	}
 
-	watchNamespace := config.MustGetWatchNamespace()
```
Amper, line 68: It seems to me that MustGetWatchNamespace doesn't work the way you expect. ListOptions doesn't filter the list of namespaces by namespace - as a result all namespaces will always be in the list.

Issue #648, opened 2023-05-18
VM operator aborts processing CRDs mid-way


sachinm24:
I'm using victoriametrics-operator to manage vmagents on kubernetes cluster. I have servicemonitor->vmservicescrape conversion enabled in all namespace.
If a servicemonitor config is not correct, VM operator aborts processing in the middle. So, if it discovered servicemonitors A, B, C, D and found an issue like missing field/incorrect field type in servicemonitor C, it aborts and does not create vmservicescrape for servicemonitor D.

This has serious implications -- a single incorrect service monitor can disable metrics scraping for many other important targets in the cluster. We also suspect that it removes the original vmservicescrape created from servicemonitor D

f41gh7:
Hello,  it shouldn't happen in any case. Operator must skip broken ServiceMonitor processing, log error and move to the next object. Need to investigate this case. Could you provide some operator logs with errors?

 Also, it doesn't delete any ServiceMonitors or PodMonitors. It's responsibility of cluster administrator. 


 Only one possible case, when operator creates VMServiceScrape from ServiceMonitor and it contains some incorrect configuration (like missing secret reference). Operator may stop processing config generation for `VMAgent`. This issue must be addressed and fixed at the next release.

f41gh7:
Version v0.34.0 contains a potential fix for this issue - it skips misconfigured scrape objects for `vmagent`.

sachinm24:
> Version v0.34.0 contains a potential fix for this issue - it skips misconfigured scrape objects for `vmagent`.

thanks, we are currently using v0.20.1 currently. Will try it with upgrade

hagen1778:
Closing as stale. Feel free to re-open

PR #651, opened 2023-05-23
controllers/vmagent: skips misconfiguration for user defined objects

f41gh7:
End users may misconfigure VMPodScrape, VMServiceScrape or other object. And it breaks entire VMAgent service until admin fixes it. With those changes such objects will be skipped and VMAgent continues it's work. https://github.com/VictoriaMetrics/operator/issues/648

### DIFF
--- a/controllers/factory/scrapes.go
+++ b/controllers/factory/scrapes.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	stderrors "errors"
 	"fmt"
 	"strings"
 
@@ -70,7 +71,8 @@ func CreateOrUpdateConfigurationSecret(ctx context.Context, cr *victoriametricsv
 
 	ssCache, err := loadScrapeSecrets(ctx, rclient, sScrapes, nodes, pScrapes, probes, statics, cr.Spec.APIServerConfig, cr.Spec.RemoteWrite, cr.Namespace)
 	if err != nil {
-		if _, ok := err.(*utils.ErrGroup); !ok {
+		var ge *utils.ErrGroup
+		if !stderrors.As(err, &ge) {
 			return nil, fmt.Errorf("cannot load scrape target secrets for api server or remote writes: %w", err)
 		}
 		vmagentSecretFetchErrsTotal.Inc()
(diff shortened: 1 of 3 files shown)


controllers/factory/scrapes.go
```diff
@@ -91,11 +120,13 @@ func CreateOrUpdateConfigurationSecret(ctx context.Context, cr *victoriametricsv
...
-	if errors.IsNotFound(err) {
-		log.Info("creating new configuration secret for vmagent")
-		return ssCache, rclient.Create(ctx, s)
+	curSecret := &corev1.Secret{}
+	if err := rclient.Get(ctx, types.NamespacedName{Namespace: cr.Namespace, Name: s.Name}, curSecret); err != nil {
+		if errors.IsNotFound(err) {
+			log.Info("creating new configuration secret for vmagent")
+			return ssCache, rclient.Create(ctx, s)
+		}
+		return nil, fmt.Errorf("cannot ")
```
Amper: It seems to me that the text of the error is incomplete

controllers/factory/scrapes.go
```diff
@@ -54,9 +68,24 @@ func CreateOrUpdateConfigurationSecret(ctx context.Context, cr *victoriametricsv
 		return nil, fmt.Errorf("selecting PodScrapes failed: %w", err)
 	}
 
-	ssCache, err := loadScrapeSecrets(ctx, rclient, smons, nodes, pmons, probes, statics, cr.Spec.APIServerConfig, cr.Spec.RemoteWrite, cr.Namespace)
+	ssCache, err := loadScrapeSecrets(ctx, rclient, sScrapes, nodes, pScrapes, probes, statics, cr.Spec.APIServerConfig, cr.Spec.RemoteWrite, cr.Namespace)
 	if err != nil {
-		return nil, fmt.Errorf("cannot load basic secrets for ServiceMonitors: %w", err)
+		if _, ok := err.(*utils.ErrGroup); !ok {
```
Amper: I think it's can be too confusing, maybe it would be better to create specialized error for that (based on utils.ErrGroup) + check with errors.Is?

f41gh7: Changed with errors.As

Issue #658, opened 2023-06-02
vmalert has two default port


yimenghua:
internal]/config/config.go
VMAlertDefault.Port : 8080

api/v1beta1/vmalert_types.go
```go
func (cr *VMAlert) AsURL() string {
	port := cr.Spec.Port
	if port == "" {
		port = "8880"
	}
	return fmt.Sprintf("http://%s.%s.svc:%s", cr.PrefixedName(), cr.Namespace, port)
}
```

8080 or 8880 ?

f41gh7:
Hello, it's a bug. Thx for pointing. It affects only service discovery for vmauth.

f41gh7:
Fixed at [v0.35.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.35.0) release

PR #659, opened 2023-06-05
fix vmalert default port

Haleygo:
fix https://github.com/VictoriaMetrics/operator/issues/658
@f41gh7 Feel free to close this if you already fixed :)

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -21,10 +21,8 @@ var (
 	initNamespace sync.Once
 )
 
-const (
-	prefixVar         = "VM"
-	UnLimitedResource = "unlimited"
-)
+const prefixVar = "VM"
+const UnLimitedResource = "unlimited"
 
 // WatchNamespaceEnvVar is the constant for env variable WATCH_NAMESPACE
 // which specifies the Namespace to watch.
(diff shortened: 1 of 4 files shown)


internal/config/config.go
```diff
@@ -52,7 +54,7 @@ type BaseOperatorConf struct {
 	VMAlertDefault            struct {
 		Image               string `default:"victoriametrics/vmalert"`
 		Version             string `default:"v1.91.0"`
-		Port                string `default:"8080"`
+		Port                string `default:"8880"`
```
f41gh7: I think, it's better to keep default port `8080` as is. And fix `cr *VMAlert) AsURL()`.

 This change may break configuration for some users, that configured `vmalert.proxyURL` param.  For instance `k8s-stack` uses it.

Haleygo: Then we need to update the default port[8880->8080] vmalert listening by default in [VictoriaMetrics/VictoriaMetrics ](https://github.com/search?q=repo%3AVictoriaMetrics%2FVictoriaMetrics+8880&type=code), right?

f41gh7: No, it could be different ports. I believe, we could sync it with `vmalert` defaults values before `v1.0.0` operator release.

 The same happened to `VMSingle`. Operator configures it to listen `:8429` port, but default port is `:8428`.

Haleygo: Got it, fixed :)

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #665, opened 2023-06-10
VMAgent skip invalid scheme configuration


f41gh7:
```sh
2023-06-09T08:25:59.827Z fatal VictoriaMetrics/lib/promscrape/scraper.go:117 cannot read "/etc/vmagent/config_out/vmagent.env.yaml": cannot parse Prometheus config from "/etc/vmagent/config_out/vmagent.env.yaml": cannot parse `scrape_config`: unexpected `scheme` for `job_name` "serviceScrape/XXX/YYY/0": "tcp"; supported values: http or https
```
During config generation, operator has to validate `scheme` param and skip invalid objects with logging an error.

PR #666, opened 2023-06-10
vmagent:fix validation when generate static scrape config

Haleygo:
fix https://github.com/VictoriaMetrics/operator/issues/665


controllers/factory/static_targets_build.go
```diff
@@ -66,6 +68,9 @@ func generateStaticScrapeConfig(
 		cfg = append(cfg, yaml.MapItem{Key: "params", Value: ep.Params})
 	}
 	if ep.Scheme != "" {
```
f41gh7: I think about adding validation at `UnmarshalJSON` method. `Scheme` could be wrapped into custom type.
 It allows to validate it with `AdmissionWebhook` and fail early. 

 It's already done for alertmanager config:
https://github.com/VictoriaMetrics/operator/blob/master/api/v1beta1/vmalertmanagerconfig_types.go#L193

Haleygo: Updated, using kubebuilder CRD validation markers now.

api/v1beta1/vmstaticscrape_types.go
```diff
@@ -33,6 +35,7 @@ type TargetEndpoint struct {
 	Path string `json:"path,omitempty"`
 	// HTTP scheme to use for scraping.
 	// +optional
+	// +kubebuilder:validation:Enum=http;https
```
f41gh7, line 37: Could you also add it to the other Scheme definitions at
`VMPodScrape`, `VMServiceScrape`, `VMNodeScrape`.

Haleygo: Sure, done.
And I will check other fields to see if more checks can be done in advance. Maybe replace some sanity checks in webhook.
https://github.com/VictoriaMetrics/operator/blob/ea0c0cefec8bdae4c42bd5f0148da6e3f1889b5f/api/v1beta1/vmagent_types.go#L175-L180
https://github.com/VictoriaMetrics/operator/blob/ea0c0cefec8bdae4c42bd5f0148da6e3f1889b5f/api/v1beta1/vmagent_webhook.go#L52-L55

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #619, opened 2023-04-03
download configuration with init containers


f41gh7:
For vmauth/vmagent we have to init configuration before service start. It could be done with initContainers:

```yaml
kind: VMAgent
spec:
 initContainers:
  - args:
    - -c
    - gunzip -c /etc/vmagent/config/vmagent.yaml.gz > /etc/vmagent/config_out/vmagent.env.yaml
    command:
    - /bin/sh
    image: quay.io/prometheus-operator/prometheus-config-reloader:v0.58.0
    name: config-init
    volumeMounts:
    - mountPath: /etc/vmagent/config
      name: config
    - mountPath: /etc/vmagent/config_out
      name: config-out
```
 But it wouldn't work for custom config reloader, since it download configuration directly from secret/configmap.

 Add oneshot config download mode for config-reloader and automatically add init container with gunzip command for `prometheus-config-reloader`.

FrederikNS:
@f41gh7

I recently tried out operator v0.33.0, which seems to include commit you reference here, but deploying a vmagent seems broken with this change.

The init container tries to gunzip the file `/etc/vmagent/config/vmagent.yaml`, while the actual secret contains a file called `vmagent.yaml.gz`.

So I think there's regression according to the filenames in the actual secret created by the operator, and the filename the init container expects to gunzip.

My VMAgent object simply looks like this:

    apiVersion: operator.victoriametrics.com/v1beta1
    kind: VMAgent
    metadata:
      name: victoria-metrics-agent
    spec:
      selectAllByDefault: true
      remoteWrite:
      - url: "http://vmsingle-victoria-metrics:8429/api/v1/write"

And the init container for vmagent simply throws:

    gunzip: /etc/vmagent/config/config.yaml: No such file or directory

f41gh7:
@FrederikNS  thanks for pointing. It was fixed at related commit and will be a part of next patch release

FrederikNS:
Thank you!

f41gh7:
Finished at [v0.35.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.35.0) release

PR #667, opened 2023-06-15
vmauth/vmagent:implement configuration initiation for custom config reloader

Haleygo:
implement https://github.com/VictoriaMetrics/operator/issues/619
1. add one-shot mode for config-reloader
2. use one-shot mode in initContainer if `VM_CUSTOMCONFIGRELOADERIMAGE` enabled

### DIFF
--- a/internal/config-reloader/main.go
+++ b/internal/config-reloader/main.go
@@ -25,7 +25,7 @@ import (
 )
 
 var (
-	mode                   = flag.String("mode", "", "default is normal. If using one-shot mode, will read config and write to config-envsubst-file once")
+	onlyInitConfig         = flag.Bool("only-init-config", false, "enables will read config and write to config-envsubst-file once before exit")
 	configFileName         = flag.String("config-file", "", "config file watched by reloader")
 	configFileDst          = flag.String("config-envsubst-file", "", "target file, where conent of configFile or configSecret would be written")
 	configSecretName       = flag.String("config-secret-name", "", "name of kubernetes secret in form of namespace/name")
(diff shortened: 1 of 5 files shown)


internal/config-reloader/main.go
```diff
@@ -54,6 +55,12 @@ func main() {
 	}
 
 	configWatcher.startWatch(ctx, updatesChan)
+	if *mode == "one-shot" {
+		logger.Infof("config-reloader one-shot mode completed, exit now")
```
Haleygo: If the process[download,parse,write] failed in one-shot mode, should we exit it with code nonzero?

internal/config-reloader/main.go
```diff
@@ -25,6 +25,7 @@ import (
 )
 
 var (
+	mode                   = flag.String("mode", "", "default is normal. If using one-shot mode, will read config and write to config-envsubst-file once")
```
Haleygo: Not sure about the mode name too :)

Haleygo:
@f41gh7 
So the new mode only works in new version of config-reloader, should I add some version control?

internal/config-reloader/main.go
f41gh7: I think, boolean flag should be enough  `only-init-config`.

internal/config-reloader/main.go
f41gh7: Yep, it should exit with non zero code.

internal/config-reloader/main.go
```diff
@@ -236,7 +245,7 @@ func writeNewContent(data []byte) error {
 		}
 	}
 	tmpDst := *configFileDst + ".tmp"
-	if err := os.WriteFile(tmpDst, data, 0644); err != nil {
+	if err := os.WriteFile(tmpDst, data, 0o644); err != nil {
```
f41gh7: is it a typo?

f41gh7 (commented):
Looks good!

f41gh7:
> @f41gh7
> So the new mode only works in new version of config-reloader, should I add some version check in operator or modify the default CustomConfigReloaderImage?

 Yes, it makes sense. I think check for v0.35.0 release is needed. 

 If version cannot be parsed (user defined own docker image with incorrect version name), assume that config-reloader supports new mode.

internal/config-reloader/main.go
Haleygo: Sorry, my format tool did that.

f41gh7 (approved):
LGTM

PR #672, opened 2023-06-19
vmuser: remove rate limit on delete

Haleygo:
If a lot of vmusers got deleted in short term, it will be rate limited now. 
So most of them will stay there until manager resync or operator restart, along with their secret.


f41gh7 (approved):
LGTM

Issue #670, opened 2023-06-18
VMSingle add status field


f41gh7:
It will be great to have an actual status of pods for single node installation and tracks it's availability.

 It could use the same status as cluster has and poll for pod status (health/unhealthy) during upgrades and periodic resync.

 Main goal - simplify `VMSingle` usage for programmatic access, user could subscribe for `vmsingle.status` and make some action of it's not healthy/upgrade failed instead of listing pods.

f41gh7:
Added at [v0.35.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.35.0) release

PR #673, opened 2023-06-20
vmsingle: add status field

Haleygo:
https://github.com/VictoriaMetrics/operator/issues/670
Status flow:
changes detected -> status expanding
CreateOrUpdateVMSingle returned an error -> status failed with reason.
CreateOrUpdateVMSingle and other functions returned no error -> status operational and update lastAppliedSpec.

### DIFF
--- a/controllers/vmsingle_controller.go
+++ b/controllers/vmsingle_controller.go
@@ -124,7 +124,7 @@ func (r *VMSingleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 	instance.Status.Reason = ""
 	instance.Status.SingleStatus = victoriametricsv1beta1.SingleStatusOperational
 	if err := r.Client.Status().Update(ctx, instance); err != nil {
-		return result, fmt.Errorf("cannot update single status : %w", err)
+		return result, fmt.Errorf("cannot update single status: %w", err)
 	}
 
 	if specChanged {
--- a/controllers/vmsingle_controller.go
+++ b/controllers/vmsingle_controller.go
@@ -87,9 +87,17 @@ func (r *VMSingleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 		if err != nil {
 			return result, fmt.Errorf("failed to get deployment for vmsingle %s: %w", req.NamespacedName, err)
 		}
-		if currentDeploy.Status.AvailableReplicas == currentDeploy.Status.Replicas {
+
+		instance.Status.ReadyReplicas = currentDeploy.Status.ReadyReplicas
+		instance.Status.Replicas = currentDeploy.Status.Replicas
+		instance.Status.UpdatedReplicas = currentDeploy.Status.UpdatedReplicas
+		instance.Status.AvailableReplicas = currentDeploy.Status.AvailableReplicas
+		instance.Status.UnavailableReplicas = currentDeploy.Status.UnavailableReplicas
+		if instance.Status.ReadyReplicas == instance.Status.UpdatedReplicas && instance.Status.ReadyReplicas == instance.Status.Replicas {
+			instance.Status.Reason = ""
 			instance.Status.SingleStatus = victoriametricsv1beta1.SingleStatusOperational
 		} else {
+			instance.Status.Reason = "not all pods are updated and ready"
 			instance.Status.SingleStatus = victoriametricsv1beta1.SingleStatusFailed
 		}
 		if err := r.Client.Status().Update(ctx, instance); err != nil {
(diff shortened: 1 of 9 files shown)


controllers/vmsingle_controller.go
```diff
@@ -73,6 +75,44 @@ func (r *VMSingleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 	if instance.Spec.ParsingError != "" {
 		return handleParsingError(instance.Spec.ParsingError, instance)
 	}
+	lastAppliedSingleSpec, err := instance.GetLastAppliedSpec()
+	if err != nil {
+		reqLogger.Error(err, "cannot parse last applied single spec")
+	}
+	singleChanges := deep.Equal(lastAppliedSingleSpec, &instance.Spec)
+	if len(singleChanges) == 0 {
```
f41gh7: I think, we shouldn't skip reconciliation for single. Since, `VMSingle` object could be updated without changing an annotation. By kubectl edit for instance.

 Tracking changes is needed for changing object status.

 E.g. if changes detected, update status for `expanding`.

controllers/vmsingle_controller.go
```diff
@@ -90,6 +130,11 @@ func (r *VMSingleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 
 	_, err = factory.CreateOrUpdateVMSingle(ctx, instance, r, r.BaseConf)
```
f41gh7, line 102: It would be great to add pod status check with `waitExpanding` at  `CreateOrUpdateVMSingle` function.

controllers/vmsingle_controller.go
f41gh7: Status flow:
1) changes detected -> status `expanding`
2) CreateOrUpdateVMSingle returned an error -> status `failed` with reason.
3) CreateOrUpdateVMSingle and other functions returned no error -> status `operational` and update lastAppliedSpec.

controllers/vmsingle_controller.go
```diff
@@ -73,6 +75,44 @@ func (r *VMSingleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
...
+	singleChanges := deep.Equal(lastAppliedSingleSpec, &instance.Spec)
+	if len(singleChanges) == 0 {
+		// only update status by deployment pod status if single has no change
+		var currentDeploy appsv1.Deployment
+		err := r.Client.Get(ctx, types.NamespacedName{Name: instance.PrefixedName(), Namespace: instance.Namespace}, &currentDeploy)
+		if err != nil {
+			return result, fmt.Errorf("failed to get deployment for vmsingle %s: %w", req.NamespacedName, err)
+		}
+
+		instance.Status.ReadyReplicas = currentDeploy.Status.ReadyReplicas
```
f41gh7: I think, we could skip Replicas fields changing, `VMSingle` always has only 1 replica.

controllers/vmsingle_controller.go
```diff
@@ -104,6 +149,20 @@ func (r *VMSingleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (r
 			reqLogger.Error(err, "cannot create serviceScrape for vmsingle")
 		}
 	}
+
+	specPatch, err := instance.LastAppliedSpecAsPatch()
```
f41gh7: It's better to apply patch only if len(singleChanges) > 0

f41gh7 (changes requested):

controllers/vmsingle_controller.go
Haleygo: > I think, we shouldn't skip reconciliation for single. Since, `VMSingle` object could be updated without changing an annotation. By kubectl edit for instance.
> 
> Tracking changes is needed for changing object status.
> 
> E.g. if changes detected, update status for `expanding`.

kubectl edit instance will make the `len(singleChanges)>0`, doesn't it?
I think if there is no singleChanges, means the single.spec has no change. Since annotation only update when there is a change and reconcil is done in here.
https://github.com/VictoriaMetrics/operator/blob/2771c947528850e6b8cd83d493d5246a8e04abce/controllers/vmsingle_controller.go#L153-L160
So I only check the pod status to update the status of single, and return anyway to skip the rest of reconcil.

>CreateOrUpdateVMSingle and other functions returned no error -> status operational and update lastAppliedSpec.

I have different `operational` flow here, only mark single as  `operational` when readyReplicas == updatedReplicas == wantedReplicas.
Maybe it's too strict then.

controllers/vmsingle_controller.go
Haleygo: Oh, I stored ready pod replicas of deployment here to see if single is `operational`.

controllers/vmsingle_controller.go
Haleygo: But we don't know how long pod will take to be ready[ apply for ip, pulling image,etc] or even created[there could be some webhook to block pod creation in cluster], so the timeout is ambiguous. 
So I thought only do the pod status check in extra reconcil.

Updated: saw the waitExpanding logic under vmcluster, will update:)

controllers/vmsingle_controller.go
Haleygo: Notice this note "if you need more - use vm cluster" until now :)
Fixed, not adding `ReadyReplicas` fields and keep rest replicas untouched for backward compatibility.

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #677, opened 2023-06-23
VMCluster status is always "expanding"


vozerov:
Hey everyone!

Got some error with Victoria Metrics operator. I am using latest version (victoriametrics/operator:v0.34.1). 

My VMCluster looks like:

```sh
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"operator.victoriametrics.com/v1beta1","kind":"VMCluster","metadata":{"annotations":{},"labels":{"app.kubernetes.io/instance":"vm","app.kubernetes.io/managed-by":"Helm","app.kubernetes.io/name":"victoria-metrics-k8s-stack","app.kubernetes.io/version":"v1.91.2","helm.sh/chart":"victoria-metrics-k8s-stack-0.16.3"},"name":"vm-victoria-metrics-k8s-stack","namespace":"p-platform"},"spec":{"imagePullSecrets":[],"replicationFactor":1,"retentionPeriod":"14","vminsert":{"extraArgs":{},"image":{"tag":"v1.91.2-cluster"},"nodeSelector":{"usecase":"system"},"replicaCount":2,"resources":{}},"vmselect":{"cacheMountPath":"/select-cache","extraArgs":{"vmalert.proxyURL":"http://vmalert-vm-victoria-metrics-k8s-stack.p-platform.svc:8080"},"image":{"tag":"v1.91.2-cluster"},"nodeSelector":{"usecase":"system"},"replicaCount":2,"resources":{},"storage":{"volumeClaimTemplate":{"spec":{"resources":{"requests":{"storage":"10Gi"}}}}}},"vmstorage":{"image":{"tag":"v1.91.2-cluster"},"nodeSelector":{"usecase":"system"},"replicaCount":2,"resources":{},"storage":{"volumeClaimTemplate":{"spec":{"resources":{"requests":{"storage":"50Gi"}}}}},"storageDataPath":"/vm-data"}}}
[...]

PR #679, opened 2023-06-23
vmcluster: fix spec change check

Haleygo:
https://github.com/VictoriaMetrics/operator/issues/677
Previous check will fail when there are fields like `imagePullSecrets: []` been specified in spec, they will be omitted by json.marshall. That will cause constant status update and reconciliation for vmcluster, and possible resource update conflicts.

### DIFF
--- a/api/v1beta1/vmcluster_types.go
+++ b/api/v1beta1/vmcluster_types.go
@@ -1,6 +1,7 @@
 package v1beta1
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"path"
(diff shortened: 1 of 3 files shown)


api/v1beta1/vmcluster_types.go
```diff
@@ -948,17 +952,16 @@ func (cr *VMCluster) LastAppliedSpecAsPatch() (client.Patch, error) {
 	return client.RawPatch(types.MergePatchType, []byte(patch)), nil
 }
 
-// GetLastAppliedSpec returns last applied cluster spec
-func (cr *VMCluster) GetLastAppliedSpec() (*VMClusterSpec, error) {
+// CompareSpecChange compares cluster spec with last applied cluster spec stored in annotation
+func (cr *VMCluster) CompareSpecChange() ([]string, error) {
```
f41gh7: WDYT about changing this function for:
```go
func (cr *VMCluster) HasSpecChanges() (bool, error)
```

api/v1beta1/vmcluster_types.go
```diff
@@ -948,17 +952,16 @@ func (cr *VMCluster) LastAppliedSpecAsPatch() (client.Patch, error) {
...
+	lastAppliedClusterJSON := cr.Annotations["operator.victoriametrics/last-applied-spec"]
+	if err := json.Unmarshal([]byte(lastAppliedClusterJSON), &prevClusterSpec); err != nil {
+		return []string{""}, fmt.Errorf("cannot parse last applied cluster spec value: %s : %w", lastAppliedClusterJSON, err)
 	}
-	if err := json.Unmarshal([]byte(prevClusterJSON), &prevClusterSpec); err != nil {
-		return nil, fmt.Errorf("cannot parse last applied cluster spec value: %s : %w", prevClusterJSON, err)
-	}
-	return &prevClusterSpec, nil
+	instanceSpecData, _ := json.Marshal(cr.Spec)
+	changes := deep.Equal([]byte(lastAppliedClusterJSON), instanceSpecData)
```
f41gh7: Probably, we could use just simple `bytes.Compare` or check string equality, since `deep.Equal` checks structs.

f41gh7 (commented):
Thanks for quick fix!

api/v1beta1/vmcluster_types.go
Haleygo: Sounds good, updated.

f41gh7 (approved):
LGTM

api/v1beta1/vmcluster_types.go
```diff
@@ -1,18 +1,22 @@
...
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
-	"k8s.io/apimachinery/pkg/types"
 	"path"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"strings"
 
+	"github.com/go-test/deep"
```
f41gh7: Linter is unhappy with unused import

f41gh7:
Thanks!

Issue #637, opened 2023-05-01
Set readOnlyRootFilesystem to true for all containers


timp87:
My proposal is to set (hardcode) readOnlyRootFilesystem to true by operator for all containers [securityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#securitycontext-v1-core). For example prometheus-operator currently deploys the stack with hardcoded readOnlyRootFilesystem in container's securityContext. I. e. it hardcodes this option for all resources deployed withing CRs.

I find this option as a nice security measure and in my experience many companies enforce this setting to be on for all containers.

f41gh7:
Also it'd be great to add `PodSecurityContext` https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod.

 It has really nice feature `fsGroupChangePolicy`. https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods

```
By default, Kubernetes recursively changes ownership and permissions for the contents of each volume to match the fsGroup specified in a Pod's securityContext when that volume is mounted. For large volumes, checking and changing ownership and permissions can take a lot of time, slowing Pod startup.
```

 It should simplify managing volume permissions. 

 cc @Amper  @hagen1778

uhthomas:
I agree, this was also my proposal in https://github.com/VictoriaMetrics/operator/pull/692#issuecomment-1648256933

Haleygo:
PodSecurityContext already added in https://github.com/VictoriaMetrics/operator/pull/692.
As for `fsGroupChangePolicy: OnRootMismatch`, I need to do some research on default fs group for pod volume, see if it's necessary here.

Haleygo:
[Strict SecurityContext](https://github.com/VictoriaMetrics/operator/blob/master/internal/config/config.go#L271) will be added by default since [v0.36.0 ](https://github.com/VictoriaMetrics/operator/releases/tag/v0.36.0), users can disable it by setting `VM_ENABLESTRICTSECURITY=false`.
If SecurityContext already defined before for pod or containers, the default Strict SecurityContext won't be applied too.

Close this as completed.

PR #682, opened 2023-06-27
feat: Forbid write access to root filesystem

lvyanru8200:
set vmalert container SecurityContext the value of ReadOnlyRootFilesystem is set to true

fix: https://github.com/VictoriaMetrics/operator/issues/637

### DIFF
--- a/controllers/factory/vmalert.go
+++ b/controllers/factory/vmalert.go
@@ -3,6 +3,7 @@ package factory
 import (
 	"context"
 	"fmt"
+	"k8s.io/utils/pointer"
 	"path"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


lvyanru8200:
cc @f41gh7

lvyanru8200:
Should I make changes to the individual components of VictoriaMetrics as one pr or multiple

Haleygo:
> Should I make changes to the individual components of VictoriaMetrics as one pr or multiple

I think it's ok to add them all in this pr.

f41gh7:
I think, it's better to make this behavior configurable:
1) global setting, that enables security context for all components (non-root user, read-only root fs). With `BaseOperatorConfig.enableStrictSecurity`
2) new setting must respect user define configuration ( if read-only root fs disabled for some reason).

Something like:

```go
func addStrictSecuritySettingsToPod(p *v1.PodSpec, c *config.BaseOperatorConfig){
    /// fast path, disabled configuration or user controls it
    if !c.enableStrictSecurity || p.SecurityContext != nil{
          return 
   }
   // apply global security context
   // like FSGuuiD
   p.SecurityContext = &corev1.PodSecurityContext{
            RunAsNonRoot: true,
            RunAsUser: some_global_var,
            RunAsGroup: some_global_var,
            FSGroup: some_global_var,
   }
   for idx := range p.Containers {
       container := &p.Containers[idx]
       container.SecurityContext = &corev1.SecurityContext{
              ReadOnlyRootFilesystem: True,
      }
   }
}

```

Amper:
Hey @lvyanru8200, thank you very much for your contribution.
We have not received any responses to comments and have decided to finalize your work in a [separate PR](https://github.com/VictoriaMetrics/operator/pull/692). This PR will be closed.

PR #692, opened 2023-07-18
Enable default strict security context

Haleygo:
implement https://github.com/VictoriaMetrics/operator/issues/637
Add default strict security contexts for pods and containers.


controllers/factory/builders.go
```diff
@@ -525,3 +523,32 @@ func formatContainerImage(globalRepo string, containerImage string) string {
...
 	return globalRepo + containerImage[len("quay.io/"):]
 }
+
+func addStrictSecuritySettingsToPod(p *v1.PodSecurityContext, enableStrictSecurity bool) *v1.PodSecurityContext {
+	if !enableStrictSecurity || p != nil {
+		return p
+	}
+	return &v1.PodSecurityContext{
+		RunAsNonRoot: pointer.Bool(true),
+		// 'nobody' has uid `65534' in alpine image
```
Amper: It's worth noting that in addition to alpline-based VM images, we also use:

- `quay.io/prometheus-operator/prometheus-config-reloader` based on busybox
- `jimmidyson/configmap-reload` based on debian
- `prom/alertmanager` based on busybox

Fortunately they all use `uid=65534` for nobody as well, so it seems to me that adding new parameters for such a small thing would be unnecessary.

However, we do have the ability to override images in the operator, so it seems to me that:

- It's worth correcting this comment a bit since it's not just about alpine.
- It should be written in the documentation about exactly what enabling the parameter  `EnableStrictSecurity` does. In particular about the use of `uid=65534`. Also we should specify that "if you use alternative images, you will need to either use an image with the same uid, or specify security settings explicitly".

Amper (dismissed):
Overall looks great, but see one comment

controllers/factory/builders.go
Haleygo: >Fortunately they all use uid=65534 for nobody as well

Yeah, '65534' for 'nobody' looks like a reserved one.

Fixed the comments, thanks)

Amper (commented):
LGTM

Amper (approved):
LGTM

uhthomas:
I already set my own security context. Won't these changes override them?

https://github.com/uhthomas/automata/blob/b15f9282b37bd19d1c854a245ff929327f6f6f5c/k8s/unwind/vm/vm_cluster_list.cue#L42-L46

uhthomas:
Hmm, no, I see. It looks like it checks if the security context already exists. Thanks for making this change! Though, it might be nice to make a follow-up PR which makes it look more like:

https://github.com/uhthomas/automata/blob/b15f9282b37bd19d1c854a245ff929327f6f6f5c/k8s/unwind/vm/vm_cluster_list.cue#L22-L34

Amper:
Hey @uhthomas. 
Yes, it can be done at the helm-chart level.

uhthomas:
> Hey @uhthomas. 
> 
> Yes, it can be done at the helm-chart level.
> 
> 

Yes, I understand, but it would be nice to make the defaults even more sensible. It's currently missing the seccomp profile and should have a read only root file system.

PR #695, opened 2023-07-19
configure resource requests for config reloader container

blesswinsamuel:
As per the kubernetes docs (https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/), 

> If you specify a limit for a resource, but do not specify any request, and no admission-time mechanism has applied a default request for that resource, then Kubernetes copies the limit you specified and uses it as the requested value for the resource.

In the kubernetes cluster I use, there is an admission controller preventing the pods from getting created/updated if they don't have resource requests set. This change should help to fix the problem. Here is the error currently logged by victoriametrics operator that this commit should help fix:

```
cannot update deployment for app: vmagent-obs-vm, err: deployments.apps \"vmagent-obs-vm\" is forbidden: container config-reloader is missing resource requests values for all
```


Haleygo (approved):
LGTM!

Issue #699, opened 2023-07-23
[VMAlert] Wrong path for tlsCAFile when ca is from secret


orblazer:
Hello,
I trying to make my `VMAlert` work on my development env, so for that i need to define custom CA, but the operator seem to generate wrong path in arguments for `tlsCAFile`.

I create my `VMAlert` with this definition (this is partial with minimal parameters
```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAlert
metadata:
  name: vm-stack
  namespace: monitoring
spec:
  datasource:
    url: https://metrics.local.vm/
    tlsConfig:
      ca:
        secret:
          key: tls.crt
          name: vm-ca-tls-secret
  notifiers:
    - url: https://metrics.local.vm:9093/
      tlsConfig:
        ca:
          secret:
            key: tls.crt
            name: vm-ca-tls-secret
  remoteRead:
    url: https://metrics.local.vm/
    tlsConfig:
      ca:
        secret:
          key: tls.crt
          name: vm-ca-tls-secret
  remoteWrite:
    url: https://metrics.local.vm/api/v1/write
    tlsConfig:
      ca:
        secret:
          key: tls.crt
          name: vm-ca-tls-secret
```
And this is generated `args` in deployment :
```yaml
args:
- '-datasource.tlsCAFile=monitoring_vm-ca-tls-secret_tls.crt'
- '-datasource.url=https://metrics.local.vm/'
- '-external.label=cluster=salamandre'
- '-httpListenAddr=:8080'
- '-notifier.tlsCAFile=/etc/vmagent-tls/certs/monitoring_vm-ca-tls-secret_tls.crt'
- '-notifier.url=https://metrics.local.vm:9093/'
- '-remoteRead.tlsCAFile=monitoring_vm-ca-tls-secret_tls.crt'
- '-remoteRead.url=https://metrics.local.vm/'
- '-remoteWrite.tlsCAFile=monitoring_vm-ca-tls-secret_tls.crt'
- '-remoteWrite.url=https://metrics.local.vm/api/v1/write'
```

**BUT** the issue is `tlsCAFile` args missing prefix path of file, only `notifier.tlsCAFile` have right path.

So my current solution is to pass `caFile` argument in operator manifest with `/etc/vmagent-tls/certs/monitoring_vm-ca-tls-secret_tls.crt` as they value.

Thanks.

Amper:
[...]

PR #700, opened 2023-07-23
fix tlsCAFile args when using secret or configMap

Haleygo:
releated https://github.com/VictoriaMetrics/operator/issues/699


Amper (approved):
LGTM

Issue #702, opened 2023-07-24
VMAlert cannot find notifiers using selectors


uhthomas:
I have VMAlert configured as such:

```json
"notifiers": [
    {
        "selector": {
            "labelSelector": {
                "matchLabels": {
                    "app.kubernetes.io/name": "vmalertmanager"
                }
            }
        }
    }
],
```

I know this is valid because:

```sh
❯ k -n vm get po -lapp.kubernetes.io/name=vmalertmanager
NAME                  READY   STATUS    RESTARTS   AGE
vmalertmanager-vm-0   2/2     Running   0          4h39m
vmalertmanager-vm-1   2/2     Running   0          4h39m
vmalertmanager-vm-2   2/2     Running   0          4h38m
❯ k -n vm get svc -lapp.kubernetes.io/name=vmalertmanager
NAME                                   TYPE           CLUSTER-IP       EXTERNAL-IP                                     PORT(S)                      AGE
vmalertmanager-vm                      ClusterIP      None             <none>                                          9093/TCP,9094/TCP,9094/UDP   20h
vmalertmanager-vm-additional-service   LoadBalancer   10.111.185.167   vmalertmanager-unwind-k8s.tailnet-fbec.ts.net   80:30687/TCP                 20h
```

The operator is not happy about it though:

```
[...]

PR #703, opened 2023-07-24
fix(vmalert): correctly remove notifiers without a selector

uhthomas:
The original implementation would always return an empty list of notifiers.

Fixes: #702


uhthomas:
Upon further thinking and reading, I'm not sure it will fix the linked issue. However, it definitely does address some problematic assumptions made by the original filtering code. There are a few scenarios where it will do the complete wrong thing, which isn't ideal.

Haleygo:
> Upon further thinking and reading, I'm not sure it will fix the linked issue. However, it definitely does address some problematic assumptions made by the original filtering code. There are a few scenarios where it will do the complete wrong thing, which isn't ideal.

Thanks for the pr! 
Can you add some scenarios that could be wrong for the origin code in the test file?

uhthomas:
@Haleygo Will do.

AndrewChubatiuk:
covered notifiers discovery [here](https://github.com/VictoriaMetrics/operator/pull/1570). Nothing suspicious was found besides order, in which discovered notifiers are sorted (its reverse). Closing this as it doesn't fix any issue

Issue #706, opened 2023-07-26
vmalertmanager: operator always requests memory


uhthomas:
The operator will always apply a memory request of `200Mi` to the Alertmanager pod, even if limits are already set. This is problematic if the limit is lower than the request, or if the user does not want an explicit request. I usually only set limits, as the Kubernetes API will automatically set the requests to match those limits and the QOS class will then be considered "Guaranteed". However, the QOS class is considered "Burstable" because of the default memory request. This should just be removed.

https://github.com/VictoriaMetrics/operator/blob/aff55e6191a844ee84e07ebc04116aa79853f961/controllers/factory/alertmanager.go#L124-L126

![image](https://github.com/VictoriaMetrics/operator/assets/9749173/d8e2ec52-b692-45ad-8f12-042c48ee9d03)

Amper:
Added at [v0.36.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.36.0) release

PR #707, opened 2023-07-27
chore(vmalertmanager): remove default 200Mi request

uhthomas:
None of the other VM objects have this behaviour and it can cause problems if the memory limit is lower than the request.

Fixes: #706


Haleygo:
Thanks for the fix! But there is default resources for vmalertmanager that haven't be used correctly, should be fixed together in https://github.com/VictoriaMetrics/operator/pull/710. Will close this)

PR #712, opened 2023-07-30
[feature] Add more default securityContext

Haleygo:
Part of https://github.com/VictoriaMetrics/operator/issues/637
1. If KubeVersion>=1.20, set `FSGroupChangePolicy: onRootMismatch`to skip the recursive permission change when the root of the volume already has the correct permissions in default PodSecurityContext.
2. SeccompProfile: Use `RuntimeDefault` seccomp profile by default, which is defined by the container runtime, instead of using the Unconfined (seccomp disabled) mode.
3. Capabilities: {Drop: []v1.Capability{"ALL"}}

### DIFF
--- a/controllers/factory/builders.go
+++ b/controllers/factory/builders.go
@@ -528,13 +528,18 @@ func addStrictSecuritySettingsToPod(p *v1.PodSecurityContext, enableStrictSecuri
 	if !enableStrictSecurity || p != nil {
 		return p
 	}
-	return &v1.PodSecurityContext{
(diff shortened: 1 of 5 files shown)


Haleygo:
PTAL @f41gh7 @Amper

f41gh7 (approved):
LGTM

Haleygo:
@f41gh7  Need to fix the test, can you please reviewing this again)

f41gh7 (approved):
LGTM

Issue #669, opened 2023-06-18
VMUser introduce an option for backend authorization.


f41gh7:
It would be great to have an option to define Bearer token or Basic Auth configuration for targets.

 For instance:
```yaml
kind: VMUser
  spec
    targetRefs:
     paths:
     - static:
       - ursl: [http://host1:8429,http://host2:8429,http://host3:8429]
       basicAuth:
        password:
          key: password
          name: secretname
        username:
          key: username
         name: secretname
```

 Must produce following `vmauth` config with Authorization header:
```yaml
users:
- username: "user"
  password: "***"
  headers:
  - Authorization: Basic BASE64_ENCODED_VALUE
  url_prefix:
  - "http://host1:8429/"
  - "http://host2:8429/"
  - "http://host3:8429/"
```

 It's safe to store headers at configuration file, since it uses secure kubernetes secret.

 For multiple, url_map headers should be used.

 Related issue https://github.com/VictoriaMetrics/VictoriaMetrics/issues/4361

mohammadkhavari:
Can I send a pull request for this feature?

Haleygo:
> Can I send a pull request for this feature?
> 
> 

Feel free)

hagen1778:
@Haleygo @f41gh7 I see this FR was implemented in operator repo. Should we close this ticket as done or follow-up in vmauth is required?

Haleygo:
> @Haleygo @f41gh7 I see this FR was implemented in operator repo. Should we close this ticket as done or follow-up in vmauth is required?

Yes, https://github.com/VictoriaMetrics/operator/pull/714 provides field `TargetRefBasicAuth` in vmuser, it builds seperated header with authorization token for each target, close this issue as completed.
And according to this [comment](https://github.com/VictoriaMetrics/VictoriaMetrics/issues/4361#issuecomment-1588964895), no further move is needed in vmauth.

Issue #522, opened 2022-09-05
refactor auth options for components


f41gh7:
Unify auth options with HTTPAuth.
Check the following:
1) Check basicAuth correctly uses passwordFile option.
2) OAuth2 - adds it for remote options
3) bearerToken - add for remote options

Haleygo:
@f41gh7 
I'd like to divide this into following tasks:

- [ ] unify `oauth2` field tag with "oauth2"
Will introduce breaking change in vmalert config, now using "OAuth2".
- [ ] modify `TokenFilePath` under BearerAuth with "bearerTokenFile", the same with others under scrape objects.
Will introduce breaking change in vmalert config, now using "bearerTokenFilePath".
- [ ] unify/implement HTTPAuth for vmagent.remoteWriteSpec, scrape objects...
Shouldn't introduce any breaking change to user.

hagen1778:
@Haleygo could you please update me on the state of this ticket?

PR #714, opened 2023-08-01
Add backend authentication for targetRefs on vmusers by secret adapt …

mohammadkhavari:
This pull request will add backend basic authentication support on vmuser to add them as header on vmauth configuration.
if this approach is accepted I'll add other authentication formats (like bearer) if needed.
https://github.com/VictoriaMetrics/operator/issues/669


Haleygo:
Hello, we are trying to unify the auth config with `HTTPAuth`, can you reuse it on here?
https://github.com/VictoriaMetrics/operator/blob/f2b8cf701a33f91cef19848c857fd6efb7db59dd/api/v1beta1/additional.go#L211-L223

mohammadkhavari:
Sure.

mohammadkhavari:
I'll fix this.

mohammadkhavari:
I faced some issues about using HttpAuth in VMuser definitions.
It is clear that HttpAuth should place [inline] should place into TargetRef as the [feature request](https://github.com/VictoriaMetrics/operator/issues/669#issue-1762310166) is mentioning it will be transform to the targetRef headers. 
And all properties in VMuser's finally should be transformed to vmauth configuration, here goes the challenges I have in writing the logic of this transformation for each field of HttpAuth:

1) Headers: as it is present already in the targetRef struct it will override the outer Header field, the logic could remain the same . but the description will override by the HttpAuth Headers description that is mentioning wrong format of headers format:

>  HTTPAuth Headers: semicolon separated header with value e.g. headerName:headerValue 
> TargetRef Headers: ["header_key: value1,value2"]

2) BasicAuth, bearerToken:
It can be implemented using headers configurations the only drawback are passwordFiles field, they usage are on CRDs run a pod and there's filesystem so we can refer a file. But here there's another operation.

3) About TLSConfig, Oauth: as I have investigated  these cannot be satisfy only through headers or current vmauth configuration we have here for each user.

And the logic behind this transformation in something like vmalert is much simpler and straightforward: as we can see theres [options](https://docs.victoriametrics.com/vmalert.html ) like `-remoteWrite.oauth2.clientID string`

my recommendation is to first implement HttpAuth features in vmauth component ([as I have created issue for basicAuth](https://github.com/VictoriaMetrics/VictoriaMetrics/issues/4361)) and support them in configuration formats then update operator.

What do you think? @Haleygo @f41gh7

Haleygo:
>I faced some issues about using HttpAuth in VMuser definition.

@mohammadkhavari Thank you for all the work! Sorry I didn't look too deep when I propose that.
Now I think it's ok to just use `BasicAuth` here like you already did, I might do some follow up when I work on [this](https://github.com/VictoriaMetrics/operator/issues/522) later.

mohammadkhavari:
Thank you, Please inform me if any refactoring, code completion, or other modifications, such as adding tests or fixing code, are required.
@Haleygo

mohammadkhavari:
@Haleygo Are there any updates on the feature?

api/v1beta1/vmuser_types.go
```diff
@@ -78,6 +78,9 @@ type TargetRef struct {
 	// https://docs.victoriametrics.com/vmauth.html#ip-filters
 	// +optional
 	IPFilters VMUserIPFilters `json:"ip_filters,omitempty"`
+	// BasicAuth allow an endpoint to authenticate over basic authentication
+	// +optional
+	BasicAuth *BasicAuth `json:"basicAuth,omitempty"`
```
Haleygo: I think we don't need to reuse `BasicAuth` here, the only field missing in vmuser now is `Username v1.SecretKeySelector ` and we can add it directly to `VMUserSpec`.
`Password v1.SecretKeySelector` in BasicAuth should already be covered by `PasswordRef *v1.SecretKeySelector` https://github.com/VictoriaMetrics/operator/blob/0628def2e2821a860fc3a9bf4f911f80cabb8cd2/api/v1beta1/vmuser_types.go#L26-L27
And `PasswordFile` field is not used.

Haleygo (commented):
Sorry for the long silence!
Please see comment.

api/v1beta1/vmuser_types.go
mohammadkhavari: There seems to be a bit of confusion regarding the nature of this feature. As far as my usage experience goes, a vmuser can encompass multiple path routes, and each of these paths includes various hosts as backends (though typically, we utilize only one host per targetRef entry). Our objective is to implement basic authentication for these backends. The challenge lies in setting distinct basic authentication credentials for each endpoint, but how we can set basic auth on these endpoints with some field that has appeared on top level field like vmuser's spec itself?

While I've built a custom operator to address our requirements, it has served our needs effectively for quite some time. However, we want to align it with the latest version of VictoriaMetrics and utilize the officially released version to ensure ongoing compatibility.

Haleygo: Oh, sorry, you're right!
But I still think we can add `Username v1.SecretKeySelector` and `Password v1.SecretKeySelector` directly instead of reuse `BasicAuth` since `PasswordFile` is not used.

mohammadkhavari: so should we add it under vmuserSpec.TargetRef ?
It will be a little confusing and not straightforward to know they are for basicAuth? and also it does not seem to be a good solution to provide other methods of http authentication as you've mentioned[ you're working on. ](https://github.com/VictoriaMetrics/operator/issues/522)

Haleygo: >It will be a little confusing and not straightforward to know they are for basicAuth? 

I think with description, it's ok. 
I'm more against introducing invalid field, which looks like a bug.
What do you think @f41gh7

f41gh7: I agree, it's better to use username and password directly. Or wrap it into new struct `TargetRefAuth`.

 It must have  `username` and `password` for basic Authorization. 

 E.g.

```go
type TargetRefAuth struct{
  Username v1.SecretKeySelector
  Password v1.SecretKeySelector
}
```
 
 I think, it less confusing.

mohammadkhavari: I think It's good enough to proceed.
I'll go ahead and update the code with the suggested interface. 
Do you Agree?

Haleygo: @mohammadkhavari yes, please go ahead!

mohammadkhavari:
@Haleygo @Amper 
Hi, I've updated the implementation due to the suggested interface. A test case has been added, and api.md has been updated as well. I would appreciate it if you could review my changes.

f41gh7:
> @Haleygo @Amper Hi, I've updated the implementation due to the suggested interface. A test case has been added, and api.md has been updated as well. I would appreciate it if you could review my changes.

 Thanks! I'll take a look soon and most probably, it'll a part of upcoming release.

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

mohammadkhavari:
@f41gh7 According to the failed pipeline, Ive noticed that I missed the refactor structure commit, I have fixed implementations to the new interfaces and make lint, make test are also passsing locally.

PR #689, opened 2023-07-17
fix oauth2 field tag

Haleygo:
part of https://github.com/VictoriaMetrics/operator/issues/522
1. fix misspelled oauth2 field, unify tag using "oauth2"
2. upgrade controller-gen to latest [v0.12.1](https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.12.1)

PR #688, opened 2023-07-17
fix TokenFilePath field tag for vmalert

Haleygo:
part of https://github.com/VictoriaMetrics/operator/issues/522
fix tag "bearerTokenFilePath" -> "bearerTokenFile"

PR #717, opened 2023-08-03
make up changelogs for fields change

Haleygo:
make up changelog for https://github.com/VictoriaMetrics/operator/pull/688 and https://github.com/VictoriaMetrics/operator/pull/689

### DIFF
--- a/docs/CHANGELOG.MD
+++ b/docs/CHANGELOG.MD
@@ -4,7 +4,7 @@
 
 ### Breaking changes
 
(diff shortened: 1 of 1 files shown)


Amper:
> these fields need to be modified to `oauth2` before upgrade.

They can't be changed "before updating" because they won't pass CRD validation.
That's the thing, they'd have to do something like that:
- Stop the operator
- Update the CRD (at which time they will lose the current values of the disappeared fields)
- Update the value of the new fields
- Start the operator

Haleygo:
> > these fields need to be modified to `oauth2` before upgrade.
> 
> They can't be changed "before updating" because they won't pass CRD validation. That's the thing, users have to do something like that:
> 
> * Stop the operator
> * Update the CRD (at which time they will lose the current values of the disappeared fields)
> * Update the value of the new fields
> * Start the operator

Right, then I think the step should be:
1. backup all the vmalert objects in files, modify those fields locally
2. upgrade the opeartor
3. re-apply those files

But how do we describe those steps in changelog, do we need to? This sounds like a standard upgrade procedure for crd.

Haleygo:
Refer to some changelogs in [kubernetes](https://github.com/rkamudhan/kubernetes/blob/9f75e03481429d66be21a914059ea8041f347f9d/CHANGELOG-1.9.md?plain=1#L2173).

Issue #663, opened 2023-06-08
disable PSP on a running k8s cluster < 1.25


frankconrad:
In prepare to update to 12.5,... we need to remvoe all PodSecurityPolicy

Which change config to:
operator.psp_auto_creation_enabled: false
There still the generated PSP.
The ty to remove it via kubectl dleete psp ... hangs forever. As you have an finalizer and there is blocking it.

Best would be that in case disable PSP you make sure that any eventualy before generated PSP get removed.

We use vmagent, vmalert and vmcluster and all 3 have same issue.

The code look like only take care of create node delete

zekker6:
@f41gh7 @Amper Could you please take a look?

Haleygo:
@frankconrad Hi, what operator version do you use?
>There still the generated PSP.

You mean those psp already generated are still there, right? 
Turn off `psp_auto_creation_enabled` won't let operator clean psp created before, but stop creating new psp when new vmalert/vmagent/vmcluster applied.
So can you try delete the old vmalert/vmagent/vmcluster resources to let the operator do the cleanup thing and recreate them.

Haleygo:
@frankconrad Hi, what operator version do you use?
>There still the generated PSP.

You mean those psp already generated are still there, right? 
Turn off `psp_auto_creation_enabled` won't let operator clean psp created before, but stop creating new psp when new vmalert/vmagent/vmcluster applied.
So can you try delete the old vmalert/vmagent/vmcluster resources to let the operator do the cleanup thing and recreate them.

frankconrad:
I use victoria-metrics-operator chart 0.23.1
which use operator v0.34.1

yes there are still there
But I can't delete them, kubectl hangs

frankconrad:
On a prodcution system you can't delete them...
I think there should be an better handling on operator side.
Remove finalizer is a hack.

f41gh7:
> In prepare to update to 12.5,... we need to remvoe all PodSecurityPolicy
> 
> Which change config to: operator.psp_auto_creation_enabled: false There still the generated PSP. The ty to remove it via kubectl dleete psp ... hangs forever. As you have an finalizer and there is blocking it.
> 
> Best would be that in case disable PSP you make sure that any eventualy before generated PSP get removed.
> 
> We use vmagent, vmalert and vmcluster and all 3 have same issue.
> 
> The code look like only take care of create node delete

[...]

PR #726, opened 2023-08-18
disable PodSecurityPolicy creation by default

Haleygo:
PodSecurityPolicy already got deprecated since [kubernetes v1.25](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#psp-v125), change option `VM_PSPAUTOCREATEENABLED` default value from `true` to `false`.
It won't effect those users who already got psp on their cluster when they upgrade operator version, see https://github.com/VictoriaMetrics/operator/issues/663.


Amper (approved):
LGTM

PR #732, opened 2023-08-25
fix vmselect cache dir mount when enableStrictSecurity

Haleygo:
After [enable strict securityContext](https://github.com/VictoriaMetrics/operator/blob/master/vars.MD#L119) in v0.36.0 by default, if `vmselect.cacheMountPath` not set, vmselect won't be start successfully cause default cache dir "/tmp" is protected. 
Fix this by using `/cache` dir and mount to emptyDir if `vmselect.cacheMountPath` not set.

Another fix is for vmselect and vmstorage, both podSecurityContext was set to non-nil, thus default podSecurityContext won't be applied to them.


Amper (approved):
LGTM

Issue #730, opened 2023-08-24
Chained `if` conditions for relabeling not supported in vmagent CRD


towolf:
### Describe the bug

In the code for `IfExpression` it states that two modes are supported, a single string matcher and a chained mode with an array of conditions which are combined with OR:

https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/lib/promrelabel/if_expression.go#L13-L23

However the CRD for vmagent does not support this: 
![image](https://github.com/VictoriaMetrics/VictoriaMetrics/assets/345210/956e1253-9fe2-47ea-83ac-9c1477e7c73c)

We would like to use a chained config to avoid a gnarly regex matcher, like this:

```yaml
          inlineUrlRelabelConfig:
            - if:
                # Limit metrics sent to the central obs cluster to this set (keep)
                - '{job=~"(kube-etcd|agones.+|node-exporter|apiserver)"}'
                # Also pass on the metrics from recording rules created by
                # kube-prometheus-stack and vmstack to make the multi-cluster k-p-s
                # dashboards work
                - '{__name__=~"(cluster:namespace|cluster:node_cpu|cluster_quantile:|code_resource:|instance:|instance_device:|namespace_cpu:|namespace_memory:|namespace_workload_pod:|node_namespace_pod_container).+"}'
              action: keep
```
This means, the single `if` contains two matchers in an array which should be combined with OR.

### To Reproduce

Add a relabel config like this to the vmagent CR:

```yaml
if:
 - 'foo{bar="baz"}'
 - '{x=~"y"}'
 ```

### Version

`image: docker.io/victoriametrics/vmagent:v1.93.0`

### Logs

`Error: cannot patch "vmstack" with kind VMAgent: VMAgent.operator.victoriametrics.com "vmstack" is invalid: [spec.remoteWrite[1].inlineUrlRelabelConfig[0].if: Invalid value: "array": spec.remoteWrite[1].inlineUrlRelabelConfig[0].if in body must be of type string: "array", spec.remoteWrite[1].inlineUrlRelabelConfig[1].if: Invalid value: "array": spec.remoteWrite[1].inlineUrlRelabelConfig[1].if in body must be of type string: "array"]`

zekker6:
@towolf Thank you for the report!
I've transferred this issue to operator repository as it is related to the operator, but not vmagent itself.

Amper:
@towolf, yes, you should open another issue for that.

PR #738, opened 2023-08-29
supported multiple `if` conditions for relabeling(#730)

Amper:
issue #730


Amper:
Hey @Haleygo and @f41gh7, I decided to make the `if` and `match` schemaless fields to support `string or array of strings` and avoid breaking backward compatibility, wdyt about it?

f41gh7:
> Hey @Haleygo and @f41gh7, I decided to make the `if` and `match` schemaless fields to support `string or array of strings` and avoid breaking backward compatibility, wdyt about it?

Is it possible to apply those changes over exist operator installation?

Amper:
> Is it possible to apply those changes over exist operator installation?

Yes, it works, I tested it - current values are saved.
The only problem with this approach is that, without webhooks enabled, any object of the wrong type can be mapped into this field and will be deployed - it will break only in the operator's log

f41gh7 (approved):
LGTM

Issue #675, opened 2023-06-22
How to determine which version of victoriametrics the operator will install by default


willemm:
Hi,

We're trying to build an automated upgrade pipeline for some applications including VictoriaMetrics operator.
Once the operator is upgraded, we would like to check if it's functioning properly.  The easiest way to do this would seem to be to check if all components are up and running, and have the correct version.  (e.g. `vm_app_version{namespace="victoria-metrics",short_version =~ "^v1.91.0.*"} and up{namespace="victoria-metrics"} == 1)` )

However, it seems there is no way to deduce from the operator which version it will be installing by default.  The default version is hardcoded into the operator: https://github.com/VictoriaMetrics/operator/blob/4c97f70c9a775d2bfff401862acabd5452ef0cf8/internal/config/config.go#L131

Note: I do not want to know how to change the version, nor override the default.  I want the default version.  But I want my automated pipeline to know which version that is.

Haleygo:
@f41gh7 I think the default version of each components in operator are hidden from the users, and there is no doc to summarize that. Should we maintain it somewhere?
For now, we seems to update them to the latest tag when operator released.

f41gh7:
> @f41gh7 I think the default version of each components in operator are hidden from the users, and there is no doc to summarize that. Should we maintain it somewhere or expose [some of them](https://github.com/VictoriaMetrics/operator/blob/4c97f70c9a775d2bfff401862acabd5452ef0cf8/vars.MD?plain=1#L12) in operator pod? For now, we seems to update them to the latest tag when operator released.

  I think, we have to mention at doc how to manage CR objects versions - default values, global operator settings, per CR value. And best practice for it. I recommend to define versions at CR spec definition and manage it via git-ops. As we usually do internally with `argo-cd`.

 For current issue, best option for me looks a `--printDefaults` flag. That will print out json with defaults. Aka:

```sh
./bin/manager --printDefault
{
   "flags": {"logger.Level":  "info"  },
    "config": {
          "vmalert_default": {"version": "v1.90.1","image": "victoriametrics/vmalert" }
 }
}
```

 and other values for the `BaseOperatorConf` struct.

 In this case, automated pipeline could parse it with `jq` and make some decision.

Amper:
Hey. 

In PR #742 implemented the following flags for the operator:

- **`-printDefaults`** - *print all variables with their default values and exit*
[...]

PR #742, opened 2023-09-02
add ability to print default values for all operator variables (#675)

Amper:
issue #675

### DIFF
--- a/e2e/suite/suite.go
+++ b/e2e/suite/suite.go
@@ -2,6 +2,7 @@ package suite
 
 import (
 	"context"
+	"github.com/VictoriaMetrics/operator/internal/config"
 	"os"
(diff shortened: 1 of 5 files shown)


e2e/suite/suite.go
```diff
@@ -2,6 +2,7 @@ package suite
 
 import (
 	"context"
+	"github.com/VictoriaMetrics/operator/internal/config"
```
Haleygo: Need some format here.

Haleygo (approved):
LGTM!

e2e/suite/suite.go
Amper: fixed

Haleygo (approved):
LGTM!

PR #745, opened 2023-09-04
controllers/vmuser: fixes ip_filters indent

f41gh7:
previously config was incorrectly rendered if multiple targetRef was defined


Amper (approved):
LGTM

Issue #749, opened 2023-09-04
[vmstorage] panic: FATAL: cannot remove "/vm-data/indexdb/...": unlinkat /vm-data/indexdb/.../metadata.json: permission denied


ziouf:
### Describe the bug

At startup, VMStorage crashes with the following message (full log bellow):

> panic: FATAL: cannot remove "/vm-data/indexdb/177E9C9207F871EC/1780E117A2091B9B": unlinkat /vm-data/indexdb/177E9C9207F871EC/1780E117A2091B9B/metadata.json: permission denied

### To Reproduce

I dont know. It happen on pod restart after upgrade to the latest operator helm chart. 
VMStorage is out-of-service even if I force to rollback to previous versions by forcing value of `spec.vmstorage.image.tag` field in VMCluster CRD. 

### Version

vmstorage-20230902-002932-tags-v1.93.3-cluster-0-gf78d8b994d

### Logs

[...]

PR #750, opened 2023-09-06
turn off `EnableStrictSecurity` by default

Haleygo:
address https://github.com/VictoriaMetrics/operator/issues/749
turn off `EnableStrictSecurity` by default


Amper (approved):
LGTM

PR #751, opened 2023-09-07
vmagent: clarify Labels param description

hagen1778:
See https://github.com/VictoriaMetrics/VictoriaMetrics/pull/4898#discussion_r1307447532

### DIFF
--- a/api/v1beta1/vmagent_types.go
+++ b/api/v1beta1/vmagent_types.go
@@ -416,7 +416,7 @@ type VMAgentRemoteWriteSettings struct {
 	// +optional
 	// +kubebuilder:validation:Pattern:="[0-9]+(ms|s|m|h)"
 	FlushInterval *string `json:"flushInterval,omitempty"`
-	// Optional labels in the form 'name=value' to add to all the metrics before sending them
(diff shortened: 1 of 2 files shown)


api/v1beta1/vmagent_types.go
```diff
@@ -416,7 +416,7 @@ type VMAgentRemoteWriteSettings struct {
 	// +optional
 	// +kubebuilder:validation:Pattern:="[0-9]+(ms|s|m|h)"
 	FlushInterval *string `json:"flushInterval,omitempty"`
-	// Optional labels in the form 'name=value' to add to all the metrics before sending them
+	// Labels in the form 'name=value' to add to all the metrics before sending them. This overrides the label if it already exists.
```
Amper, line 419: I suppose that this comment (about `'name=value'`) may be taken the wrong way, because in the custom resource yaml it looks like following:

```yaml
# ...
labels:
  name: value
````

Also after changing the API you need to run `make build` to generate a new CRD bundle.

hagen1778: Thanks! Updated

Amper (approved):

Issue #753, opened 2023-09-11
[vmcluster]:  About the `HasSpecChanges` method and the `AnnotationsFiltered` method


z-anshun:
Hi, everyone!
When I read the source code, I had some doubts about `vmcluster`, they may be bugs:
1. The `HasSpecChanges` method does not judge null values, which may cause an error when querying crd for the first time.
code:
https://github.com/VictoriaMetrics/operator/blob/master/api/v1beta1/vmcluster_types.go#L947
error:
```
1.6944398628889349e+09  ERROR   controller      failed to check if cluster spec changed {"Request.Namespace": "vm", "Request.Name": "example-vmcluster-persistent", "error": "cannot parse last applied cluster spec value:  : unexpected end of JSON input"}
...
```

2. [annotationFilterPrefixes](https://github.com/VictoriaMetrics/operator/blob/master/api/v1beta1/additional.go#L44C1-L44C1) will filter annotations prefixed by `operator.victoriametrics.com/`, but in the [LastAppliedSpecAsPatch](https://github.com/VictoriaMetrics/operator/blob/master/api/v1beta1/vmcluster_types.go#L937) function, the key of annotation is set to `operator.victoriametrics/last-applied-spec`. I think `annotationFilterPrefixes` expects to filter it out.

Haleygo:
Hmm, that's right. But there is no pods or other resources for vmcluster to wait or show it's ready, I can only come up the idea to have a hardcode timeout or recheck it's spec, both sound not so good. 
Please let me know if you get other idea)

f41gh7:
Linked commit must remove misleading error message.

f41gh7:
Fix included into the https://github.com/VictoriaMetrics/operator/releases/tag/v0.39.0 release

PR #754, opened 2023-09-12
vmcluster: remove redundant annotation from created workloads

Haleygo:
address https://github.com/VictoriaMetrics/operator/issues/753
remove redundant annotation `operator.victoriametrics/last-applied-spec` from created workloads like vmstorage statefulset.


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #760, opened 2023-09-21
fix pvc expansion for sts multiple VolumeClaimTemplates

Haleygo:
pvc should be checked and updated separately when there are multiple VolumeClaimTemplates, previously it's all mixed up.

### DIFF
--- a/controllers/factory/k8stools/expansion.go
+++ b/controllers/factory/k8stools/expansion.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/VictoriaMetrics/VictoriaMetrics/lib/logger"
 	victoriametricsv1beta1 "github.com/VictoriaMetrics/operator/api/v1beta1"
 	"github.com/VictoriaMetrics/operator/internal/config"
 	"github.com/go-test/deep"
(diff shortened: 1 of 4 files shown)


controllers/factory/k8stools/expansion.go
```diff
@@ -139,24 +140,46 @@ func getPVCFromSTS(pvcName string, sts *appsv1.StatefulSet) *corev1.PersistentVo
...
+	}
+	for _, pvc := range pvcs.Items {
+		var isExist bool
+		for _, tpvc := range targetPVCs {
+			if strings.HasPrefix(pvc.Name, fmt.Sprintf("%s-%s", tpvc.Name, sts.Name)) {
+				isExist = true
+				// check if storage class is expandable
+				isExpandable, err := isStorageClassExpandable(ctx, rclient, &pvc)
+				if err != nil {
+					logger.Errorf("failed to check storageClass expandability for pvc %s: %v", pvc.Name, err)
```
f41gh7: I think, it's better to return errors back to the caller. Otherwise it performs partial update and actual resize may never happen.

controllers/factory/k8stools/expansion.go
```diff
@@ -139,24 +140,47 @@ func getPVCFromSTS(pvcName string, sts *appsv1.StatefulSet) *corev1.PersistentVo
...
 		return err
 	}
-
-	return growPVCs(ctx, rclient, pvc.Spec.Resources.Requests.Storage(), sts.Namespace, sts.Spec.Selector.MatchLabels, isExpandable)
+	if len(pvcs.Items) == 0 {
+		log.Info("PVCs select call returned 0 pvcs, it could be a bug, want match %v in namespace %s", sts.Spec.Selector.MatchLabels, sts.Namespace)
+		return nil
+	}
+	for _, pvc := range pvcs.Items {
+		var isExist bool
```
f41gh7, line 157: I think, we can perform checks for isExpandable at the outer loop:

```
for _, pvc := range pvcs.Items {
		var isExist bool
                var isExpandable bool
                isExpandable, err := isStorageClassExpandable(ctx, rclient, &pvc)
				if err != nil {
				return fmt.Errorf("failed to check storageClass expandability for pvc %s: %w", pvc.Name, err)
}
```

 It reduces the number of possible API calls for storageClassList and simplifies code a bit.

f41gh7 (commented):
Good catch! Looks good overall, have few comments.

controllers/factory/k8stools/expansion.go
Haleygo: Right, fixed!
I wasn't sure to break multiple update processes by returning once get an error, or just log it and give the rest obejcts a chance to be updated. 
In this case, seems better to return it, since it's already a partial update, and by returning it, the error message would also be persisted to vmcluster.status.reason)

controllers/factory/k8stools/expansion.go
Haleygo: fixed, thanks!

controllers/factory/k8stools/expansion.go
```diff
@@ -167,8 +188,7 @@ func isStorageClassExpandable(ctx context.Context, rclient client.Client, pvc *c
 	}
 	// fast path at single namespace mode, listing storage classes is disabled
 	if !config.IsClusterWideAccessAllowed() {
-		log.Info("cannot detect if storageClass expandable at single namespace mode, expand PVC manually or enforce resizing with annotation to true", "pvc annotation", victoriametricsv1beta1.PVCExpandableLabel)
-		return false, nil
+		return false, fmt.Errorf("cannot detect if storageClass expandable at single namespace mode, need to expand PVC manually or enforce resizing by adding annotation `%s=true`", victoriametricsv1beta1.PVCExpandableLabel)
```
f41gh7, line 191: I think, it's not an error. Since, in single namespace mode it'd be not possible to use PVC. Operator will produce following error each time and cannot reconcile resources without user action.

 I think, it's better to return false, nil.

Haleygo: >Operator will produce following error each time

It won't be there each time, since sts will be updated for the first time and won't be marked as [isRecreated](https://github.com/VictoriaMetrics/operator/blob/fix-grow-sts-pvc/controllers/factory/k8stools/sts.go#L82), thus skip the pvc check.

I thought we should treat it as same as `failed to list storageClass` cause both of them can't list sc. In single namespace mode, I assume user should notice this error and fix it according to the notes.

f41gh7: I's a breaking change for current single namespace mode users. I suggest to mention it at our docs + add an example how to configure vmsingle/vmcluster with manual annotation with storage class params

f41gh7 (approved):
LGTM

controllers/factory/k8stools/expansion.go
f41gh7: you can update docs with additional PR.

PR #763, opened 2023-09-25
corrected endpointslices apigroup

soanni86:
Hello VictoriaMetrics team,

please correct the apiGroup for endpointsices.

tested on a local k8s cluster with single mode vm

```
cat <<EOF | k apply -f -
kind: VMSingle
metadata:
  name: example-vmsingle-notpersisted
spec:
  retentionPeriod: "1"
EOF
```

```
cat <<EOF | k apply -f -
kind: VMAgent
metadata:
  name: example-vmagent
spec:
  selectAllByDefault: true
  replicaCount: 1
  remoteWrite:
    - url: "http://vmsingle-example-vmsingle-notpersisted.default.svc:8429/api/v1/write"
EOF
```

before the fix:

```
kubectl auth can-i get endpointslices --as=system:serviceaccount:default:vmagent-example-vmagent -n default
no
```

after the fix
```
kubectl auth can-i get endpointslices --as=system:serviceaccount:default:vmagent-example-vmagent -n default
yes
```

its correct here for vmagent in charts repository https://github.com/VictoriaMetrics/helm-charts/blob/8d35f1f224861327e40e91e0bdd151ac467d1ab7/charts/victoria-metrics-agent/templates/clusterrole.yaml#L21

was fixed quite a while ago by this commit https://github.com/VictoriaMetrics/helm-charts/commit/acf5d8aadeb1a7c4c92296f8456d9ab4786686fd

### DIFF
--- a/config/examples/operator_rbac_for_single_namespace.yaml
+++ b/config/examples/operator_rbac_for_single_namespace.yaml
@@ -24,10 +24,17 @@ metadata:
   namespace: default
 rules:
   - apiGroups:
+      - "discovery.k8s.io"
+    resources:
+      - endpointslices
+    verbs:
+      - 'list'
+      - 'watch'
+      - 'get'
+  - apiGroups:
       - ""
     resources:
       - endpoints
-      - endpointslices
     verbs:
       - 'list'
       - 'watch'
--- a/config/examples/vmagent_rbac.yaml
+++ b/config/examples/vmagent_rbac.yaml
@@ -9,7 +9,7 @@ kind: ClusterRole
 metadata:
   name: vmagent
 rules:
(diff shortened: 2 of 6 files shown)


Haleygo:
@soanni86 Thanks for the fix! 
LGTM)
Do you mind to add some changelog to [doc](https://github.com/VictoriaMetrics/operator/blob/master/docs/CHANGELOG.MD#L13) with something like 
`- [vmoperator](https://docs.victoriametrics.com/operator/): fix wrong api group of endpointsices, before vmagent won't able to access endpointsices resources with default rbac rule.`

soanni86:
@Haleygo added the line in changelog, thanks

Haleygo (approved):
LGTM!
Thanks!

PR #764, opened 2023-09-26
controllers/k8stools: fix grow sts pvcs behavior

Haleygo:
follow up https://github.com/VictoriaMetrics/operator/pull/760
1. check if pvcs need to be reize as long as there is pvc since error could happen when grow pvcs and sts was updated already
2. not return error to caller if there is no point to retry when growPVC, like sc is unexpandable
If sc is unexpandable or vm opearator is under single namespace mode, pvc must be expand with user's intervention. 
Breaking reconcile and requeue under those circumstance will have no other effects except cause more pressure on cluster[list/get request and will perform sts rolling update once again].


f41gh7 (approved):
LGTM!

f41gh7:
Thanks!

PR #774, opened 2023-10-02
Fixed docs pipeline

Amper:


.github/workflows/docs.yaml
```diff
@@ -71,20 +71,21 @@ jobs:
...
-          cat ./vars.md > ../__vm-docs-repo/docs/operator/vars.md
-          cat ./doc_api.md > ../__vm-docs-repo/docs/operator/api.md
           for md in $(find docs/**/*.md); do
-            mkdir -p ../__vm-docs-repo/$(dirname $md)
-            cat $md > ../__vm-docs-repo/$md
+            newmd=$(echo $md | sed -e 's/^docs\//&operator\//g')
+            mkdir -p ../__vm-docs-repo/$(dirname $newmd)
+            cat $md > ../__vm-docs-repo/$newmd
           done
+          cat ./vars.md > ../__vm-docs-repo/docs/operator/vars.md
```
zekker6, line 79: It seems like `newmd` uses path with `docs/operator`  removed, but this path is using `docs/operator`. Is it expected?

Amper: not removed, path `docs/...` in `newmd` replaced by `docs/operator/...`, because operator have separate directory in main repo

zekker6 (approved):

PR #789, opened 2023-10-24
passing `-replicationFactor` to vmselect

Amper:
from now on operator passes `-replicationFactor` (if it set in `vmcluster`) for `vmselect` 

issue: #778

### DIFF
--- a/controllers/factory/vmcluster.go
+++ b/controllers/factory/vmcluster.go
@@ -399,16 +399,21 @@ func makePodSpecForVMSelect(cr *v1beta1.VMCluster, c *config.BaseOperatorConf) (
 		args = append(args, fmt.Sprintf("-loggerFormat=%s", cr.Spec.VMSelect.LogFormat))
 	}
 	if cr.Spec.ReplicationFactor != nil && *cr.Spec.ReplicationFactor > 1 {
-		args = append(args, fmt.Sprintf("-replicationFactor=%d", *cr.Spec.ReplicationFactor))
-
+		var replicationFactorIsSet bool
 		var dedupIsSet bool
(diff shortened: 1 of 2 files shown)


controllers/factory/vmcluster.go
```diff
@@ -399,6 +399,8 @@ func makePodSpecForVMSelect(cr *v1beta1.VMCluster, c *config.BaseOperatorConf) (
 		args = append(args, fmt.Sprintf("-loggerFormat=%s", cr.Spec.VMSelect.LogFormat))
 	}
 	if cr.Spec.ReplicationFactor != nil && *cr.Spec.ReplicationFactor > 1 {
+		args = append(args, fmt.Sprintf("-replicationFactor=%d", *cr.Spec.ReplicationFactor))
```
Haleygo: I think we should also allow user to override `-replicationFactor` using cr.Spec.VMSelect.ExtraArgs, since vmselect can be set up to query other vmstorages with different settings.

Amper: Done

Amper: > According to [this comment](https://github.com/VictoriaMetrics/operator/issues/778#issuecomment-1744695815), should we add `-replicationFactor` only when vmselects>=[v1.91.3](https://github.com/VictoriaMetrics/VictoriaMetrics/releases/tag/v1.91.3) by default. For lower version, users can add it by hand.

It's a pretty old version, I can't imagine a case where users roll back so far from the current VM version in an operator - a lot of things can break, they usually upgrade to a higher version.

Haleygo (approved):
lgtm

PR #793, opened 2023-10-30
CI: autoupdate job will now reuse branch to avoid multiple PRs

k1rk:
so autoupdate job is now consistent with the one used for VM components.

Also now since same branch is used, same PR will updated rather then creating multiple


Amper (approved):
LGTM

PR #794, opened 2023-10-30
ci: add missing path param

k1rk:


.github/workflows/sandbox.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v5
         with:
-          author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
```
k1rk, line 48: removed author config as it's default value anyway

Amper (approved):

PR #801, opened 2023-11-01
controller/factory/k8stools: optimize `HandleSTSUpdate`

Haleygo:
1. even if statefulset needs recreate because VolumeClaimTemplates changed, sometimes pod doesn't need to be recreated. Since pod's volume only cares VolumeClaimTemplate's name, so other attributes like size or storageClassName won't effect that.
```
    volumes:
    - name: vmstorage-db
      persistentVolumeClaim:
        claimName: vmstorage-db-vmstorage-test-insight-victoria-metrics-k8s-stack-0
```
2. remove hack code from https://github.com/VictoriaMetrics/operator/issues/344, this should be resolved by https://github.com/kubernetes/kubernetes/pull/109694 and other fixes. Statefulset's currentRevision&updateRevision should be updated by statefulset controller now.

ps. also add some comments on why updating sts.Status.CurrentRevision to UpdateRevision by hand, please check if that's right @f41gh7

### DIFF
--- a/controllers/factory/k8stools/expansion.go
+++ b/controllers/factory/k8stools/expansion.go
@@ -30,13 +30,18 @@ func cleanUpFinalize(ctx context.Context, rclient client.Client, instance client
 	return nil
 }
 
-// recreateSTS if needed.
+// recreateSTSIfNeed will check if sts needs recreate and perform recreate if needed,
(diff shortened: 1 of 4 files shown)


f41gh7 (approved):
LGTM

PR #802, opened 2023-11-02
fixes sts rolling update minor issues

f41gh7:


controllers/factory/k8stools/sts.go
```diff
@@ -113,6 +115,12 @@ func performRollingUpdateOnSts(ctx context.Context, podMustRecreate bool, rclien
 	if sts.Spec.Replicas != nil {
 		neededPodCount = int(*sts.Spec.Replicas)
 	}
+	if err := sortStsPodsByID(podList.Items); err != nil {
+		return fmt.Errorf("cannot sort statefulset pods: %w", err)
+	}
+	for _, p := range podList.Items {
```
Haleygo: Looks like a test code)

Haleygo:
I get the part that if sts's replicas>10, listing pod result can be misordered. 
But why do we need the list to be ordered? We already set `podManagementPolicy: "Parallel"` and also have https://github.com/VictoriaMetrics/operator/blob/b89ce3b1ecc12568f877e69c4fb9fbc00d63babd/controllers/factory/k8stools/sts.go#L149-L153
that could mess the order.

f41gh7:
> I get the part that if sts's replicas>10, listing pod result can be misordered. But why do we need the list to be ordered? We already set `podManagementPolicy: "Parallel"` and also have
> 
> https://github.com/VictoriaMetrics/operator/blob/b89ce3b1ecc12568f877e69c4fb9fbc00d63babd/controllers/factory/k8stools/sts.go#L149-L153
> 
> that could mess the order.

It shouldn't mess an order, since sorting happens after list requests to the kubernetes. It makes even an order for pod updates. Parallel order cannot be applied for this case, since it's manually managed rollout process.

Haleygo (approved):
LGTM!

Issue #811, opened 2023-11-22
[VMAlert] cr.spec.Notifier, cr.spec.notifierConfigRef and discovered alertmanager are empty prevent using flag notifier.blackhole


tantm3:
I am currently using the VictoriaMetrics Operator version: victoria-metrics-k8s-stack-0.18.6.
I want to deploy VMAlert `just for using recording rule` so I don't need to configure the spec.notifier and I used spec.extraArgs to enable flag `notifier.blackhole: "true"`.
The configuration file looks like this:
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAlert
metadata:
  name: vmalert-vm-k8s
  namespace: monitoring
spec:
  datasource:
    url: "http://vmselect-vm-k8s-victoria-metrics-k8s-stack.monitoring:8481/select/0/prometheus"
  remoteWrite:
    url: "http://vminsert-vm-k8s-victoria-metrics-k8s-stack.monitoring:8480/insert/0/prometheus"
  remoteRead:
    url: "http://vmselect-vm-k8s-victoria-metrics-k8s-stack.monitoring:8481/select/0/prometheus"
  evaluationInterval: "30s"
  ruleSelector:
    matchLabels:
     metric: "true"
  ruleNamespaceSelector: 
    matchLabels:
      kubernetes.io/metadata.name: monitoring
  # for accessing to vmalert via vmauth with path prefix
  extraArgs:
    http.pathPrefix: /vmalert
    notifier.blackhole: "true"
```

It seems like we have a bug here.
We can not left the cr.spec.Notifier, cr.spec.notifierConfigRef empty If we do the operator can not create VMAlert pod.
But when you enter a random URL just like in the configuration file above, the VMAlert pod shows error:
```
2023-11-22T08:42:55.943Z	fatal	VictoriaMetrics/app/vmalert/main.go:170	failed to init: failed to init notifier: only one of -notifier.blackhole, -notifier.url and -notifier.config flags must be specified
```

Haleygo:
We should allow `notifier` and `notifiers` both to be empty at the first space, since vmalert only needs `-notifier.url ` when there is alerting rules and operator can't check that.

Haleygo:
Should be fixed since [0.39.4](https://github.com/VictoriaMetrics/operator/releases/tag/v0.39.4), close as completed.

tantm3:
Thank @Haleygo for your support!

iPapanya:
[...]

PR #813, opened 2023-11-23
vmalert: remove notifier null check

Haleygo:
, since `-notifier.url` is optional and is needed only if there are alerting rules. 
>https://new.docs.victoriametrics.com/vmalert/#topology-examples
-notifier.url is optional and is needed only if you have alerting rules.

Also vmalert has `-notifier.blackhole` flag which supports evaluating alerting rules without sending any notifications to external receivers (eg. alertmanager), see https://github.com/VictoriaMetrics/operator/issues/811.
I think we can leave the check to vmalert itself.

Also, add minor sanity check for vmauth. 
If `vmauth.Spec.Ingress.TlsSecretName` is used, `vmauth.Spec.Ingress.TlsHosts` is also needed according to the code, otherwise IngressRule will be illegal.
https://github.com/VictoriaMetrics/operator/blob/9a481f2c9d1b30f5fba096ac50af1e7b7cf7f839/controllers/factory/vmauth.go#L413-L425


f41gh7 (approved):
LGTM

PR #815, opened 2023-11-23
Fixed VMAlertmanagerConfig discovery for VMAlertmanager. 

Amper:
+ Fixed VMAlertmanagerConfig discovery for VMAlertmanager according to the docs
+ Fixed markdown renderings issues in docs.

Based on this thread: https://victoriametrics.slack.com/archives/CTC488T1P/p1700734803990519


docs/resources/vmalertmanager.md
```diff
@@ -129,22 +130,21 @@ of `VMAlertmanager` spec and environment variable `WATCH_NAMESPACE` for operator
 
 Following rules are applied:
 
-- If `configNamespaceSelector` and `configSelector` both undefined, then by default select nothing. With option set - `spec.selectAllByDefault: true`, select all vmalertmanagerconfigs.
-- If `configNamespaceSelector` defined, `configSelector` undefined, then all vmalertmaangerconfigs are matching at namespaces for given `configNamespaceSelector`.
+- If `configSelector` undefined, then by default select nothing. With option set - `spec.selectAllByDefault: true`, select all vmalertmanagerconfigs.
```
Haleygo: I think it's a bug, we should support using `configNamespaceSelector` alone to pick up vmalertmanagerconfig, like vmagent does for scrape resources.
I think this check can be removed
https://github.com/VictoriaMetrics/operator/blob/fec3f9d2efa72ed574824c8d45cbf7e3f2a3b184/controllers/factory/alertmanager.go#L544

docs/resources/vmalertmanager.md
```diff
@@ -129,22 +130,21 @@ of `VMAlertmanager` spec and environment variable `WATCH_NAMESPACE` for operator
...
 - If `configNamespaceSelector` undefined, `configSelector` defined, then all vmalertmaangerconfigs at `VMAgent`'s namespaces are matching for given `configSelector`.
```
Haleygo: vmalertmaangerconfigs -> vmalertmanagerconfigs

docs/resources/vmalertmanager.md
Amper: > I think it's a bug, we should support using configNamespaceSelector alone to pick up vmalertmanagerconfig, like vmagent does for scrape resources.
I think this check can be removed

Yes, I too want it to work the same way for all resources. I wanted to remove that condition too at first, but I thought it might break something for someone, so I decided to just describe the current condition correctly.

@f41gh7 wdyt?

docs/resources/vmalertmanager.md
Amper: Fixed

docs/resources/vmalertmanager.md
Haleygo: >but I thought it might break something for someone

I don't see it would break something, since we're loosening the check.

leoskyrocker: btw, configSelector is a superset of configNamespaceSelector.
That is saying, you can simply use:
```
configSelector:
  matchLabels:
    kubernetes.io/metadata.name: monitoring
```
to achieve the same as:
```
configNamespaceSelector:
  matchLabels:
    kubernetes.io/metadata.name: monitoring
```

If I were to vote for the fix here, it'd be simply eliminating `configNamespaceSelector`.

f41gh7: Yes, looks like a bug to me as well.

Amper: Ok, i'll fix code instead of docs

Amper: Fixed

f41gh7 (approved):
LGTM

PR #818, opened 2023-11-28
Fix wrong example in recording rules

tanisdlj:
The example partially presents an alert as example instead of a recording rule. The name also contains white spaces, which is not supported.
Opening a PR here as discussed in this other PR: https://github.com/VictoriaMetrics/VictoriaMetrics/pull/5405


Amper (approved):
LGTM

tanisdlj:
A pleasure @Amper :)

PR #822, opened 2023-12-15
add missing status route in vmauth example

Haleygo:
`/prometheus/api/v1/status` is needed for top/active queries in vmui.
related thread https://victoriametrics.slack.com/archives/CGZF1H6L9/p1702542069333989?thread_ts=1702387812.406289&cid=CGZF1H6L9


Amper (approved):
LGTM

PR #827, opened 2023-12-19
fix operator-sdk generate kustomize manifests

Haleygo:
process will fall into [endless loop](https://github.com/operator-framework/operator-sdk/blob/542966812906456a8d67cf7284fc6410b104e118/internal/generate/clusterserviceversion/bases/definitions/ast.go#L107) before when trying generate manifests, because it's not allowed to have package that uses the same name as a local type, see https://sdk.operatorframework.io/docs/olm-integration/generation/#kustomize-files and https://github.com/operator-framework/operator-sdk/issues/6028.
Remove deprecated packagemanifests cmd, [The operator-framework is removing support for the packagemanifests format in the near future. Hence, operator-sdk generate packagemanifests and operator-sdk run packagemanifests commands are deprecated.](https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.7.1/#deprecate-support-for-packagemanifests)


Haleygo:
I generated /bundle with `operator-sdk generate kustomize manifests` and `operator-sdk bundle validate` passed according to the [doc](https://github.com/operator-framework/community-operators/blob/master/docs/README.md#validating-the-bundle-with-sdk).
About the size check, it should be improved in https://github.com/operator-framework/operator-registry/pull/685 and `bundle validate` check has been updated in https://github.com/operator-framework/api/pull/210 as well.

f41gh7 (approved):
LGTM

Issue #526, opened 2022-09-14
Create alerting rules for operator


f41gh7:
Currently it's not even how to monitor operator properly and which metrics should be tracked.

 Alerting rules should cover:
1) latency between operator and API server.
2) converter errors by object types.
3) building config errors (vmauth, vmalertmanager, vmagent, vmalert).
4) validation webhook errors.
5) runtime issues with statefulset updates. 

 Alerts should be actionable for user.

 If there is not enough metrics, add needed.

f41gh7:
an example of alerts
```yaml

    - name: operator
      rules:
        - alert: logErrors
          expr: rate(operator_log_messages_total{level="error"}) > 0
          for: 15m
          labels:
            severity: high
          annotations:
            description: "operator has too many errors at logs: {{ $value}} check it logs"
            dashboard: "{{ $externalURL }}/d/1H179hunk/victoriametrics-operator?ds={{ $labels.dc }}&orgId=1"
        - alert: reconcileErrors
          expr: rate(controller_runtime_reconcile_errors_total{job="vm-operator-victoria-metrics-operator"}) > 0
          for: 10m
          labels:
            severity: high
          annotations:
            description: "operator cannot parse response from k8s api server, possible bug: {{ $value}}, check it logs"
            dashboard: "{{ $externalURL }}/d/1H179hunk/victoriametrics-operator?ds={{ $labels.dc }}&orgId=1"
        - alert: highQueueDepth
          expr: workqueue_depth{job="vm-operator-victoria-metrics-operator"} > 10
          for: 15m
          labels:
            severity: high
          annotations:
            description: "operator cannot handle reconciliation load for controller: {{ $labels.name }} current depth: {{ $value }}"
            dashboard: "{{ $externalURL }}/d/1H179hunk/victoriametrics-operator?ds={{ $labels.dc }}&orgId=1&viewPanel=20"
```

PR #828, opened 2023-12-20
Add alerting rules for operator (#526)

Amper:
#526

### DIFF
--- a/config/alerting/vmoperator-rules.yaml
+++ b/config/alerting/vmoperator-rules.yaml
@@ -7,40 +7,63 @@ spec:
     - name: vmoperator
       rules:
         - alert: LogErrors
-          expr: sum(rate(operator_log_messages_total{level="error", job=~".*((victoria.*)|vm)-?operator"}[5m])) > 0
+          expr: |
+            sum(
+              rate(
+                operator_log_messages_total{
+                  level="error",job=~".*((victoria.*)|vm)-?operator"
+                }[5m]
+              )
+            ) > 0
           for: 15m
           labels:
-            severity: high
+            severity: warning
             show_at: dashboard
           annotations:
             description: "Operator has too many errors at logs: {{ $value}}, check operator logs"
             dashboard: "{{ $externalURL }}/d/1H179hunk/victoriametrics-operator?ds={{ $labels.dc }}&orgId=1&viewPanel=16"
             summary: "Too many errors at logs of operator: {{ $value}}"
         - alert: ReconcileErrors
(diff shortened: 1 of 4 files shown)


Haleygo (approved):
lgtm

config/alerting/vmoperator-rules.yaml
```diff
@@ -0,0 +1,48 @@
...
+  name: vmoperator-rules
+spec:
+  groups:
+    - name: vmoperator
+      rules:
+        - alert: LogErrors
+          expr: sum(rate(operator_log_messages_total{level="error", job=~".*((victoria.*)|vm)-?operator"}[5m])) > 0
+          for: 15m
+          labels:
+            severity: high
```
Haleygo: Should we use the same set of `severity` like [other](https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/deployment/docker/alerts-cluster.yml) VictoriaMetrics rules here, like "critical"?

Amper: I don't think that it's `critical`, because it doesn't directly affect the deployments and monitoring work, but you are right, we have to use the same severity names. Maybe change to `warning`, wdyt? (+ @f41gh7)

config/alerting/vmoperator-rules.yaml
```diff
@@ -0,0 +1,48 @@
...
+          for: 10m
+          labels:
+            severity: high
+            show_at: dashboard
+          annotations:
+            description: "Operator cannot parse response from k8s api server, possible bug: {{ $value }}, check operator logs"
+            dashboard: "{{ $externalURL }}/d/1H179hunk/victoriametrics-operator?ds={{ $labels.dc }}&orgId=1&viewPanel=10"
+            summary: "Too many errors at reconcile loop of operator: {{ $value}}"
+        - alert: HighQueueDepth
+          expr: (sum(workqueue_depth{job=~".*((victoria.*)|vm)-?operator"}) by (name)) > 10
```
f41gh7: Based on my experience, workqueue_depth for vmuser and vmalertmanager config can be slow and trigger given alert. I propose to exclude it from matching.

 vmuser reconcile is slow because of additional secret creation. The same for alertmanagerconfig

config/alerting/vmoperator-rules.yaml
f41gh7: I'm fine with warning.

Amper: done

config/alerting/vmoperator-rules.yaml
Amper: done

f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #839, opened 2024-01-02
Update api.md

bxy4543:
`selectAllScrapes: true` seems to be changed to `selectAllByDefault: true`


Haleygo:
Hello, thanks for the fix!
But the api.md is generated by types struct and need to be updated by `make doc`, please check https://github.com/VictoriaMetrics/operator/pull/840 instead.

bxy4543:
ok

PR #842, opened 2024-01-03
follow up https://github.com/VictoriaMetrics/operator/commit/991536d

Haleygo:
follow-up https://github.com/VictoriaMetrics/operator/commit/991536d212ab9dc2d81f82260d89aae877f6ba28
add `revisionHistoryLimitCount` for all the workload CRD resources, previously only for deployments.
remove dup config in validating webhook, update docs.


f41gh7 (approved):
LGTM

PR #843, opened 2024-01-05
fix user selector for vmauth quick-start

Haleygo:
Should select the demo vmuser by default.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -377,6 +377,7 @@ kind: VMAuth
 metadata:
   name: demo
 spec:
+  selectAllByDefault: true
   userNamespaceSelector: {}
   userSelector: {}
   ingress:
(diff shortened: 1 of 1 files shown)


Haleygo:
PTAL @Amper @f41gh7

docs/quick-start.md
```diff
@@ -377,6 +379,7 @@ kind: VMAuth
 metadata:
   name: demo
 spec:
+  selectAllByDefault: true
```
Amper, line 382: I suppose that `selectAllByDefault: true` and `userNamespaceSelector: {}, userSelector: {}` are doing the same thing, aren't it?

f41gh7: yes, it could be skipped if selectAllByDefault is defined.

f41gh7 (approved):
LGTM

f41gh7:
> PTAL @Amper @f41gh7

sorry for delay

PR #852, opened 2024-01-23
docs/resources: fix link text for alertmanager API docs

jaskeerat789:
The link text of AlertManager API docs was set to `VMAlert` instead of `VMAlertmanager`


Haleygo (approved):
LGTM!
Thank you for the fix!

PR #854, opened 2024-01-23
fix docs CI

Amper:
follow up after https://github.com/VictoriaMetrics/operator/commit/7965dba77210a8a62712fd196a10f9c18db9a8ec


f41gh7 (approved):
LGTM

Issue #864, opened 2024-02-06
operator api usage deprecated autoscaling/v2beta2 blocks update of Azure Kubernetes Service


SaschaJohn:
Hello VM team,

we're currently doing an upgrade from AKS 1.25 to 1.27. During the upgrade, the service scans for usage of deprecated APIs, that will be removed in the current upgrade path. 
We came across the error of autoscaling/v2beta2 used from vmoperator which is deprecated since 1.23 and was removed in 1.26.

You already made the operator compatible with https://github.com/VictoriaMetrics/operator/issues/583

However I see a problem in the order of the API usage.

1,25 supports API version supports v2beta2, therefore function IsHPAV2BetaSupported returns true
and https://github.com/VictoriaMetrics/operator/blob/bfc521d3b22d696c2244318a2d93a5fee7a93452/controllers/factory/builders.go#L405
takes the beta in favor of the stable version. In 1.26 the beta api was removed. This blocks the upgrade path.

Do you any need to support the beta api? Otherwise I suggest to remove it from the controllers factory builders,

Best

/Sascha

f41gh7:
Hello, thanks for detailed bug report @Haleygo  can you please take a look?

SaschaJohn:
Bug fixed in next release after v0.41.1 from 1 Feb 2024.
Thank you vermy much @f41gh7 @Haleygo !

Issue #758, opened 2023-09-20
VMOperator drops Ingress created without using it


thunderbird86:
### Describe the bug

VMOperator drops Ingress created without using it if it has the same name as an operator-managed resource. Probably it can be related not only to Ingress but to other resources too

### To Reproduce

Deploy resource using VMOperater e.g. vmauth
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAuth
metadata:
  name: vmauth-sample
spec:
  replicaCount: 1
  resources:
    limits:
      cpu: 500m
      memory: 850Mi
    requests:
      cpu: 250m
      memory: 350Mi
  selectAllByDefault: true
```
Deploy Ingress using kubectl
```
---
apiVersion: v1
kind: Service
metadata:
  name: vmauth-sample
spec:
  type: ClusterIP
  ports:
    - name: service
      port: 8427
      protocol: TCP
      targetPort: 8427
  selector:
    app.kubernetes.io/name: vmauth
```
### Expectation

Both resources work fine

### Actual Result

VMOperator drops ingress

f41gh7:
It would be great to store dynamic linked resources ( ingress, hpa, psp, pdb, rbac policy) at resource annotation. And perform 3-way merge ( new config state, cluster config state, annotations config state) like kubectl does.

user81230:
Finally, I found the cause of VMOperator dropping my ingresses. This affected us too, and renaming ingresses solved the problem. Maybe it is a good practice to give all the resources unique names, regardless of their kind.

I suggest parsing Kubernetes resources together with their name (like "ingress/vmauth-sample" instead of just "vmauth-sample") in the operator's internals.

f41gh7:
Bug was fixed  at [v0.48.3](https://github.com/VictoriaMetrics/operator/releases/tag/v0.48.3) release. 

 Now operator will perform `Ingress` delete operation only if it was defined at `VMAuth` spec.

PR #865, opened 2024-02-07
remove deprecated autoscaling/v2beta1 HPA objects

Haleygo:
address https://github.com/VictoriaMetrics/operator/issues/864,
since operator supports k8s version from 1.25 now, there is no need to have autoscaling/v2beta1.

### DIFF
--- a/controllers/factory/k8stools/version.go
+++ b/controllers/factory/k8stools/version.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	"k8s.io/api/autoscaling/v2"
-	"k8s.io/api/autoscaling/v2beta2"
 	"k8s.io/apimachinery/pkg/version"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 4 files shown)


f41gh7 (approved):
LGTM

controllers/factory/k8stools/version.go
```diff
@@ -79,12 +68,9 @@ func IsFSGroupChangePolicySupported() bool {
 	return false
 }
 
-// NewHPAEmptyObject returns HorizontalPodAutoscaler object for given kubernetes version
+// NewHPAEmptyObject returns HorizontalPodAutoscaler object
 func NewHPAEmptyObject(opts ...func(obj client.Object)) client.Object {
```
f41gh7, line 72: I think, we should remove `client.Object from this function and return exact type. It can be done at follow-up commit.

f41gh7:
Thanks!

f41gh7:
> @f41gh7 @Haleygo Deleting the old autoscaling/v2beta1 will affect subsequent version upgrades for k8s clusters that cannot be upgraded. Is it more appropriate to use compatible adaptation to solve this problem?

 Unfortunately, it's not possible to support EoL versions of kubernetes properly.

I think, soon we're going to redesign resource reconciliation a bit. It must fix possible errors with incompatible resources.

If resource don't use deprecated API directly, it shouldn't trigger any errors by calling deprecated kubernetes APIs.

 Related issue https://github.com/VictoriaMetrics/operator/issues/758#issuecomment-1729257683

PR #866, opened 2024-02-07
cleanup code for deprecated resources

Haleygo:
1. PodSecurityPolicy
2. policyv1beta1.PodDisruptionBudget, use only policyv1.PodDisruptionBudget(https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125)


api/v1beta1/vmagent_types.go
```diff
@@ -109,10 +109,6 @@ type VMAgentSpec struct {
 	// cannot be used with HostNetwork.
 	// +optional
 	HostAliases []v1.HostAlias `json:"host_aliases,omitempty"`
-	// PodSecurityPolicyName - defines name for podSecurityPolicy
-	// in case of empty value, prefixedName will be used.
-	// +optional
-	PodSecurityPolicyName string `json:"podSecurityPolicyName,omitempty"`
```
f41gh7, line 115: Is it safe to delete objects definition from CRD? 

maybe we should keep it as is and remove deprecated fields at `v1` version of API?

f41gh7: currently it could be no-op field

f41gh7 (approved):
LGTM

f41gh7:
Thanks for initial PR, all changes were ported to f1ac59e077b1e8991f074ec4739f9bf122155943

Issue #875, opened 2024-02-26
VictoriaMetrics operator should be cgroup aware


uhthomas:
[Go is not cgroup aware](https://github.com/golang/go/issues/33803). This means the VictoriaMetrics operator will be throttled hard in containerised environments with CPU limits.

The below screenshot shows the container with a CPU limit in Kubernetes. The first half of the time series is on a 13600k, with 20 threads. The second half is when the machine changed to use an EPYC 7763 with 128 threads.

![image](https://github.com/VictoriaMetrics/operator/assets/9749173/a300f5c4-0cf7-4b78-ad07-26194e754cfe)

[automaxprocs](https://github.com/uber-go/automaxprocs) can use cgroups automatically.

Haleygo:
Hello @uhthomas ,
AFAIK the operator code itself doesn't use muti-thread to process resource objects, so I'm not sure the performance benefit here by setting `GOMAXPROCS`.  Did you see some good results with new `GOMAXPROCS`?

f41gh7:
Hello, thanks for reporting the issue.

 Currently, operator doesn't use GOMAXPROCS for paralleling  CPU bound tasks. Actually, it doesn't perform any heavy CPU bound operations. And unlikely it will change in the future.

 Meanwhile the issue itself looks to me more like golang runtime problem, neither something that each application should fix on it's own.

PR #876, opened 2024-02-26
chore: use Linux container CPU quota

uhthomas:
Go is not cgroup aware and by default will set GOMAXPROCS to the number of available threads, regardless of whether it is within the allocated quota. This behaviour causes high amount of CPU throttling and degraded application performance.

Fixes: #875

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/prometheus/client_golang v1.18.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.4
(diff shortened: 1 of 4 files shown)


f41gh7:
WDYT about using https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/lib/cgroup/cpu.go ?

f41gh7:
Thank you for your time and PR. 

Closing it with comment https://github.com/VictoriaMetrics/operator/issues/875#issuecomment-2272229935

PR #885, opened 2024-03-01
fix(apis_test.go): fix relabeling config

incubator4:
my apologize for  vm relabeling is different from prometheus.  Due to this test case aim to probe, replace relabeling with minimal values.

cc @f41gh7


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

Issue #894, opened 2024-02-26
'notifier.url' always set even when 'notifier.blackhole' is set to true


hconnan:
Hello,

'notifier.url' parameter is not longer needed by default in Victoria Metrics.

When we would like to disable, we need to set the `notifier.blackhole` parameter to true. Since Victoria Metrics 1.96, when this parameter is set, we cannot set `notifier.url` parameter in the same time ([source](https://github.com/VictoriaMetrics/VictoriaMetrics/issues/5703#issuecomment-1913092237))
However, in the server-deployment template of the victoria-metric-alerts chart, the `notifier-url` is always set whatever the extra arguments.

How to reproduce it? Deploy VM alert with `notifier.blackhole` extraArgs set to `true`.

Fix suggestion: I suggest to add a condition to set the `notifier.url` parameter when there is no `-notifier.blackhole` or `-notifier.config` extra arguments.

hconnan:
All is good for me! Great job! Thank you very much 😄 🥳

PR #895, opened 2024-03-07
controllers/factory/vmalert: do not add `notifier.*` flags if `notifier.blackhole` is set by user

zekker6:
Previously, setting "notifier.blackhole" and not using any notifiers would lead to CrashLoopBackoff because vmalert would receive an empty "notifier.url" value.

Updates:
- https://github.com/VictoriaMetrics/operator/issues/894
- https://github.com/VictoriaMetrics/operator/pull/813


f41gh7 (approved):
LGTM

f41gh7:
I think, it'd be great to set blackhole as a default value if there is no any notifiers configured.

Issue #893, opened 2024-03-06
ServiceMonitor conversion drops "attachMetadata"


k1rk:
Hi
During automatic resource conversion by operator seems like data inside `attachMetadata` is lost

Following ServiceMonitor
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: node-exporter
  namespace: monitoring
spec:
  attachMetadata:
    node: true
  endpoints:
    - metricRelabelings:
        - action: drop
          regex: /var/lib/kubelet/pods.+
          sourceLabels:
            - mountpoint
      port: metrics
      scheme: http
      scrapeTimeout: 10s
  jobLabel: app.kubernetes.io/name
  selector:
    matchLabels:
      app.kubernetes.io/instance: node-exporter
      app.kubernetes.io/name: node-exporter
```

results in following VMServiceScrape
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMServiceScrape
metadata:
  name: node-exporter
  namespace: monitoring
spec:
  endpoints:
    - attach_metadata: {}
      metricRelabelConfigs:
        - action: drop
          regex: /var/lib/kubelet/pods.+
          sourceLabels:
            - mountpoint
      port: metrics
      scheme: http
      scrapeTimeout: 10s
  jobLabel: app.kubernetes.io/name
  namespaceSelector: {}
  selector:
    matchLabels:
      app.kubernetes.io/instance: node-exporter
      app.kubernetes.io/name: node-exporter
```

as you can see now `attach_metadata` is an empty object

no errors observed in VMOperator logs

f41gh7:
Must be fixed at [v0.43.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.43.0) release

PR #905, opened 2024-03-13
vmservicescrape&vmpodscrape: add `attach_metadata` option under VMSer…

Haleygo:
…viceScrapeSpec&VMPodScrapeSpec, the same way like prometheus serviceMonitor&podMonitor do.

The original `attach_metadata` option  under Endpoint has higher priority.

address https://github.com/VictoriaMetrics/operator/issues/893


f41gh7 (approved):
LGTM

Issue #902, opened 2024-03-12
suggestion: remove pods * permissions for vmagent controller 


ebensom:
It seems that the [`rbac:groups="",resources=pods,verbs=*`](https://github.com/VictoriaMetrics/operator/blob/de886759f741a50c148595645ad2c43a1599f571/controllers/vmagent_controller.go#L56) permission  is excessive and not being used by vm-operator.  
From K8S security point of view, it is advised to remove this permission if not needed, because a compromised vm-operator pod's SA has permission to create pods and exec into pods in all namespaces.

Haleygo:
Hi @ebensom , operator does need delete permission to performing roll update for some service, check https://github.com/VictoriaMetrics/operator/pull/907#discussion_r1535391190.
And the role in examples should all be fixed in https://github.com/VictoriaMetrics/operator/pull/907.

PR #907, opened 2024-03-22
chore: minor fixes

Haleygo:
1. follow up https://github.com/VictoriaMetrics/operator/commit/a54efbc647b0e099a583885afa263aa9fc2f9853, fix type struct field tags
2. remove unnecessary pod permission for operator, see https://github.com/VictoriaMetrics/operator/issues/902
3. remove fields `VMClusterSpec.VMInsert.Name`, `VMClusterSpec.VMStorage.Name`, `VMClusterSpec.VMSelect.Name`, they're marked as deprecated since v0.21.0 in https://github.com/VictoriaMetrics/operator/commit/141c82370da6f27fd32502dc14709e844978778a.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -40,12 +40,6 @@ rules:
 - apiGroups:
     - ""
   resources:
-    - pods
-  verbs:
-    - '*'
-- apiGroups:
-    - ""
-  resources:
     - secrets
     - secrets/finalizers
   verbs:
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -40,6 +40,12 @@ rules:
 - apiGroups:
     - ""
   resources:
+    - pods
+  verbs:
+    - '*'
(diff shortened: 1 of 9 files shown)


config/rbac/role.yaml
```diff
@@ -37,12 +37,6 @@ rules:
     - persistentvolumeclaims/finalizers
   verbs:
     - '*'
```
f41gh7: Could it break statefulset rollout update process? 

 It uses delete operation for pods.

https://github.com/VictoriaMetrics/operator/blob/master/controllers/factory/k8stools/sts.go#L201

Haleygo: Yes, you're right!
Sorry I forgot about this, fixed other insufficient roles in the latest commit, please check.

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #909, opened 2024-03-22
expose reconcile error message

Haleygo:
follow up https://github.com/VictoriaMetrics/operator/commit/403a78ac8815a3fddf1666e383d885266fcfdc2d

### DIFF
--- a/controllers/vmagent_controller.go
+++ b/controllers/vmagent_controller.go
@@ -120,6 +120,7 @@ func (r *VMAgentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 		return result, nil
 	})
 	if err != nil {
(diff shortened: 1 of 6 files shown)


controllers/vmagent_controller.go
```diff
@@ -120,6 +120,7 @@ func (r *VMAgentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 		return result, nil
 	})
 	if err != nil {
+		reqLogger.Error(err, "failed to reconcile vmagent")
```
f41gh7, line 123: I think, this error must be tracked and logged by reconcile controller.

Haleygo:
Missed the error logs printed by reconcile controller before, close as false alarm.

Issue #808, opened 2023-11-21
Support discord_configs in alertmanager config


recht:
Alertmanager supports discord_configs in receivers, so it would be nice to have that reflected in VM too.

bh-tt:
Additionally, I'm missing: msteams_configs, sns_configs and webex_configs.

bh-tt:
@Haleygo I'm willing to implement this, do you have an overview of what needs to be changed?

I suppose I need to add the new receiver types [here](https://github.com/VictoriaMetrics/operator/blob/50ae201687aace71b2e1a0528301caf39c7ac154/api/victoriametrics/v1beta1/vmalertmanagerconfig_types.go#L263) and then add something to the alertmanagerconfig generation code [here](https://github.com/VictoriaMetrics/operator/blob/50ae201687aace71b2e1a0528301caf39c7ac154/controllers/factory/alertmanager/config.go#L390)?

Haleygo:
> @Haleygo I'm willing to implement this, do you have an overview of what needs to be changed?
> 
> I suppose I need to add the new receiver types [here](https://github.com/VictoriaMetrics/operator/blob/50ae201687aace71b2e1a0528301caf39c7ac154/api/victoriametrics/v1beta1/vmalertmanagerconfig_types.go#L263) and then add something to the alertmanagerconfig generation code [here](https://github.com/VictoriaMetrics/operator/blob/50ae201687aace71b2e1a0528301caf39c7ac154/controllers/factory/alertmanager/config.go#L390)?

@bh-tt Yes, that's about it)

bh-tt:
I've tried msteams_configs today and it works!

PR #912, opened 2024-03-26
bump alertmanager to v0.27.0 to support new receivers

Haleygo:
The new MSTeamsConfig receiver added in https://github.com/VictoriaMetrics/operator/issues/808 is valid since alertmanager v0.26.0, so need to bump the default version.


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #914, opened 2024-04-02
Makefile: update alpine for 1.19.1 to CVE fixes

f41gh7:


Amper (approved):
LGTM, but i think also changelog need to be updated

Amper (approved):
LGTM

Issue #932, opened 2024-04-22
bearerTokenSecret missing on Conversion from ServiceMonitor to VMServiceScrape since 0.43.0


marrau:
When Upgrading from 0.42.4 to 0.43.1 any ServiceMonitor with a bearerTokenSecret set, gets Permission denied on scraping.

Example:
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: test-service
spec:
  endpoints:
  - bearerTokenSecret:
      key: PRIVATE_API_KEY
      name: test-secret
    honorLabels: false
    interval: 15s
    path: /metrics
    scrapeTimeout: 15s
    targetPort: 80
  jobLabel: test
  selector:
    matchLabels:
      test: test
```

Result on 0.43.+ (shortened Output):
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMServiceScrape
metadata:
  name: test-service
spec:
  attach_metadata: {}
  endpoints:
  - attach_metadata: {}
    interval: 15s
    path: /metrics
    scrapeTimeout: 15s
    targetPort: 80
  jobLabel: test
  namespaceSelector: {}
  selector:
    matchLabels:
      test: test
```

Result on 0.42.4 (shortened Output):
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMServiceScrape
metadata:
  name: test-service
spec:
  endpoints:
  - attach_metadata: {}
    bearerTokenSecret:
      key: PRIVATE_API_KEY
      name: test-secret
    interval: 15s
    path: /metrics
    scrapeTimeout: 15s
    targetPort: 80
  jobLabel: test
  namespaceSelector: {}
  selector:
    matchLabels:
      test: test
```

marrau:
Found it to be removed in [Commit 0500abb](https://github.com/VictoriaMetrics/operator/commit/0500abb3625cb1ecc045eac2a8032b16384c5ed6#diff-6abd928d6b2d8688a50683817e479df19ad03146c6b14c7e856b7c8b43b78508L354)

f41gh7:
Looks like bug to me, @Haleygo  can you confirm?

Haleygo:
Hi @marrau ,
Thanks for the issue, it should be fixed in https://github.com/VictoriaMetrics/operator/pull/935.
[...]

PR #935, opened 2024-04-23
Operator fixes:

Haleygo:
1. address https://github.com/VictoriaMetrics/operator/issues/932, fix `ServiceMonitor` to `VMServiceScrape` conversion, before `bearerTokenSecret` is dropped by mistake since v0.43.0
2. fix selector match for config resources like VMUser, VMRule... , before it could be ignored when update resource labels.


f41gh7 (approved):
LGTM

PR #938, opened 2024-04-24
Fix: docs with new file

Davidffry:
Issue with your website generator, It does not take the README.md file but in markdown is ok but not with your website.
So I duplicate the README.md (I not delete it ) to vmcustomresources.md
And change the redirection into the docs.


Davidffry:
Another question : 
it is Victoria[_space_]Metrics Or VictoriaMetrics ?

Haleygo:
I'm not sure having another new vmcustomresources.md is the best solution here, since this means we need to maintain both files.
As i can see, the link `./resources/README.md` in `docs/README.md` is available, while same link `./resources/README.md` in `docs/quick-start.md` is invalid.
@tenmozes @Amper Do we have a better way to fix this?

Haleygo:
> Another question : it is Victoria[_space_]Metrics Or VictoriaMetrics ?

Should be VictoriaMetrics)

AndrewChubatiuk:
hey @Davidffry
this issue was fixed and will be available with a new release of operator

PR #940, opened 2024-04-25
controllers/reconcile: adds waitReady function for statefulset

f41gh7:
it ensures, that all pods for statefulset is ready to serve requests. It must fix an issue when newly created cluster has operational status, when in fact pod is still not at ready state.

Adds more context to the wait timeout error. Reason of pod not to be ready. It must help to debug possible issues with pod configuration.

Deletes few typos.


Amper (approved):
LGTM

PR #949, opened 2024-05-07
vmscrapeconfig: `authorization` section in sd configs works properly …

Amper:
…with empty `type` field (default value for this field is `Bearer`)

related links:
- Thread in slack: https://victoriametrics.slack.com/archives/CGZF1H6L9/p1714925950145659
- Description in our API docs: https://docs.victoriametrics.com/operator/api/#authorization
- Description in prometheus operator API docs: https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1.Authorization

The default value for this field should work like "Bearer".


f41gh7 (approved):
LGTM

PR #952, opened 2024-05-08
controllers/prometheus-converter: fix owner reference type for VMScrapeConfig

jfroy:
This patches fixes the owner reference type for VMScrapeConfig objects. Without this patch, on many clusters, newly-created VMScrapeConfig objects would immediately be deleted because the owner reference would not resolve.


Amper (approved):
LGMT, thanks for your contribution

Issue #954, opened 2024-05-16
Operator duplicates values from VMAlertManagerConfig to config secret


mariapoliss:
Hello!

I found a problem with updating Secret with VM Alert Manager config. 
I deployed the last release version v0.44.0 for this moment.
For example, if I create VMAlertManagerConfig like this:

```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAlertmanagerConfig
metadata:
  name: test-amc
spec:
  inhibit_rules:
    - equal:
        - alertname
      source_matchers:
        - severity="critical"
      target_matchers:
        - severity="warning"
    - source_matchers:
        - alertname="QuietWeeklyNotifications"
      target_matchers:
        - alert_group="l2ci_weekly"
  receivers:
    - name: l2ci_receiver
      webhook_configs:
        - url: http://notification_stub_ci1:8080
    - name: blackhole
    - name: ca_em_receiver
      webhook_configs:
        - url: http://notification_stub_ci2:8080
  route:
    group_by:
      - alertname
      - l2ci_channel
    routes:
      - matchers:
          - alertname="QuietWeeklyNotifications"
        receiver: blackhole
      - matchers:
          - alertname="QuietDailyNotifications"
        receiver: blackhole
      - matchers:
          - alert_group=~"^l2ci.*"
        receiver: l2ci_receiver
```

Victoria Metrics operator watches for `test-amc`, updates secret with name set in VMAlertManager (spec.configSecret) and makes it every reconcile cycle.

After updating the secret I see duplicated values (receivers, inhibit_rules and route.routes) in secret data `alertmanager.yaml`:

```
route:
  receiver: default-test-amc-blackhole
  routes:
    - routes:
        - matchers:
            - alertname="QuietWeeklyNotifications"
          receiver: default-test-amc-blackhole
          continue: false
        - matchers:
            - alertname="QuietDailyNotifications"
          receiver: default-test-amc-blackhole
          continue: false
        - matchers:
            - alert_group=~"^l2ci.*"
          receiver: default-test-amc-l2ci_receiver
          continue: false
      matchers:
        - namespace = "default"
      group_by:
        - alertname
        - l2ci_channel
      receiver: default-test-amc-blackhole
      continue: true
    - routes:
        - matchers:
            - alertname="QuietWeeklyNotifications"
          receiver: default-test-amc-blackhole
          continue: false
        - matchers:
            - alertname="QuietDailyNotifications"
[...]

PR #955, opened 2024-05-16
vmalertmanager: ignore content of configSecret if name clashes with default

f41gh7:
Ignores any content from secret define at cr.spec.configSecret, if it name clashes with name of the secret created by operator.

https://github.com/VictoriaMetrics/operator/issues/954

### DIFF
--- a/controllers/factory/alertmanager/statefulset.go
+++ b/controllers/factory/alertmanager/statefulset.go
@@ -452,6 +452,10 @@ func createDefaultAMConfig(ctx context.Context, cr *victoriametricsv1beta1.VMAle
 	switch {
 	// fetch content from user defined secret
 	case cr.Spec.ConfigSecret != "":
+		if cr.Spec.ConfigSecret == cr.ConfigSecretName() {
+			l.Info("ignoring content of ConfigSecret, since it has the same name as secreted created by operator for config", "secretName", cr.Spec.ConfigSecret)
+			break
+		}
 		// retrieve content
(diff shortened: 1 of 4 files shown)


controllers/factory/alertmanager/statefulset.go
```diff
@@ -452,6 +452,10 @@ func createDefaultAMConfig(ctx context.Context, cr *victoriametricsv1beta1.VMAle
 	switch {
 	// fetch content from user defined secret
 	case cr.Spec.ConfigSecret != "":
+		if cr.Spec.ConfigSecret != cr.ConfigSecretName() {
```
Haleygo: I believe it should be `if cr.Spec.ConfigSecret == cr.ConfigSecretName() `?

f41gh7: Good catch! I incorrectly copy-pasted line.

Haleygo:
But this also means we will override the original secret contents that user had, should we avoid this by changing default secret name if conflict happens?

f41gh7:
> But this also means we will override the original secret contents that user had, so the wrost scenario is: user predefined all the configs under secret `vmalertmanager-example-config` and specified secret in vmalertmanager, now operator will ignore the configs and erase them. What about trying merging configs or changing default secret name if conflict happens?

 There is no easy way to check an owner of this secret. It's possible to use heuristic and check labels. But it's not reliable solution. 

 I think, it's fine to overwrite user created secret and notify about it. And it ll be more predictable behavior.

Haleygo (approved):
LGTM

Issue #953, opened 2024-05-15
handle non zero deletionTimestamp for child objects.


f41gh7:
Currently, if by mistake ServiceAccount ( or any other object) was deleted, operator doesn't remove finalizer for it. Since parent object wasn't marked as deleted. It blocks kubernetes control-plain operations. 

 Proposed solution:
1) check for deletion timestamp.
2) if timestamp is not zero, remove finalizer and raise an error
3) object must be recreated and the next reconcile loop.

jmleddy:
Here's the reproducer if needed:

```
james@M-D19FP4G4QL ~/src/kube-resources $ kubectl get sa vmagent-az1
NAME          SECRETS   AGE
vmagent-az1   0         95d
james@M-D19FP4G4QL ~/src/kube-resources $ kubectl get sa vmagent-az1 -o yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  creationTimestamp: "2024-02-09T15:55:25Z"
  finalizers:
  - apps.victoriametrics.com/finalizer
<snip>
james@M-D19FP4G4QL ~/src/kube-resources $ kubectl delete sa vmagent-az1
serviceaccount "vmagent-az1" deleted
^Z
[1]+  Stopped                 kubectl delete sa vmagent-az1
james@M-D19FP4G4QL ~/src/kube-resources $ kubectl get sa vmagent-az1  -o yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  creationTimestamp: "2024-02-09T15:55:25Z"
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2024-05-15T12:56:27Z"
  finalizers:
  - apps.victoriametrics.com/finalizer
<snip
james@M-D19FP4G4QL ~/src/kube-resources $ kubectl get pods
NAME            READY   STATUS    RESTARTS   AGE
vmagent-az1-0   3/3     Running   0          6d15h
vmagent-az2-0   3/3     Running   0          6d15h
james@M-D19FP4G4QL ~/src/kube-resources $ kubectl rollout restart sts vmagent-az1
statefulset.apps/vmagent-az1 restarted
james@M-D19FP4G4QL ~/src/kube-resources $ kubectl get pods
NAME            READY   STATUS               RESTARTS      AGE
vmagent-az1-0   1/3     PostStartHookError   0 (13s ago)   89s
vmagent-az2-0   3/3     Running              0             6d16h
[...]

PR #956, opened 2024-05-20
controllers/finalize: Remove finalizer for child objects for non zero DeletionTimestamp

f41gh7:
Kubernetes performs soft delete and waits for object hard delete until finalizers == 0. During that period any actions for soft deleted objects aren't performend. It caues weird behavior for service accounts, deployments and etc. When kubernetes-controller manager ignores actions that must be performend with it. For instance, pod creation for soft deleted deployment.

 Operator now detects soft delete and free objects. An object ll be recreate at the next reconcile loop and error message ll be logged.

https://github.com/VictoriaMetrics/operator/issues/953

### DIFF
--- a/controllers/factory/alertmanager/statefulset.go
+++ b/controllers/factory/alertmanager/statefulset.go
@@ -514,6 +514,13 @@ func createDefaultAMConfig(ctx context.Context, cr *victoriametricsv1beta1.VMAle
 		}
 		return err
 	}
+	if err := finalize.FreeIfNeeded(ctx, rclient, &existAMSecretConfig); err != nil {
+		return err
+	}
+
+	if err := finalize.FreeIfNeeded(ctx, rclient, &existAMSecretConfig); err != nil {
+		return err
+	}
 
 	newAMSecretConfig.Annotations = labels.Merge(existAMSecretConfig.Annotations, newAMSecretConfig.Annotations)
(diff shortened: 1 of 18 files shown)


controllers/factory/alertmanager/statefulset.go
```diff
@@ -514,6 +514,13 @@ func createDefaultAMConfig(ctx context.Context, cr *victoriametricsv1beta1.VMAle
 		}
 		return err
 	}
+	if err := finalize.FreeIfNeeded(ctx, rclient, &existAMSecretConfig); err != nil {
+		return err
+	}
+
+	if err := finalize.FreeIfNeeded(ctx, rclient, &existAMSecretConfig); err != nil {
```
Haleygo: dup codes)

controllers/factory/vmsingle/vmsingle.go
```diff
@@ -42,7 +42,9 @@ func CreateVMSingleStorage(ctx context.Context, cr *victoriametricsv1beta1.VMSin
 			if err := rclient.Create(ctx, newPvc); err != nil {
 				return fmt.Errorf("cannot create new pvc for vmsingle: %w", err)
 			}
-
+			if !existPvc.DeletionTimestamp.IsZero() {
```
Haleygo, line 45: should be outside of
```
if errors.IsNotFound(err) {
```

f41gh7: There was unhandeled error and it broke changes line.

Issue #957, opened 2024-05-21
skip storageClass check if there is no PVC size change


shichanglin5:
<img width="1228" alt="image" src="https://github.com/VictoriaMetrics/operator/assets/48707349/42aa58c6-03b7-45c9-91fc-eb57c187f26c">

after helm upgrade : "storage class for PVC doesn't support live resizing" begin appear in the log
**I have not modified the size of pvc, this kind of log should not appear**

问题分析：
1. should first judge whether the size of pvc changes
2. then judge whether expand is supported, 
3. but now the judgment of whether pvc size changes is in the following growPVCs.

<img width="1062" alt="image" src="https://github.com/VictoriaMetrics/operator/assets/48707349/23cdf406-2620-441c-9e59-a74ea4280875">

<img width="1099" alt="image" src="https://github.com/VictoriaMetrics/operator/assets/48707349/dc8859ec-c93d-42b9-bcad-3704190841d8">

PR #958, opened 2024-05-21
skip storageClass check if there is no PVC size change

shichanglin5:
fix https://github.com/VictoriaMetrics/operator/issues/957


shichanglin5:
> Thank you for the fix! Could you also help adding a description to the [changelog](https://github.com/VictoriaMetrics/operator/blob/master/docs/CHANGELOG.md#L0-L1), something like
> 
> > Skip storageClass check if there is no PVC size change

done

Haleygo (approved):
LGTM

Issue #962, opened 2024-05-23
[BUG] Failing VMAuth/VMUser CR's after 0.42.0 Release - On Manually Created 


seanocca:
This bug has been an error since the [0.42.0 release](https://github.com/VictoriaMetrics/operator/releases/tag/v0.42.0). 
When you create a VMAuth and VMUser CR the operator fails to connect the managed secret to the deployment. The error is 
`cannot create or update vmauth deploy: cannot update deployment for app: vmauth-main, err: Deployment.apps "vmauth-main" is invalid: spec.template.spec.containers[0].volumeMounts[1].name: Not found: "config"`

The VMAuth CR creates the secret but the deployment searches for a configSecretRef value that does not exist.

This requires us to import and create our own secret with the potential to leak our username and password.

Why can we set the username, password, or passwordRef (with a password imported from another secret) on the VMUser CR, if those values don't build a secret that is to be used by the VMAuth deployment?

We have implemented the suggested work around by creating our own secret with the `config.yaml` file in the secret and changing the VMAuth `configSecret` setting to point to that. However the old method of the operator creating the secret and managing it is much more preferred. As it also won't require further config-reloader settings to be required.

Steps to reproduce:
Operator Deploy method: VM Operator Helm Chart (0.31.1) into Kubernetes (1.28)
VMAuth/VMUser Deployment Method: Manual CR's created and deployed

Haleygo:
Hello @seanocca ,
I can't seem to reproduce it by creating vmauth like
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAuth
metadata:
  name: test
spec:
  extraArgs:
    discoverBackendIPs: "true"
    logInvalidAuthTokens: "true"
  image:
    tag: v1.101.0
  replicaCount: 1
  unauthorizedAccessConfig:
  - src_paths:
    - /insert/.*
    url_prefix:
    - http://vminsert:8480
```
Could you please provide an example here?

seanocca:
Hi @Haleygo 
This is the config we are using to deploy the VMUser and VMAuth

```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAuth
metadata:
  name: vmauth-main
spec:
  selectAllByDefault: true
  ingress:
    host: your.url.com    
  image:
    repository: victoriametrics/vmauth
    pullPolicy: IfNotPresent
    tag: v1.101.0
  replicaCount: 1
  configSecret: entain-venues-vmauth
  resources:
[...]

PR #964, opened 2024-05-23
vmauth: don't mount config in the vmauth container when using the cus…

jfroy:
…tom reloader

When the custom reloader is in use, there is no config volume defined and the vmauth container cannot start with a config volume mount.

This patch also adds basic tests for the vmauth pod spec.

### DIFF
--- a/controllers/factory/vmauth/vmauth_test.go
+++ b/controllers/factory/vmauth/vmauth_test.go
@@ -7,6 +7,8 @@ import (
 	victoriametricsv1beta1 "github.com/VictoriaMetrics/operator/api/v1beta1"
 	"github.com/VictoriaMetrics/operator/controllers/factory/k8stools"
 	"github.com/VictoriaMetrics/operator/internal/config"
+	"github.com/google/go-cmp/cmp"
+	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


Haleygo:
address https://github.com/VictoriaMetrics/operator/issues/962

controllers/factory/vmauth/vmauth_test.go
```diff
@@ -154,3 +156,304 @@ func TestCreateOrUpdateVMAuth(t *testing.T) {
...
+	f := func(t *testing.T, cr *victoriametricsv1beta1.VMAuth, c *config.BaseOperatorConf, wantYaml string) {
+		t.Helper()
+		setDefaultForVMAuth(cr, c)
+
+		// this trick allows to omit empty fields for yaml
+		var wantSpec corev1.PodSpec
+		if err := yaml.Unmarshal([]byte(wantYaml), &wantSpec); err != nil {
+			t.Fatalf("not expected wantYaml: %q: \n%q", wantYaml, err)
+
+			setDefaultForVMAuth(cr, c)
```
Haleygo: Already set in L163.

Haleygo (commented):
LGTM, please see one comment.

Haleygo (approved):
Thanks for contribution!

PR #965, opened 2024-05-28
prometheus-converter: removed dependence on getting the list of API resources for all API groups in the cluster...

Amper:
… (including those that are not used by the operator). Now API resources are requested only for the required groups (monitoring.coreos.com/*).

### DIFF
--- a/controllers/vmprometheusconverter_controller.go
+++ b/controllers/vmprometheusconverter_controller.go
@@ -235,31 +235,20 @@ func waitForAPIResource(ctx context.Context, client discovery.DiscoveryInterface
 	for {
 		select {
 		case <-tick.C:
-			groups, err := client.ServerGroups()
+			api, err := client.ServerResourcesForGroupVersion(apiGroupVersion)
 			if err != nil {
-				l.Error(err, "cannot get server groups")
+				if !errors.IsNotFound(err) {
+					l.Error(err, "cannot get  server resource for api group version")
+				}
 				continue
 			}
-			for _, g := range groups.Groups {
-				for _, v := range g.Versions {
-					if v.GroupVersion != apiGroupVersion {
-						continue
-					}
-					api, err := client.ServerResourcesForGroupVersion(apiGroupVersion)
-					if err != nil {
-						l.Error(err, "cannot get  server resource for api group version")
-						continue
-					}
-					for _, r := range api.APIResources {
-						if r.Kind != kind {
-							continue
-						}
-						l.Info("api resource is ready")
-						return nil
-					}
+			for _, r := range api.APIResources {
+				if r.Kind != kind {
+					continue
 				}
+				l.Info("api resource is ready")
+				return nil
 			}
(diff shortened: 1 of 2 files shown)


controllers/vmprometheusconverter_controller.go
```diff
@@ -235,17 +235,27 @@ func waitForAPIResource(ctx context.Context, client discovery.DiscoveryInterface
...
-				if apiList.GroupVersion == apiGroupVersion {
-					for _, r := range apiList.APIResources {
-						if r.Kind == kind {
-							l.Info("api resource is ready")
-							return nil
+			for _, g := range groups.Groups {
+				for _, v := range g.Versions {
+					if v.GroupVersion != apiGroupVersion {
+						continue
+					}
```
Haleygo: I think we don't need this check at all, it doesn't provide extra info than `client.ServerResourcesForGroupVersion(apiGroupVersion)`. 
`apiGroupVersion` is already passed by caller and can be used in `api, err := client.ServerResourcesForGroupVersion(apiGroupVersion)` directly.

Amper: The reason for advance calling `ServerGroups` is that calling `ServerResourcesForGroupVersion` for a group missing in the cluster causes an error - the point of this piece of code is to wait for the group to appear in the cluster.

Haleygo: > wait for the group to appear in the cluster.

If group is not there yet, `client.ServerResourcesForGroupVersion` should also give error, no?

Amper: > If group is not there yet, client.ServerResourcesForGroupVersion should also give error, no?

Generally yes, but in this case `ServerResourcesForGroupVersion` will be called only for those groups returned by `ServerGroups`, i.e. known to exist in the cluster.

Haleygo: hmm, I still don't see the harm of calling `ServerResourcesForGroupVersion` directly. 
If `monitoring.coreos.com` apiGroup is not exist, we can just ignore the error of `ServerResourcesForGroupVersion`, as user may not using prometheus CRDs. 
```
		case <-tick.C:
			api, err := client.ServerResourcesForGroupVersion(apiGroupVersion)
			if err != nil {
				if !strings.Contains(err.Error(), "could not find the requested resource") {
					l.Error(err, "cannot get server resource for api group version")
				}
				continue
			}
			for _, r := range api.APIResources {
				if r.Kind != kind {
					continue
				}
				l.Info("api resource is ready")
				return nil
			}
```

Amper: I'm not a fan of using errors in this way to create business logic, but it would make the code simpler, ok, I'll make the fix and test it today.

Amper: Fixed

Haleygo (approved):
LGTM!

PR #966, opened 2024-05-30
auto sync crd resources to chart repo

Haleygo:


.github/workflows/crds.yaml
```diff
@@ -0,0 +1,59 @@
+name: Update CRDs for chart repo
+
+on:
+  push:
+    branches:
+      - master
```
f41gh7: WDYT about sync only at release tag?

Haleygo: Sounds better, updated

Haleygo:
PTAL @Amper @zekker6

.github/workflows/crds.yaml
```diff
@@ -0,0 +1,56 @@
+name: Update CRDs for chart repo
+
+on:
+  release:
+    types: [published]
```
Amper, line 5: Should we set filter to avoid RC tags here?

Haleygo: We could, but i haven't seen any rc tag, should we filter out with `-rc` suffix?

Amper: Hm... it's weird, i definitely remember that @f41gh7 sometimes releases RC versions to test them in our environments. Maybe he deletes them afterwards.

Amper: https://hub.docker.com/r/victoriametrics/operator/tags?page=&page_size=&ordering=&name=rc

Haleygo: thanks, updated

PR #972, opened 2024-06-10
Add docs and example for multi-level cluster setup

Amper:


docs/resources/vmcluster.md
```diff
@@ -630,6 +630,12 @@ Possible configuration options for backup crd can be found at [link](../api.md#v
 
 Also see VMCLuster example spec [here](https://github.com/VictoriaMetrics/operator/blob/master/config/examples/vmcluster_with_backuper.yaml).
 
+## Multi-level cluster setup
```
tenmozes, line 633: should we add a picture here?

Amper:
I did it wrong, I need to redo it so I put it in the draft for a while

docs/resources/vmcluster.md
Amper: Ok, I'll add picture

PR #981, opened 2024-06-17
yq v2 -> v4 upgrade

AndrewChubatiuk:
upgraded yq to version, which has arm64 docker images
diff is huge, but only ident was changed in yaml files


f41gh7:
Thanks for contribution!

PR #984, opened 2024-06-19
controllers/vmalertmanagerconfig: adds missing err handle callback

f41gh7:
Fixes missing err handkle func after bc09a9b


Haleygo (approved):
LGTM

PR #988, opened 2024-06-23
replaced yq with kustomize

AndrewChubatiuk:
use built-in facilities to patch CRDs, which is much faster and simplier than using yq


f41gh7 (approved):
LGTM

f41gh7:
Thanks! Looks really cool

PR #989, opened 2024-06-25
removed duplicated code

AndrewChubatiuk:


Makefile
```diff
@@ -17,7 +17,7 @@ OPERATOR_BIN=operator-sdk
 DOCKER_REPO=victoriametrics/operator
 TARGET_PLATFORM=linux/amd64,linux/arm,linux/arm64,linux/ppc64le,linux/386
 TEST_ARGS=$(GOCMD) test -covermode=atomic -coverprofile=coverage.txt -v
-APIS_BASE_PATH=api/v1beta1
+APIS_BASE_PATH=api/operator/v1beta1
```
Haleygo, line 20: Any reason for the path change?

AndrewChubatiuk: in case of single group setup kubebuilder generates code at `/api/v1beta1`, but client-gen and other generators expect to have group in a path, that's why there was a step in makefile which copies `/api/v1beta1` to `/api/victoriametrics/v1beta1`. in this PR I've converted project to a multi-group (since anyway there's a need to add support for victorialogs manifests), which requires to change path to `/api/{{ group_name }}/{{ version }}`, which is in our case is `/api/operator/v1beta1`, as our manifests are in group `operator`. I would also suggest to change api group to something like `metrics` and corresponding api path to `/api/metrics/v1beta1` to make it more meaningful, but causes breaking changes for users as it changes CRDs as well

f41gh7 (approved):
LGTM

Haleygo (approved):
LGTM

PR #999, opened 2024-06-27
vmagent: use `-enableMultitenantHandlers` command-line flag for vmage…

Haleygo:
…nt remote write multiTenant mode, previous flag `remoteWrite.multitenantURL` is already deprecated.

### DIFF
--- a/internal/controller/operator/factory/vmagent/vmagent.go
+++ b/internal/controller/operator/factory/vmagent/vmagent.go
@@ -1106,9 +1106,12 @@ func buildRemoteWrites(cr *vmv1beta1.VMAgent, ssCache *scrapesSecretsCache) []st
 	remoteTargets := cr.Spec.RemoteWrite
 
 	url := remoteFlag{flagSetting: "-remoteWrite.url=", isNotNull: true}
+	var multiTenantModeSuffix string
 	if cr.Spec.RemoteWriteSettings != nil && cr.Spec.RemoteWriteSettings.UseMultiTenantMode {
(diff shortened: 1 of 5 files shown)


AndrewChubatiuk (dismissed):

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -1106,9 +1106,12 @@ func buildRemoteWrites(cr *vmv1beta1.VMAgent, ssCache *scrapesSecretsCache) []st
 	remoteTargets := cr.Spec.RemoteWrite
 
 	url := remoteFlag{flagSetting: "-remoteWrite.url=", isNotNull: true}
+	var multiTenantModeSuffix string
 	if cr.Spec.RemoteWriteSettings != nil && cr.Spec.RemoteWriteSettings.UseMultiTenantMode {
```
f41gh7: I'd suggest do not modify remoteWrite url path. It brings more harm then good.

 vmagent may accept data with multitenant handlers and forward it further to remote storage without such support. For instance, chain of vmagents or vmagent and prometheus.

 Just mention breaking change at change log and provide solution for configuration migration.

f41gh7: My motivation is:
- useMultiTenantMode  is not a popular option and it's fine to break compatibility for it.
- it makes behavior obvious for people just started to use this option.

Haleygo: I like the chain vmagent idea! Updated, please check.

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1001, opened 2024-06-28
vmalert: add configmap support for `VMAlertSpec.notifierConfigRef`, c…

Haleygo:
…onfig type needs to be specified explicitly

### DIFF
--- a/api/operator/v1beta1/vmalert_types.go
+++ b/api/operator/v1beta1/vmalert_types.go
@@ -186,10 +186,10 @@ type VMAlertSpec struct {
 	// +optional
 	Notifiers []VMAlertNotifierSpec `json:"notifiers,omitempty"`
 
-	// NotifierConfigRef reference for secret with notifier configuration for vmalert
+	// NotifierConfigRef reference for configmap or secret with notifier configuration for vmalert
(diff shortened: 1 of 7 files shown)


api/operator/v1beta1/vmalert_types.go
```diff
@@ -186,10 +186,10 @@ type VMAlertSpec struct {
 	// +optional
 	Notifiers []VMAlertNotifierSpec `json:"notifiers,omitempty"`
 
-	// NotifierConfigRef reference for secret with notifier configuration for vmalert
+	// NotifierConfigRef reference for configmap or secret with notifier configuration for vmalert
 	// only one of notifier options could be chosen: notifierConfigRef or notifiers +  notifier
 	// +optional
-	NotifierConfigRef *v1.SecretKeySelector `json:"notifierConfigRef,omitempty"`
+	NotifierConfigRef *SecretOrConfigMap `json:"notifierConfigRef,omitempty"`
```
f41gh7, line 192: It's a serious breaking change without easy migration path. 

Since notifierConfig is widely used I cannot recommend to do that.

f41gh7: I think it's better to add inlineNotifierConfig, which allows to post insecure configuration directly to `VMAlert` definition.

PR #1004, opened 2024-07-02
docs cleanup

AndrewChubatiuk:


.gitignore
```diff
@@ -1,4 +1,5 @@
 # Temporary Build Files
+.DS_Store
```
f41gh7, line 2: Looks like .DS_Store itself was added to PR. Could you please remove it?

Issue #994, opened 2024-06-26
AsURL always uses HTTP URLs


naveenbharadwaj-nc:
Hi,

In every `*_types.go`, the implementation of `AsURL()` function returns HTTP URL. 
`return fmt.Sprintf("http://%s.%s.svc:%s", cr.PrefixedName(), cr.Namespace, port)`

I'm trying to configure TLS between victoriametrics components which means all components must be accessed via HTTPS. For example, VMAuth must invoke https://vmsingle-.... However, because of hard-coding of `http`, VMAuth cannot proxy requests to other components. I get following error.

> remoteAddr: "10.130.252.154:59080, X-Forwarded-For: 10.230.17.71"; requestURI: /; cannot proxy the request to http://vmsingle-k8s.svc:8429: read tcp 10.130.252.164:33032->172.30.133.124:8429: read: connection reset by peer

Please fix this if its a bug or advise any work arounds.

Haleygo:
Protocol check is included in [v0.46.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.46.0), close as completed.
Feel free to reopen if there are further questions.

naveenbharadwaj-nc:
It seems like there is one point that is missed. To configure TLS for VMAlertManager, we don't specify `tls: true` in `flags`. As per alertmanager documentation, we need to specify `tls_config` in `web.config.file` (see [documentation](https://prometheus.io/docs/alerting/latest/configuration/#tls_config)). 
I can see that no change has been made for [vmalertmanager_types.go](https://github.com/VictoriaMetrics/operator/blob/master/api/operator/v1beta1/vmalertmanager_types.go#L422). It means vmauth will try to proxy the request to HTTP URL of VMAlertmanager. But VMAlertmanager is running on HTTPS. This needs to be addressed.

naveenbharadwaj-nc:
![image](https://github.com/user-attachments/assets/4db417c8-5b35-4219-87db-7f6c1279ceb0)
Logs from vmalertmanager pod that runs on TLS:
`ts=2024-07-16T08:24:46.085Z caller=main.go:181 level=info msg="Starting Alertmanager" version="(version=0.27.0, branch=HEAD, revision=0aa3c2aad14cff039931923ab16b26b7481783b5)"
[...]

PR #1005, opened 2024-07-02
api: adds tls check at flags for https detection

f41gh7:
its possible only at best effort. https certificates must be valid.

https://github.com/VictoriaMetrics/operator/issues/994


f41gh7:
> does TLS enabled affect http probes?

Good point!

f41gh7:
Operator already adjusts probes with `HTTPS` probe handler if needed. https://github.com/VictoriaMetrics/operator/blob/master/internal/controller/operator/factory/build/container.go#L20

AndrewChubatiuk (approved):
LGTM

PR #1028, opened 2024-07-08
removed run as user from kustomize

AndrewChubatiuk:
https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/4752#discussion_r1668891562


tenmozes:
but why? if we can run from non root everywhere i would rather to do that

AndrewChubatiuk:
it's how it used to be before. adding runAsUser set to a static uid breaks an ability to run it on openshift without openshift specific SecurityContextConstraints resource, which is not possible to test locally at the moment

f41gh7 (approved):
LGTM

PR #1032, opened 2024-07-10
enabled webhook in default overlay

AndrewChubatiuk:


config/default/manager_metrics_patch.yaml
```diff
@@ -1,4 +1,7 @@
 # This patch adds the args to allow exposing the metrics endpoint securely
 - op: add
-  path: /spec/template/spec/containers/0/args/0
+  path: /spec/template/spec/containers/0/args/-
   value: --metrics-bind-address=:8080
+- op: add
```
f41gh7: installation will fail, if there is no mounted secrets. 

Also, it's a bit weird, that manager_metrics_patch configures webook for admission controllers.

f41gh7: I'd suggest it either move to webhook related configs somehow or keep it as part of manifests CSV generation commands

AndrewChubatiuk: updated manifests and tested locally on kind with cert-manager installed

config/default/manager_webhook_patch.yaml
```diff
@@ -1,15 +1,22 @@
...
+    containerPort: 9443
+    name: webhook-server
+    protocol: TCP
+- op: add
+  path: /spec/template/spec/containers/0/volumeMounts/-
+  value:
+    mountPath: /tmp/k8s-webhook-server/serving-certs
+    name: cert
+    readOnly: true
+- op: add
```
f41gh7: it ll produce incorrect deployment definition for olm package.

```yaml
                    volumeMounts:
                      - mountPath: /tmp/k8s-webhook-server/serving-certs
                        name: cert
                        readOnly: true
                securityContext:
                  runAsNonRoot: true
                  seccompProfile:
                    type: RuntimeDefault
                serviceAccountName: vm-operator
                terminationGracePeriodSeconds: 10
                volumes:
                  - name: cert
                    secret:
                      defaultMode: 420
                      secretName: webhook-serving-cert
```
I suggest to keep focus on OLM compatibility for kustomize manifests.

 First of all, it must produce valid, reproducible builds for operatorhub.

It's not recommend way for installing operator.

AndrewChubatiuk: you can check other operators in redhat repo, they also have secretName mounted

f41gh7: For readhat it could be fine, since openshift install each operator at own namespace.
 
 But for operatorhub it seems to be not, since it's not possible to share secretName: webhook-serving-cert with other operators at `operators` namespace. Only if it ll be patched with some magic.

AndrewChubatiuk: added a separate overlay `base` for kustomize, `default` is a shared one for `olm` and `base`

f41gh7 (approved):
LGTM

Issue #1036, opened 2024-07-11
VMCLUSTER not uninstalled when WATCH_NAMESPACE is set 


vmw-web:
1. Set WATCH_NAMESPACE to "test-namespace"
2. Installed the operator with the single namespace RBAC in the namespace "test-namespace" [RBAC](https://github.com/VictoriaMetrics/operator/blob/master/config/examples/operator_rbac_for_single_namespace.yaml)
3. Installed the VMCLUSTER in the namespace "test-namespace", and uninstalled, the VMCLUSTER is not uninstalled completely
4. Operator logs shows RECONCILE ERROR

vmw-web:
Updated the operator image to v0.46.4, and the issue still exists. Cluster components are not deleted.

f41gh7:
> Updated the operator image to v0.46.4, and the issue still exists. Cluster components are not deleted.

An issue will be fixed at the next release

f41gh7:
Bug must be fixed at [v0.47.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.47.0)

PR #1039, opened 2024-07-17
internal/finalize: fixes typo at PDB delete function

f41gh7:
Previously name and namespace args were misplaced at removeFinalize function call. It prevented from proper object deletion. And caused various errors during objects deletion.

https://github.com/VictoriaMetrics/operator/issues/1036


AndrewChubatiuk (approved):

Haleygo (approved):
LGTM

Issue #1033, opened 2024-07-10
Missing TLS configuration during self service scrape creation by operator


naveenbharadwaj-nc:
Hello,

> VMServiceScrape object generates part of [VMAgent](https://docs.victoriametrics.com/operator/resources/vmagent/) configuration with [kubernetes service discovery](https://docs.victoriametrics.com/sd_configs.html#kubernetes_sd_configs) targets by corresponding Service. It has various options for scraping configuration of target (with basic auth,tls access, by specific port name etc.).

As per this statement from [documentation](https://docs.victoriametrics.com/operator/resources/vmservicescrape/), we expect that VMServiceScrapes created automatically by the operator must honor `HTTPS scheme` and `TLSConfig` when TLS is configured.

The endpoints of the automatically created VMServiceScrape for all VM components contains only Port and Path. It is from the logic [here](https://github.com/VictoriaMetrics/operator/blob/master/internal/controller/operator/factory/build/vmservicescrape.go#L25-L28). When constructing endpoints, neither scheme nor TLSConfig is populated to VMServiceScrape. This violates the documentation.

Please fix it if its a bug. If my understanding is incorrect and there is some way to tell operator to add TLSConfig and Scheme to generated VMServiceScrapes, please provide such instructions.

f41gh7:
There is a workaround for it. Manually defined needed options at `spec.serviceScrapeSpec` setting. 

For example:

```yaml
kind: VMAlert
spec:
  serviceScrapeSpec:
    endpoints:
    - port: http
      scheme: https 
```

naveenbharadwaj-nc:
Sure. It would also mean that we have to disable `selfServiceSpecCreation` by setting `VM_DISABLESELFSERVICESCRAPECREATION` to `true` (just adding this information for reference). Or will it work if we dont disable?
When victoriametrics-operator has an option to create ServiceScrapes automatically, its obvious that most people would rely on such an option because it is more reliable. Using such work around will force us to add a lot of code (write entire ServiceScrapes on our side for all VM components). Another downside of using work around is that if victoriametrics-operator makes any enhancements/fixes to `selfServiceSpecCreation` creation logic in the future, we would miss it. 
Its not high priority, but definitely useful if implemented on the operator side.

naveenbharadwaj-nc:
What about VMServiceScrape for victoriametrics-operator itself? Operator sends metrics itself and how to tell the operator to send metrics to https endpoint?

[...]

PR #1045, opened 2024-07-24
api/vmalertmanager: adds webserver configuration

f41gh7:
Adds a new type for generic webserver TLS configuration.

* First of all, it allows to properly configure alertmanager and detect tls related params for it. It fixes scheme detection for probes, AsURL function.

* Second, it allows later to rollout this type to other application and have a generic way of TLS/mTLS configuration

* Fixes bug with tlsAssets mount with custom config reloader for alertmanager. Previously it was not mounted correctly and may cause issues with inaccessible assets.

related issues:
- https://github.com/VictoriaMetrics/operator/issues/994
- https://github.com/VictoriaMetrics/operator/issues/1033


api/operator/v1beta1/vmalertmanager_types.go
```diff
@@ -484,8 +490,12 @@ func (cr *VMAlertmanager) ProbePort() string {
 	return cr.Spec.PortName
 }
 
+// ProbeScheme returns scheme for probe
 func (cr *VMAlertmanager) ProbeScheme() string {
-	return strings.ToUpper(protoFromFlags(cr.Spec.ExtraArgs))
+	if cr.Spec.WebConfig != nil && cr.Spec.WebConfig.TLSServerConfig != nil {
```
AndrewChubatiuk, line 495: not sure that this validation is enough. user can set `WebserverTLSConfig.ClientAuthType` to `NoClientCert`

AndrewChubatiuk: is there any validation on alertmanager site?

f41gh7: In this case, tls must be configured. 

 This PR lacks of:
- crd server-side validation
- validation webhook checks
- mention of changes at changelog

It should be addressed before merging.

f41gh7: updated checks, added validation and changelog entry

AndrewChubatiuk (approved):

PR #1046, opened 2024-07-25
relative url in api docs

AndrewChubatiuk:


api/operator/v1beta1/common_scrapeparams.go
```diff
@@ -25,7 +25,7 @@ type VMScrapeParams struct {
 	// disable_keepalive allows disabling HTTP keep-alive when scraping targets.
 	// By default, HTTP keep-alive is enabled, so TCP connections to scrape targets
 	// could be re-used.
-	// See https://docs.victoriametrics.com/vmagent.html#scrape_config-enhancements
+	// See {{% ref "/vmagent.md#scrape_config-enhancements" %}}
```
f41gh7: I'm not sure about special syntax at struct comments.

It also added to the CRD definition as a description and will not work correctly.

AndrewChubatiuk: added sed execution to modify crds

f41gh7 (approved):
LGTM

Issue #1047, opened 2024-07-25
VMAuth skip broken VMUsers during config building


f41gh7:
Currently, a single VMUser with incorrect configuration could break VMAuth config generation.

It must be changed in the same way, as currently handled scrape objects.

It needed because VMUser object could be created by non-owner of VMAuth. 

It'd be great in addition:
- add validation to the webhook configuration (related to the required url related fields)
- add status to the VMUser object
- print summary into operator log

AndrewChubatiuk:
released in [v0.47.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.47.0)

PR #1049, opened 2024-07-29
api/vmuser: adds status.lastSyncError

f41gh7:
Changes behaviour of VMAuth config generation:
* now it skips misconfigured VMUser objects and continues config generation without it
* misconfigured VMUsers marked with error at lastSyncError field
* configuration checks better for misconfiguration (properly deduplicate user tokens, checks static url links and result url_map)

Performs refactoring for CRDRef fetching. Use generic interface and reduce copy-paste.

Introduce new status field for VMUser - `lastSyncError`. Adds column print for it.

https://github.com/VictoriaMetrics/operator/issues/1047


AndrewChubatiuk (dismissed):
LGTM

Issue #1000, opened 2024-06-27
vmagent: allow using files in StreamAggrConfig like RelabelConfig


Haleygo:
vmagent should add support for global `-streamAggr.config` flag and allow using configmap to store the aggregation config like `relabelConfig` does.

PR #1054, opened 2024-08-01
stream aggregation: sync options `dropInputLabels`, `ignoreFirstInter…

Haleygo:
…vals`, `ignoreOldSamples` from upstream, and support using configmap as the source of aggregation rules

address https://github.com/VictoriaMetrics/operator/issues/1000


Haleygo:
related bug https://github.com/VictoriaMetrics/VictoriaMetrics/pull/6744, but it shouldn't block this pull request

AndrewChubatiuk (approved):

Issue #825, opened 2023-12-19
Admission webhook for VMAlertmanagerConfig does not handle all invalid configs


bh-tt:
Applying the following VMAlertmanagerConfig to a cluster with the operator installed and admission webhooks enabled does not result in a failure from the admission webhook but a failure in the operator log. It should result in a validation admission webhook failure.

The failure in this case is a missing `require_tls` setting in the email_configs.

I'm not sure why the admission webhook would use a different set of validation rules than the actual reconciliation, would that make the validation webhook dependent on other resources present in the cluster?

```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAlertmanagerConfig
metadata:
  name: metric-target-down-test
spec:
  route:
    receiver: email
    group_by: ["service","alertname"]
    routes:
      - receiver: email
        continue: false
  receivers:
    - name: email
      email_configs:
        - to: test@example.com
```
```json
{"level":"error","ts":1702980403.351607,"logger":"factory","msg":"bad configs found during alertmanager config building","alertmanager":"alertmanager","error":"errors: incorrect email configuration, tls is required, but no config provided at
 spec in object: it/metric-target-down-test, will ignore vmalertmanagerconfig metric-target-down-test","stacktrace":"github.com/VictoriaMetrics/operator/controllers/factory.buildAlertmanagerConfigWithCRDs\n\tgithub.com/VictoriaMetrics/operat
or/controllers/factory/alertmanager.go:657\ngithub.com/VictoriaMetrics/operator/controllers/factory.createDefaultAMConfig\n\tgithub.com/VictoriaMetrics/operator/controllers/factory/alertmanager.go:547\ngithub.com/VictoriaMetrics/operator/con
trollers/factory.CreateOrUpdateAlertManager\n\tgithub.com/VictoriaMetrics/operator/controllers/factory/alertmanager.go:90\ngithub.com/VictoriaMetrics/operator/controllers.(*VMAlertmanagerReconciler).Reconcile\n\tgithub.com/VictoriaMetrics/op
erator/controllers/vmalertmanager_controller.go:86\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile\n\tsigs.k8s.io/controller-runtime@v0.13.1/pkg/internal/controller/controller.go:121\nsigs.k8s.io/controller-ru
[...]

PR #1055, opened 2024-08-05
api/vmalertmanagerconfig: rework config validation

f41gh7:
* adds more strict validation to VMAlertmanagerConfig
* skip broken configurations from VMAlertmanager during config build
* adds new status fields to the VMalertmanager config:
  * lastSyncError - that defines last error occured to the config during config generation
  * status - generic CRD status for health track
  * additional fields for sync error context, like last error timestamp and name of alertmanager
* adds stack trace only to the panic level errors. It should reduce log noise

https://github.com/VictoriaMetrics/operator/issues/825


AndrewChubatiuk (dismissed):
LGTM

PR #1069, opened 2024-08-15
minor docs fixes

AndrewChubatiuk:
- added webp images conversion
- replaced  tags with markdown
- fixed docs
- replaced platform specific sed in-place replacement with more generic ex


f41gh7 (approved):
LGTM

PR #1082, opened 2024-08-25
alertmanager added configurable port

AndrewChubatiuk:


internal/controller/operator/factory/alertmanager/statefulset.go
```diff
@@ -108,10 +109,13 @@ func CreateOrUpdateAlertManagerService(ctx context.Context, cr *vmv1beta1.VMAler
 	if cr.Spec.PortName == "" {
```
f41gh7, line 109: It also required to change listenAddr arg for statefulset https://github.com/VictoriaMetrics/operator/blob/master/internal/controller/operator/factory/alertmanager/statefulset.go#L177

AndrewChubatiuk: done

Issue #1070, opened 2024-08-15
VMCluster reports app healthy and synced on argocd when the status says otherwise


nonproto:
### Describe the bug

Deploy a VictoriaMetrics Operator and a VMCluster with STS VmSelect.  Make a change to the statefulset declaration that is not allowed and sync.  

Argo shows the app is healthy.  But the VMCluster status shows the
```
status.clusterStatus: failed to create or update vmcluster cannot perform update on sts
```

### To Reproduce

Don't have a config available 

### Version

VMOperator `v0.33.4`
App Version from chart is `v0.46.4`

f41gh7:
Hello, could you please check actual operator logs for the full error text?

nonproto:
here is the sample `vmselect` portion under the `vmcluster`
```
vmselect:
    replicaCount: 1
    revisionHistoryLimitCount: 5
    cacheMountPath: "/select-cache"
    storage:
      volumeClaimTemplate:
        spec:
          storageClassName: my-storage-class
          resources:
            requests:
              storage: 1Gi
```

I deploy the cluster, everything is up and healthy.  I change the revisionHistoryLimitCount from 5 -> 3

ArgoCD shows the VictoriaMetrics App is healthy and synced. 

`vmcluster` is stuck in an expanding status and has the error 
```
Normal   ReconcileEvent      <unknown>  victoria-metrics-operator  starting object update                                                                               
  Warning  ReconcilationError  <unknown>  victoria-metrics-operator  failed create or update vmcluster: cannot perform update on sts: vmselect-victoria-metrics, err: Stat
efulSet.apps "vmselect-victoria-metrics" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persist
entVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

Operators log show

```
[...]

PR #1084, opened 2024-08-26
reconcile: fixes statefulset endless reconcile loop

f41gh7:
* Previous commit made incorrect equal check to revisionHistoryLimit, since kubernetes automatically sets default value for revisionHistory and some other fields. It fails operator check and re-creates statefulset on each reconcile loop.

* with this change operator will set default values to generated statefulset. It must fix this issue and prevent possible future misconfiguration.

related issues
https://github.com/VictoriaMetrics/helm-charts/issues/1319 https://github.com/VictoriaMetrics/operator/issues/1070


AndrewChubatiuk (approved):
LGTM

Issue #1091, opened 2024-08-25
[victoria-metrics-k8s-stack] argocd - OutOfSync (This resource is not present in the application's source. It will be deleted from Kubernetes if the prune option is enabled during sync.)


AurimasNav:
As per picture below, pvc, pod and controllerrevision are being deleted and re-created in an infinite loop.
![image](https://github.com/user-attachments/assets/46ef2b86-2166-405c-8683-480b5350623c)

argocd version: 2.12.2
k3s version: v1.30.4+k3s1

helm via kustomization.yaml
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources: []
namespace: victoriametrics

helmGlobals:
  chartHome: charts
helmCharts:
- name: victoria-metrics-k8s-stack
  version: 0.25.3
  repo: https://victoriametrics.github.io/helm-charts
  releaseName: victoria-metrics-k8s-stack
  includeCRDs: true
  namespace: victoriametrics
  valuesFile: values.yaml
```

application.yaml
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: victoria-metrics-k8s-stack
  namespace: argocd
  finalizers:
  - resources-finalizer.argocd.argoproj.io
spec:
  destination:
    server: https://kubernetes.default.svc
    namespace: victoriametrics
  project: default
  source:
    path: gitops/victoria-metrics-k8s-stack
    repoURL: https://github.com/AurimasNav/k8s-at-home.git
    targetRevision: main
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
    syncOptions:
    - CreateNamespace=true
    - RespectIgnoreDifferences=true
  ignoreDifferences:
    - group: ""
      kind: Secret
      name: victoria-metrics-k8s-stack-victoria-metrics-operator-validation
      namespace: kube-system
      jsonPointers:
        - /data
    - group: admissionregistration.k8s.io
      kind: ValidatingWebhookConfiguration
      name: victoria-metrics-k8s-stack-victoria-metrics-operator-admission
      jqPathExpressions:
      - '.webhooks[]?.clientConfig.caBundle'
```

AndrewChubatiuk:
hey @AurimasNav 
was able to reproduce this issue and transfered it to operator repo, as it's related to how it manages pvc, which is managed by VMSingle

f41gh7:
Fix will be included to the next release.

[...]

PR #1092, opened 2024-08-30
vmsingle, vlogs: set ownerreference on all pvcs

AndrewChubatiuk:
related https://github.com/VictoriaMetrics/operator/issues/1091


api/operator/v1beta1/vmextra_types.go
```diff
@@ -130,20 +131,34 @@ func isContainsFinalizer(src []string) bool {
 
 // RemoveFinalizer - removes vm-operator finalizer from finalizers list.
 // executes provided callback if finalizer found
-func RemoveFinalizer(src client.Object, andThen func(client.Object) error) error {
+func RemoveFinalizer(src client.Object, keepOwnerReference bool, andThen func(client.Object) error) error {
```
f41gh7: It's better to create new function, that will patch remove owner reference. 

* it will reduce amount of changes
* it will not increase code complexity
* we need this function only at 1 place

AndrewChubatiuk: done

f41gh7 (dismissed):
LGTM

f41gh7:
Thanks for contribution!

PR #1109, opened 2024-09-19
controllers/finalizers: respect the enabled status of PDB

Dentrax:
Fixes https://github.com/VictoriaMetrics/VictoriaMetrics/issues/6994

This PR adds a missing check to PDB during `finalize()` phase, and skips if its not enabled. (Tested on our clusters and can confirm it works!)


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

f41gh7:
The reason behind initial behavior of object without `nil` deletion, is possible configuration change during CRD lifetime.

 If `pdb` was added to the spec definition and later removed, operator wasn't able to check that. And `pdb` exists until parent object deletion.

 This inconsistency will be address soon. And will be a part of the upcoming release.

PR #1115, opened 2024-09-25
reverted api docs template change

AndrewChubatiuk:
reverted docs template change as it renders underlying type properly


f41gh7 (approved):
LGTM

PR #1117, opened 2024-09-27
Added support for maxDiskUsagePerURL and forceVMProto in vmagent remoteWriteSpec

leandregagnonlewis:
### Describe Your Changes

Added two fields to `VMAgentRemoteWriteSpec`:
- `forceVMProto`
- `maxDiskUsage`

This allows to easily configure different remote write destination without using the `extraArgs`. 

When only some of  the remote write destinations are set, the others are set to the default value of `"1073741824"` which was the default value set in `VMAgentRemoteWriteSettings`. 

If `maxDiskUsage` is set in both `VMAgentRemoteWriteSettings` and `VMAgentRemoteWriteSpec`, only `VMAgentRemoteWriteSpec` is used. If `extraArgs` is also used, `extraArgs` still takes precedence.

For `forceVMProto`, it is also overriden by `extraArgs` if the flag is set there.


config/base/kustomization.yaml
```diff
@@ -4,3 +4,9 @@ patches:
 - path: manager_webhook_patch.yaml
   target:
     kind: Deployment
+apiVersion: kustomize.config.k8s.io/v1beta1
```
AndrewChubatiuk: @leandregagnonlewis could you please reset changes in this file?

f41gh7 (dismissed):
LGTM.

 Could you please mention changes at docs/changelog.md ?

f41gh7 (dismissed):
LGTM

f41gh7:
Thanks for contribution!

PR #1121, opened 2024-10-02
internal/config: bump vm version to 1.104.0

hagen1778:
Please note, I followed https://docs.victoriametrics.com/release-guide/#operator and it seems incorrect starting on step 2. Please, update the release guide according to the expected procedures for VM version update.


f41gh7 (approved):
LGTM

PR #1123, opened 2024-10-04
vmauth: allow override default config

AndrewChubatiuk:
In a case described [here](https://victoriametrics.slack.com/archives/CGZF1H6L9/p1727776928804079) user utilises vault sidecar for configuration generation, which is not currently supported by operator.
Added ability to disable config reloader if `auth.config` is defined in extraArgs for vmauth spec

fixes https://github.com/VictoriaMetrics/operator/issues/1125


f41gh7:
I think, it's better to add some kind of config source field at vmauth.spec. 

Which should  contain enum of possible sources for unmanaged version of vmauth. 

Because, current change creates some kind of undocumented hack and this behavior could be non-intuitive for users.

 WDYT of  kind.spec.ExternalConfigSource, which current has the following values:
- undefined ( nil value) - configuration managed by operator. Default behavior.
- secretRef - (key,name) for secret with configuration. Replaces deprecated SecretConfig.
- localPath - string for locally mounted secrets.

 Goal of this change - provide clear example, when configuration managed by operator and when it's not. 

cc @Haleygo

f41gh7:
It also, would be great to create an issue with propose change. To keep track of motivation and discussion of proposed changes (slack keeps history only for 3 months atm).

AndrewChubatiuk:
Replaced `configSecret` with `externalConfig`, which includes `secretRef` and `localPath`

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1142, opened 2024-11-01
controller/operator/factory/alertmanager: include `.spec.configMaps` to config-reloader

zekker6:
Properly include ConfigMaps provided via `.spec.configMaps` into a list of watched directories in order to trigger the reload.
Previously, a user-defined `ConfigMap` change would require a manual reload trigger or an additional sidecar.


f41gh7 (approved):
LGTM

Issue #744, opened 2023-09-03
Statefulset OnDelete Limitation and Enhancement


mohammadkhavari:
While using vmcluster, rolling and updating vmstorage and vmselect we can't actually use OnDelete policy, cause when updateStrategy is OnDelete, victoriametrics operator will delete sts pods accordingly, this feature was added due to this issue. #344 
the problem here is when we apply a minor change in vmstorage configurations operator will delete vmstorage pods in order and vminserts will start rerouting, all storage nodes will face pending indexdb items as soon as new timeseries arrive, and this leads to metrics loss in the pipeline, our vmcluster can tolerate one storage node downtime without metric loss, but multiple storage node downtime is troublesome.
Is there any way that we can actually use OnDelete Strategy on vmstorage statefulsets, or even a parameter like stsPodsRollingDelay can be save us from downtime, a time that operator waits between recreating each vmstorage pod.
In fact we can turn off the operator and update statefulset and vmcluster objects and delete pods accordingly, but if we make a mistake in consistency between our manually edited sts and the sts that vmcluster will generate (that cause different revisions), by activating operator pods may recreated again.

Haleygo:
Hello! That's a very good case to consider!
>Is there any way that we can actually use OnDelete Strategy on vmstorage statefulsets

From the [commit](https://github.com/VictoriaMetrics/operator/commit/53b34c6dfe35efbd158e0844afe2b6cdd1d98de6), operator will take over the rolling update process if `rollingUpdateStrategy=OnDelete`, that should cover some corner cases, so I don't think it's recommended to use the real OnDelete Strategy.

>or even a parameter like stsPodsRollingDelay can be save us from downtime, a time that operator waits between recreating each vmstorage pod.

Since both `OnDelete` and `RollingUpdate` wait pod to be ready to delete the next one, I'm thinking you can increase `successThreshold` or `periodSeconds` under vmstorage's readinessProbe, that should prolong the pod ready duration and let vmstorage cluster to be more stable. For example
```
   readinessProbe:
      failureThreshold: 10
      httpGet:
        path: /health
        port: 8482
        scheme: HTTP
      periodSeconds: 10
      successThreshold: 6 // under this setting, vmstorage pod will wait at least 60s to become ready, but it will start working before that 
      timeoutSeconds: 5
```

f41gh7:
I think, it's a global vmcluster issue. Changing rolling update strategy or increasing delays between component updates don't help much in this case.

[...]

PR #1147, opened 2024-11-05
feat: allow dynamic config of terminationGracePeriodSeconds to work with storage.vminsertConnsShutdownDuration flag

rl0nergan:
Relates to https://github.com/VictoriaMetrics/operator/issues/744

Currently with larger vmclusters that may need more time to handle vmstorage rollouts, there's a `storage.vminsertConnsShutdownDuration` flag we can use to handle gracefully closing connections to vminsert while it rolls out. However the issue we're running into is that the operator uses a hard-coded value of 30s to set the grace period for delete, so we increasing the value used for the `storage.vminsertConnsShutdownDuration` flag doesn't help since vmstorage pods will always be killed after 30 seconds.

### DIFF
--- a/internal/controller/operator/factory/reconcile/statefulset.go
+++ b/internal/controller/operator/factory/reconcile/statefulset.go
@@ -20,7 +20,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
-	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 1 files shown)


internal/controller/operator/factory/reconcile/statefulset.go
```diff
@@ -252,7 +252,7 @@ func performRollingUpdateOnSts(ctx context.Context, podMustRecreate bool, rclien
 	for _, pod := range podsForUpdate {
 		l.Info("updating pod", "pod", pod.Name)
 		// we have to delete pod and wait for it readiness
-		err := rclient.Delete(ctx, &pod, &client.DeleteOptions{GracePeriodSeconds: ptr.To(int64(30))})
+		err := rclient.Delete(ctx, &pod, &client.DeleteOptions{GracePeriodSeconds: terminationGracePeriodSeconds})
```
f41gh7: I think it's better to remove `client.DeleteOptions` from rclient.Delete args.
According to the documentation, it's only required to be set if lower period compared to the value defined at pod level is needed.

 It should solve an issue with ignored `terminationGracePeriodSeconds` for all statefulset components.

rl0nergan: sounds good to me. just pushed those changes up. let me know how it looks

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1152, opened 2024-11-11
Fixed deployment readiness (operational status) detection after resource update

Amper:
The algorithm is taken from the `kubectl rollout status` command.

Currently (before this update), the operator sets status operational ahead of time - when vmsingle pods are still restarting.


f41gh7:
Thanks for contribution!

PR #1161, opened 2024-11-18
Added support for discoverRole during servicemonitor conversion

ajaychoudhary-hotstar:
We are using service monitors and relying on VM operator to convert it to vmservicescrape. We are running some services that needs to scale beyond 1000 pods which doesn't work with endpoint objects. We are switching to endpointslices as it is supported. Added annotation based discoveryRole while converting from servicemonitor to vmservicescrape.


ajaychoudhary-hotstar:
@AndrewChubatiuk @Haleygo @f41gh7 Please help in review.

Haleygo:
Thanks for the pull request! 
This approach works but requires additional annotations on each ServiceMonitor during migration. 
I think we can make the default discoveryRole of vmServiceScrape(`endpointslices` or `endpoint`) configurable in vm-operator, allowing users who use `-kubelet-endpointslice=true` in Prometheus Operator to do the same in vm-operator. 
Additionally, users who prefer to always use `endpointslices` instead of `endpoint` for vmservicescrape can set it globally.

ajaychoudhary-hotstar:
Thanks, @Haleygo @f41gh7 I have updated the code as suggested. Thanks.

Haleygo:
> Thanks, @Haleygo @f41gh7 I have updated the code as suggested. Thanks.

It's not just for converting service monitors, but also sets the default role of all the VMServiceScrape(both converted and created), please change this part of code as well.
https://github.com/VictoriaMetrics/operator/blob/8a24bf543016ee687e89f8795a7e3258c4f6fdfa/internal/controller/operator/factory/vmagent/servicescrape.go#L27-L30

internal/config/config.go
```diff
@@ -285,8 +285,9 @@ type BaseOperatorConf struct {
 	FilterChildAnnotationPrefixes []string `default:""`
 	// adds compare-options and sync-options for prometheus objects converted by operator.
 	// It helps to properly use converter with ArgoCD
-	PrometheusConverterAddArgoCDIgnoreAnnotations bool `default:"false"`
-	EnabledPrometheusConverterOwnerReferences     bool `default:"false"`
+	PrometheusConverterAddArgoCDIgnoreAnnotations bool   `default:"false"`
+	EnabledPrometheusConverterOwnerReferences     bool   `default:"false"`
+	VMServiceScrapeDefaultRole                    string `default:"endpoints"`
```
Haleygo: I think var `VMServiceScrapeEnforeEndpointslices` might be easier to validate, wdyt?

ajaychoudhary-hotstar: then it will be hard to reuse the same for other possible values of discoveryRole. Want to keep it like this to reuse for all possible values of discoveryRole.

Haleygo: `VMServiceScrapeEnforeEndpointslices` can only work when `VMServiceScrapeSpec.DiscoveryRole` is not specified.
So maybe rename to `VMServiceScrapeDefaultRoleEndpointslices`:)

ajaychoudhary-hotstar: yeah true, I was just thinking what if someone wants to use "service" as a value instead of "endpoints" or |endpointslices" as it is also possible value in discoveryRole.

ajaychoudhary-hotstar:
> > Thanks, @Haleygo @f41gh7 I have updated the code as suggested. Thanks.
> 
> It's not just for converting service monitors, but also sets the default role of all the VMServiceScrape(both converted and created), please change this part of code as well.
> 
> https://github.com/VictoriaMetrics/operator/blob/8a24bf543016ee687e89f8795a7e3258c4f6fdfa/internal/controller/operator/factory/vmagent/servicescrape.go#L27-L30

hmm, this will be a big change as I need to pass config from multiple places, any quick workaround?

internal/config/config.go
Haleygo: I doubt that, `VMServiceScrapeDefaultRoleEndpointslices` should be enough.

Haleygo:
> > > Thanks, @Haleygo @f41gh7 I have updated the code as suggested. Thanks.
> > 
> > 
> > It's not just for converting service monitors, but also sets the default role of all the VMServiceScrape(both converted and created), please change this part of code as well.
> > https://github.com/VictoriaMetrics/operator/blob/8a24bf543016ee687e89f8795a7e3258c4f6fdfa/internal/controller/operator/factory/vmagent/servicescrape.go#L27-L30
> 
> hmm, this will be a big change as I need to pass config from multiple places, any quick workaround?

You can refer to other default configs under the `BaseOperatorConf`.
https://github.com/VictoriaMetrics/operator/blob/e6e68271b948b58e9a57744dbe382017dbedf56e/internal/controller/operator/factory/build/defaults.go#L19-L32

And call `Default()` for each vmservicescrape when generating scrape_config, like this
https://github.com/VictoriaMetrics/operator/blob/e6e68271b948b58e9a57744dbe382017dbedf56e/internal/controller/operator/vmagent_controller.go#L109

internal/config/config.go
f41gh7: Expected behaviour for `VMEnforceEndpointslices` option:
1) changes default value of discoveryRole from `endpoints` to `endpointslices` during vmagent config build https://github.com/VictoriaMetrics/operator/blob/master/internal/controller/operator/factory/vmagent/servicescrape.go#L28
2) keep behavior the same for explicitly set value of `discoveryRole`


 It solves issue with `endpoints` not only for objects created from `ServiceMonitor`, but also for `VMServiceScrape` components.

ajaychoudhary-hotstar:
@Haleygo @f41gh7 I Have updated the PR, Please review it once.

api/operator/v1beta1/vmextra_types.go
```diff
@@ -1069,6 +1069,11 @@ type CommonDefaultableParams struct {
 	// Has priority over `VM_DISABLESELFSERVICESCRAPECREATION` operator env variable
 	// +optional
 	DisableSelfServiceScrape *bool `json:"disableSelfServiceScrape,omitempty"`
+
+	// Use endpointslices in discoveryrole instead of endpoints
+	// for the vmservicescrape.
+	// +optional
+	VMServiceScrapeDefaultRoleEndpointslices *bool `json:"vmServiceScrapeDefaultRoleEndpointslices,omitempty"`
```
Haleygo, line 1076: `CommonDefaultableParams` is a common struct for multiple CRs like vmcluster..., so we shouldn't add non-common field here.

Haleygo:
> @Haleygo @f41gh7 I Have updated the PR, Please review it once.

@ajaychoudhary-hotstar Thanks for all the efforts! But I'm afraid it's added in the wrong place, please check https://github.com/VictoriaMetrics/operator/pull/1165.

ajaychoudhary-hotstar:
> > @Haleygo @f41gh7 I Have updated the PR, Please review it once.
> 
> @ajaychoudhary-hotstar Thanks for all the efforts! But I'm afraid it's added in the wrong place, please check #1165.

But it is not added in the servicemonitor converter?

ajaychoudhary-hotstar:
> > > @Haleygo @f41gh7 I Have updated the PR, Please review it once.
> > 
> > 
> > @ajaychoudhary-hotstar Thanks for all the efforts! But I'm afraid it's added in the wrong place, please check #1165.
> 
> But it is not added in the service monitor converter?

Do you want me to update this PR as per the shared PR or we will continue with your PR? @Haleygo

Issue #1162, opened 2024-11-18
support for global discoveryRole configuration and per-serviceMonitor discoveryRole


ethfoo:
We are operating large-scale clusters and using `ServiceMonitor` resources (via Prometheus Operator's CRDs). However, we've encountered metric collection issues due to the inability to configure `discoveryRole` as `endpointslice` in `ServiceMonitor` objects. Specifically, the issue stems from the fact that while VictoriaMetrics-Operator automatically converts `ServiceMonitor` resources into `VMServiceScrape `objects, it does not allow for the configuration of `discoveryRole`, which by default is `endpoint`.
Prometheus-Operator has introduced support for configuring the `discoveryRole` directly on Prometheus and PrometheusAgent CRs (as referenced here: [Prometheus Operator API Reference](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.ServiceDiscoveryRole)), but this feature is currently missing in VictoriaMetrics-Operator.

**Proposed Solution:**
1. **Global discoveryRole Configuration**: Add support for a global configuration in VictoriaMetrics-Operator to allow users to specify the discoveryRole (such as endpointslice) by default across all VMServiceScrape resources.
```yaml
env:
  - name: VM_ENABLEDPROMETHEUSCONVERTER_DISCOVERYROLE
    value: "endpointslices"

```
2. **Per-ServiceMonitor discoveryRole Configuration**: Enable the ability to configure the discoveryRole field individually within ServiceMonitor resources. This would provide finer control over discovery methods at the service level.
```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  annotations:
    operator.victoriametrics.com/discovery-role: endpointslices
```

This change would improve the flexibility of metric collection in large-scale clusters and allow for the use of endpointslice discovery where required, reducing missing metric data and improving the overall monitoring setup.

Haleygo:
I'm ok with Solution1, and there's already https://github.com/VictoriaMetrics/operator/pull/1161 for this.
I don't see the need for Solution 2 after Solution1 is implemented, user can use endpointslice to replace endpoints completely.
And jfyi, now it also could be configured by `VMAgent.spec.enforceEndpointSlices: true`, which replaces any usage of endpoints role with endpointslices during config generation.

PR #1165, opened 2024-11-19
add new variable `VM_VMSERVICESCRAPEDEFAULT_ENFORCEENDPOINTSLICES`

Haleygo:
to use `endpointslices` instead of `endpoints` as discovery role for VMServiceScrape when generate scrape config for VMAgent.

address https://github.com/VictoriaMetrics/operator/issues/1162


internal/controller/operator/factory/build/defaults.go
```diff
@@ -582,5 +581,15 @@ func addDefaultsToVMBackup(cr *vmv1beta1.VMBackup, useDefaultResources bool, app
...
 	cr.Resources = Resources(cr.Resources, config.Resource(appDefaults.Resource), useDefaultResources)
+}
 
+func addVMServiceScrapeDefaults(objI interface{}) {
+	cr := objI.(*vmv1beta1.VMServiceScrape)
+	if cr == nil {
+		return
+	}
+	c := getCfg()
+	if cr.Spec.DiscoveryRole == "" && c.VMServiceScrapeDefault.EnforceEndpointslices {
```
ajaychoudhary-hotstar, line 592: Please update service monitor converter as well.

f41gh7 (approved):
LGTM

internal/controller/operator/factory/build/defaults.go
f41gh7: It's not needed: 
1) ServiceMonitor will be converted to the VMServiceScrape
2) Operator selects VMServiceScrap and adds default value for it

 But it'd be great to have some test for it.

ajaychoudhary-hotstar: @f41gh7 @Haleygo This doesn't solve service monitor conversion problem for which I raised my PR. Please update that as well.

Haleygo: @ajaychoudhary-hotstar We don't need to update converter specificly, the workflow is:
```
user create servicemonitor-->vm-operator convert to vmservicescrape--> 
                                                                       vm-operator generate scrape config from vmservicescrape for vmagent
user create vmservicescrape --------------------------------------------> 
```
So we just need to change the default discovery role to `endpointslices` for all the vmservicescrape when generate the scrape config for vmagent, it doesn't matter if the vmservicescrape is created directly or converted.
You can test it using this branch.

ajaychoudhary-hotstar: Thanks for the confirmation. What I noticed in our cluster is it doesn't have discoveryRole: endpoints by default present in any of the vmservicescrape objects. Based on the current implementation it should be added for the converted servicescape objects also right?

Haleygo: >What I noticed in our cluster is it doesn't have discoveryRole: endpoints by default present in any of the vmservicescrape objects.

It's not added to vmservicescrape objects directly, otherwise, vm-operator has to make redundant update for vmservicescrape. It's added in vmagent scrape_config for job generated from vmservicescrape.

>Based on the current implementation it should be added for the converted servicescape objects also right?

yes, it will apply to all the vmservicescrape that without `discoveryRole` specified.

ajaychoudhary-hotstar:
Thanks for helping with this, By when we can expect this change released? :)

internal/controller/operator/factory/build/defaults.go
Haleygo: >But it'd be great to have some test for it.

@f41gh7 yeah, I wanted that but failed, it's more like an integration test.

Haleygo:
> Thanks for helping with this, By when we can expect this change released? :)

I'd say roughly early next month, but that depends on @f41gh7

ajaychoudhary-hotstar:
> > Thanks for helping with this, By when we can expect this change released? :)
> 
> I'd say roughly early next month, but that depends on @f41gh7

It would be great if you could help in releasing this faster. @Haleygo @f41gh7

f41gh7:
Issue was fixed at [v0.50.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.50.0) release

PR #1174, opened 2024-12-02
operator: fix the behaviors of `xxxScrapeConfigSelector` and `xxxNamespaceSelector `

Haleygo:
based on rules [here](https://docs.victoriametrics.com/operator/resources/vmagent/#scraping)

1. fix VMScrapeConfig reconcile: VMScrapeConfig is ignored in it's own reconcile when `vmagentSpec.selectAllByDefault=false`, since `vmagent.IsUnmanaged()` didn't include checks about VMScrapeConfig. But VMScrapeConfig could be pick up in vmagent's reconcile.
2. fix vmscrapeConfigNamespaceSelector in vmagent's reconcile, address https://github.com/VictoriaMetrics/VictoriaMetrics/issues/7687
3. fix object selector when `xxxNamespaceSelector` matches nothing but `selectAllByDefault=true`. Before, it ignores the `xxxNamespaceSelector` that matches nothing.
4. limit the `IsUnmanaged` check scope in reconcile for each scrape CR.


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #891, opened 2024-03-04
bug: two different vmagent's in different namespaces can get the same credentials..


ThomasADavis:
If you:

a) create an vmagent, in namespace a, with a name of  'vmagent', then
b) create a vmnodescrape for cadvisor and verify it works.
c) create a second vmagent, in namespace b, with a name of 'vmagent', then..
d) watch the vmnodscrape for cadvisor fail after the second vmagent is created.  This may take a minute or two to show up.
e) delete the 2nd vmagent, wait for and watch the 1st vmagent cadvisor scrapes start to work again.

when you check the clusterroles/clusterrolbindings, you'll find that the second vmagent has corrupted the first vmagent's credentials.

It will change back and forth each time the operator updates the vmagents.

Workaround is to give every vmagent running on the system a unique name and not to rely on namespacing.

The issue is clusteroles/clusterbindings are not namespaced, and must be unique for each vmagent, and they are not when you do the above.

PR #1176, opened 2024-12-03
vmagent: change the default ClusterRole and ClusterRoleBinding name t…

Haleygo:
…o avoid resource collisions when `vmagentSpec.ServiceAccountName` is null
cluster-scoped resources can't be reused by multiple namespace-scoped objects, since they could have different labels&annotations.

address https://github.com/VictoriaMetrics/operator/issues/891

### DIFF
--- a/internal/controller/operator/factory/vmagent/rbac.go
+++ b/internal/controller/operator/factory/vmagent/rbac.go
@@ -142,8 +142,22 @@ func ensureVMAgentCRExist(ctx context.Context, cr *vmv1beta1.VMAgent, rclient cl
 	clusterRole := buildVMAgentClusterRole(cr)
 	var existsClusterRole rbacv1.ClusterRole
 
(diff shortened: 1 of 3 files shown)


internal/controller/operator/factory/vmagent/rbac.go
```diff
@@ -229,11 +231,13 @@ func buildVMAgentClusterRoleBinding(cr *vmv1beta1.VMAgent) *rbacv1.ClusterRoleBi
 func buildVMAgentClusterRole(cr *vmv1beta1.VMAgent) *rbacv1.ClusterRole {
```
f41gh7, line 259: It'd be great to have migration path for this case. Or add docs for that.

Because after this change, `clusterRoleBinding` and `clusterRole` with previous name will be forever orphaned at kubernetes cluster. 

 Moreover, previous `clusterrole` will still point to the `serviceAccount` and it may confuse users.

Haleygo: okay, update note might not be good, added legacy resources check, please check

f41gh7:
Thanks for contribution!

PR #1199, opened 2024-12-20
fixed olm package

AndrewChubatiuk:
File-based catalog requires two consecutive PR's:
1. adds bundle, which is expected to build validated and built by redhat pipelines
2. updates catalog templates and catalog itself

added manual action to olm packages pipeline. automatic one creates PRs with new bundles and manual updates catalogs for existing bundles


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1202, opened 2024-12-24
fixed env variable name for repo sync auth

AndrewChubatiuk:


.github/workflows/operatorhub.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
         run: |
           gh repo sync ${{ matrix.repo.fork }}
           env:
-            GITHUB_TOKEN: ${{ secrets.VM_BOT_GH_TOKEN }}
+            GH_TOKEN: ${{ secrets.VM_BOT_GH_TOKEN }}
```
Haleygo, line 34: What's the reason for fixing?
According to the [doc](https://cli.github.com/manual/gh_help_environment), both are valid, and we use them both(maybe duplicated) in [chart](https://github.com/VictoriaMetrics/helm-charts/blob/3bc7c0f9bb511230d04c11f3316ee0d4165d4a59/.github/workflows/update-docs.yaml#L96-L97).

AndrewChubatiuk: https://github.com/VictoriaMetrics/operator/actions/runs/12478068097/job/34825099387

Haleygo (approved):

PR #1246, opened 2025-02-25
docs updates

AndrewChubatiuk:
- remove `type` and `required` columns and moved their data to first two columns
- added links to attribute to be able to reference attributes in issues

image


Haleygo:
Anything wrong with the current table?
image

I kinda prefer the `Scheme` and `Required` columns, as it's clear to me.

AndrewChubatiuk:
it allows to decrease width of a column as for some screen sizes it becomes scrollable horizontally, which makes last columns useless. proposed format was inspired by gateway api doc, which looks more compact and readable for me
https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteFilter

f41gh7 (approved):
LGTM

f41gh7:
Thanks!

Issue #996, opened 2024-06-26
vmagent does not recognise  -licence flag


quanghungb:
The VM operators created a sts from a VMagent CRD with following spec (https://docs.victoriametrics.com/operator/api/#vmagentspec)

```yaml 
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAgent
metadata:
  annotations:
    creationTimestamp: "2024-06-26T13:54:26Z"
  finalizers:
  - apps.victoriametrics.com/finalizer
  generation: 3
  labels:
    name: name
  namespace: victoriametrics
  spec:
 
   license:
     key: PM..

```
The sts contains all the information provided by the VMAgent but there **2 lines** of `- license=` in the `args` section for the container `vmagent`

```yaml 
...
 - args:
        - -httpListenAddr=:8429
        - -license=PM..
        - -license=PM..
        - -promscrape.config=/etc/vmagent/config_out/vmagent.env.yaml
...
```
But the most annoying part is the pod is in `CrashLoopBackOff` status  as the `vmagent` container does not start due to this error

```yaml
- containerID: containerd://046d4d85d62373b620655b750cdb0856a01ed974ef9dd76cf9bb6da0a9af1507
    image: docker.io/victoriametrics/vmagent:v1.101.0
    imageID: docker.io/victoriametrics/vmagent@sha256:a17f68fe4024cb93eec23bc5ba52a5c45f645f18e2dc929b7743a239b4212bcb
    lastState:
      terminated:
        containerID: containerd://046d4d85d62373b620655b750cdb0856a01ed974ef9dd76cf9bb6da0a9af1507
        exitCode: 2
        finishedAt: "2024-06-26T16:31:56Z"
        message: |
          flag provided but not defined: -license

          vmagent collects metrics data via popular data ingestion protocols and routes it to VictoriaMetrics.

          See the docs at https://docs.victoriametrics.com/vmagent/ .

          Run "/vmagent-prod -help" in order to see the description for all the available flags
        reason: Error
        startedAt: "2024-06-26T16:31:56Z"
    name: vmagent
```

Ofc without the `license` in the spec of the VMAgent, the vmagent pod starts correctly. 
What could be wrong here?

PR #1254, opened 2025-03-02
Fix vmagent instances setting license args twice

solidDoWant:
This arg is currently set twice for vmagent. Once [here](https://github.com/solidDoWant/operator/blob/b26dcb35ccf227079a5bf0506d722c24d7dc6b41/internal/controller/operator/factory/vmagent/vmagent.go#L610-L611), and once on the on (the now deleted) line 448. This results in it being passed twice in the pod spec.


solidDoWant:
Related issue: https://github.com/VictoriaMetrics/operator/issues/996 (see the second code block)

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #1255, opened 2025-03-04
Introduce new field disableAutoMountServiceAccount


f41gh7:
It's part of security settings. It must change deployment and statefulset field `automountServiceAccountToken:`.

 For components, that requires RBAC access, such as vmagent and vm config-reloader containers. Operator must add corresponding mount paths to volumes and volume mounts.

 Related issue: https://github.com/VictoriaMetrics/helm-charts/issues/2006#issuecomment-2686697509

f41gh7:
The feature was included into [v0.54.0 release](https://docs.victoriametrics.com/operator/changelog/#v0540)

PR #1265, opened 2025-03-09
api: add disableAutomountServiceAccountToken field

f41gh7:
This setting improves application security and simplifies it's
configuration.

  Security rules for cloud provides may forbid implicit serviceAccount
token mounting into the Pod. Such as AKS see the following example: https://github.com/VictoriaMetrics/helm-charts/issues/2006#issuecomment-2686697509

 This commit allows to disable default behavior by changing `Statefulset` and `Deployment` `spec.template.spec.automountServiceAccountToken` setting. It prevents token volume mount by controller-manager.

 In addition, operator explicitly adds Kubernetes API access to the
following objects:
* VMAgent vmagent and config-reloader containers.
* VMAuth config-reloader and init-config containers - if useVMConfigReloader: true is set.
* VMAlertmanager config-reloader and init-config - if useVMConfigReloader: true is set.

 Operator adds volume with name `kube-api-access` and volumeMount with
 path `/var/run/secrets/kubernetes.io/serviceaccount`.

Related issue:
https://github.com/VictoriaMetrics/operator/issues/1255


AndrewChubatiuk (dismissed):
LGTM

docs/security.md
```diff
@@ -133,3 +133,174 @@ spec:
...
+          name: kube-root-ca.crt
+      - downwardAPI:
+          items:
+          - fieldRef:
+              apiVersion: v1
+              fieldPath: metadata.namespace
+            path: namespace
+```
+
+ If `disableAutomountServiceAccountToken: true` is set. Operator adds `volumes` and `volumeMounts` only if application explicitly requires access to Kubernetes API:
```
Haleygo, line 224: >only if application explicitly requires access to Kubernetes API:

Should we clarify more about the applications here only include vmagent and vm config-reloader? Other components like vmcluster require users to mount it manually, can we also provide an example for that?

internal/controller/operator/factory/build/service_account.go
```diff
@@ -29,3 +31,72 @@ func ServiceAccount(cr objectForServiceAccountBuilder) *v1.ServiceAccount {
 		},
 	}
 }
+
+const seriveAccountTokenVolume = "kube-api-access"
```
Haleygo:
```suggestion
const serviceAccountTokenVolume = "kube-api-access"
```

docs/security.md
f41gh7: It's a good point. And it only makes sense if there is some side-car components, that require access to the Kubernetes API.

PR #1284, opened 2025-03-19
Fix no webhook manifest

Koalab99:
Hello,

I think the "without-webhook" never existed or changed name.

Here is the release page for reference: https://github.com/VictoriaMetrics/operator/releases

Thank you.


AndrewChubatiuk:
LGTM, thank you for contribution

AndrewChubatiuk (approved):

PR #1285, opened 2025-03-19
Fix doc VMAuth HA example

Koalab99:
Hello,

I think the example should use replicaCount instead of replicas (https://docs.victoriametrics.com/operator/api/#vmauthspec-replicacount).

Thank you :)


AndrewChubatiuk (approved):
LGTM

PR #1286, opened 2025-03-20
fix and unify namespaceSelector description in VMPodScrape and VMServiceScrape

gpchelkin:
Some info about namespaceSelector being empty was a bit misleading :)


f41gh7:
Thanks for contribution!

PR #1287, opened 2025-03-20
Fix labels merging

dmitriikazanin:
swapped labels maps in merge function


f41gh7 (changes requested):
Can you please add motivation for this change? 

Currently, if any of podMetadata labels will override value of `selectorLabels`, it may completely break operator logic and make it dysfunctional.

dmitriikazanin:
At least it makes sense provide for developers an ability to set suitable value for label "app.kubernetes.io/component" (vmagent, vmauth etc). 
Of course you can provide more suitable solution as I don't know all specific details of the operator.
Anyway it looks odd to set only hardcoded "monitoring" value for all components.

PR #1301, opened 2025-04-05
support settings license for requests load balancer

AndrewChubatiuk:
add ability to set license for cluster load balancer


internal/controller/operator/factory/vmcluster/vmcluster.go
```diff
@@ -1486,6 +1486,10 @@ func buildVMauthLBDeployment(cr *vmv1beta1.VMCluster) (*appsv1.Deployment, error
 	args = cr.Spec.License.MaybeAddToArgs(args, vmv1beta1.SecretsDir)
 
 	args = build.AddExtraArgsOverrideDefaults(args, spec.ExtraArgs, "-")
+
+	volumes, vmounts = spec.License.MaybeAddToVolumes(volumes, vmounts, vmv1beta1.SecretsDir)
```
f41gh7, line 1490: As far as I see, it should be supported already 	https://github.com/VictoriaMetrics/operator/blob/master/internal/controller/operator/factory/vmcluster/vmcluster.go#L1485

Issue #1302, opened 2025-04-07
Full `httpInternalListenAddr` support


Amper:
In v1.111.0 for vmauth was supported flag `httpInternalListenAddr`.
It's possible to set with `extraFalgs` in resource spec, but it doesn't work correctly because operator still use the port from `spec.port` for everything including health checks, config reload url, metrics path for scrape object. 
I propose to create a new spec field `internalPort`, the setting of which will result in:
- setting flag `httpInternalListenAddr` for vmauth deployment
- creating new internal service with this port
- building URLs of service endpoints (`/metrics`, `/health`, `/reload`, etc...) with using of this port in spec of all created resources

FYI: @f41gh7 @j-sokol @jgomezselles

f41gh7:
Feature was included into [v0.56.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.56.0) release

PR #1304, opened 2025-04-08
vmauth: support InternalPort property

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1302


AndrewChubatiuk:
@f41gh7 set InternalPort for all CRs as it looks reasonable not only for VMAuth. do we have any plans to support internal port in other services or is it better to move this property to VMAuth spec only?

internal/controller/operator/factory/vmauth/vmauth.go
```diff
@@ -153,6 +153,9 @@ func makeSpecForVMAuth(cr *vmv1beta1.VMAuth) (*corev1.PodTemplateSpec, error) {
 	}
 
 	args = append(args, fmt.Sprintf("-httpListenAddr=:%s", cr.Spec.Port))
+	if len(cr.Spec.InternalPort) > 0 {
+		args = append(args, "-httpInternalListenAddr=:%s", cr.Spec.InternalPort)
```
f41gh7: This syntax is incorrect. `%s` is format string arg, it requires to use: fmt.Sprintf:

```
args = append(args, fmt.Sprintf("-httpInternalListenAddr=:%s", cr.Spec.InternalPort))
```

internal/controller/operator/factory/vmauth/vmauth.go
```diff
@@ -486,7 +489,7 @@ func buildIngressConfig(cr *vmv1beta1.VMAuth) *networkingv1.Ingress {
 
 func buildVMAuthConfigReloaderContainer(cr *vmv1beta1.VMAuth) corev1.Container {
 	configReloaderArgs := []string{
-		fmt.Sprintf("--reload-url=%s", vmv1beta1.BuildReloadPathWithPort(cr.Spec.ExtraArgs, cr.Spec.Port)),
+		fmt.Sprintf("--reload-url=%s", vmv1beta1.BuildReloadPathWithPort(cr.Spec.ExtraArgs, cr.ProbePort())),
```
f41gh7: It's better to create new variable port, instead of re-using ProbePort method, which could be confusing.

f41gh7:
> @f41gh7 set InternalPort for all CRs as it looks reasonable not only for VMAuth. do we have any plans to support internal port in other services or is it better to move this property to VMAuth spec only?

No, there is no sense to add it to the other components (except vmgateway).

f41gh7:
I addressed review comments at follow-up commit - be4697f4e719df91823f8910111930ae41652d5a

It adds the following changes:
* adds `internal` port to the vmauth container and service
* uses `internal` port for `VMServiceScrape` object
* adds e2e test for new feature
* fixes minor typos
* rename `internalPort` to `internalListenPort`

PR #1312, opened 2025-04-11
vmagent: fixed shard upscale range

AndrewChubatiuk:
iterating shards in a reversed order produces a shard with a negative index

### DIFF
--- a/internal/controller/operator/factory/vmagent/vmagent.go
+++ b/internal/controller/operator/factory/vmagent/vmagent.go
@@ -343,7 +343,7 @@ func createOrUpdateShardedDeploy(ctx context.Context, rclient client.Client, cr,
 func shardNumIter(backward bool, shardCount int) iter.Seq[int] {
 	if backward {
 		return func(yield func(int) bool) {
-			for shardCount >= 0 {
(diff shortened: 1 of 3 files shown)


internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -340,26 +342,6 @@ func createOrUpdateShardedDeploy(ctx context.Context, rclient client.Client, cr,
 	return nil
 }
 
-func shardNumIter(backward bool, shardCount int) iter.Seq[int] {
-	if backward {
-		return func(yield func(int) bool) {
-			for shardCount >= 0 {
```
f41gh7, line 346: I think, this condition should be replaced with `shardCount > 0`

f41gh7: It'd be great to have a test for this function.

AndrewChubatiuk: done

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1313, opened 2025-04-13
fix: use correct TokenId value when building rocketchat config

unki:
When using rocketchat in alertmanager for alert, it appears same secret value gets inserted both into _token_id_ and _token_, although a different secret key is referenced in VmAlertManagerConfig.

```yaml
    ...
    rocketchat_configs:
    - api_url: <url>
      channel: <channel>
      token:
        name: alertmanager
        key: rocketchat_token
      token_id:
        name: alertmanager
        key: rocketchat_token_id
   ....
```

image

I think I located the issue in _config.go_, where the wrong variable is used to retrieve the secret value.


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #1323, opened 2025-04-28
config-reloader: introduce new flag reload-url-auth-key


f41gh7:
VictoriaMetrics compents has an option to protect `/-/reload` endpoints with special `authKey`.
Currently, operator doesn't provide any secure way to config such option with config-reloader container.

It's worth to introduce a new flag, which must be defined as a secret value with a possible on-disk path and env variable definition.

If auth-key is defined, client must switch to `postForm` requests for `/-/reload` API calls. 

auth-key must be refreshed on each reload call.

Current logs:
```
config-reloader 2025-04-28T13:28:08.700Z    info    /go/pkg/mod/github.com/!victoria!metrics/!victoria!metrics@v1.112.0/lib/logger/flag.go:20      -reload-url="http://127.0.0.1:8428/-/reload?authKey=XXXX"
```

Desired logs:

```
config-reloader 2025-04-28T13:28:08.700Z    info    /go/pkg/mod/github.com/!victoria!metrics/!victoria!metrics@v1.112.0/lib/logger/flag.go:20      -reload-url="http://127.0.0.1:8428/-/reload"
config-reloader 2025-04-28T13:28:08.700Z    info    /go/pkg/mod/github.com/!victoria!metrics/!victoria!metrics@v1.112.0/lib/logger/flag.go:20      -reload-url-auth-key="secret"
```

f41gh7:
Feature was included into [v0.57.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.57.0) release

PR #1333, opened 2025-05-05
api: add new field spec.configReloadAuthKeySecret

f41gh7:
This commit introduces secure way of managing reloadAuthKey
configuration for VMAgent, VMAuth and VMAlert applications.

 Previously, it was only possible to define it via `extraArgs` and set
authKey as a plain text.

 With this commit, operator uses secret reference and mounts provided
secret into application and config-reloader containers. It allows to perform secret rotation without application restarts.

 In order to mask auth key, new flag was added to the config-reloader
container - `reload-url-auth-key`. It's generic password flag from flagutil lib. It dynamically reads secret value with each reload request.

Related issue:
https://github.com/VictoriaMetrics/operator/issues/1323


internal/controller/operator/factory/build/container.go
```diff
@@ -329,3 +330,54 @@ func AddsPortProbesToConfigReloaderContainer(useVMConfigReloader bool, crContain
...
+	})
+	return dst
+}
+
+// AddConfigReloadAuthKeyToApp adds authKey env var to the given application container
+func AddConfigReloadAuthKeyToApp(container *corev1.Container, extraArgs map[string]string, spec *vmv1beta1.CommonConfigReloaderParams) {
+	if spec.ConfigReloadAuthKeySecret == nil {
+		return
+	}
+	container.Args = append(container.Args, fmt.Sprintf("-%s=file://%s/%s", authKeyAppFlagName, authKeyMountPath, spec.ConfigReloadAuthKeySecret.Key))
```
f41gh7, line 362: reloadAuthKey could be already set by extraArgs or extraEnvs. I think, it's safe to ignore those values and it's users responsibility to properly configure it. 

 But we could validation to the webhook.

PR #1352, opened 2025-05-10
docs/troubleshooting: Unexpected vmstorage cache invalidation or inconsistent cache max size

makasim:
document https://github.com/VictoriaMetrics/VictoriaMetrics/issues/8883 case in troubleshooting guide.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,0 +1,34 @@
+---
+weight: 15
+title: Troubleshooting
+menu:
+  docs:
+    parent: "operator"
+    weight: 13
+---
+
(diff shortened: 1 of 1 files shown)


hagen1778 (commented):
It looks to me that the main recommendations should be requests == limits for stateful components. If it is applied, other problems won't appear. And the cache thing wasn't a problem or trouble, tbh

makasim:
It does not work the way it is described in PR. The code always use limit https://github.com/VictoriaMetrics/VictoriaMetrics/blob/19d3c443912e737c957b2efbffaf5c8cb5c4b677/lib/memory/memory_linux.go#L21

PR #1359, opened 2025-05-14
docs/quick-start: Fix legacy links

makasim:

PR #1360, opened 2025-05-15
docs/quick-start: Fix legacy links

makasim:
follow up https://github.com/VictoriaMetrics/operator/pull/1359


makasim:
closing as the problem was in how anchor links behave: https://github.com/VictoriaMetrics/VictoriaMetrics/issues/8943

Issue #1337, opened 2025-05-06
Tag artefact operator should include similar to helm chart labels


makasim:
It would be beneficial to add the following standard labels to the Operator bundled with releases (for example [v0.56.0 release](https://github.com/VictoriaMetrics/operator/releases/tag/v0.56.0)) to maintain consistency with the Helm chart. This would ensure the provided examples (https://github.com/VictoriaMetrics/operator/pull/1328) work seamlessly across both the artifact and the Helm-based installation:

Add llabels:
```
app.kubernetes.io/instance: vmoperator
app.kubernetes.io/name: victoria-metrics-operator
```

Currently, it has only:
```
control-plane: vm-operator
```

PR #1364, opened 2025-05-16
config: sync installer operator labels with Helm-installed version

makasim:
The goal of this update is to ensure that the examples in the Operator quick start guide are also applicable when the Operator is installed via Helm charts.

Fixes https://github.com/VictoriaMetrics/operator/issues/1337

Tested in Kind with https://github.com/VictoriaMetrics/debug-notes/tree/main/operator/quick-start,

installed v0.58.0, than this version, than again v0.58.0. worked

### DIFF
--- a/config/default/metrics_service.yaml
+++ b/config/default/metrics_service.yaml
@@ -2,8 +2,11 @@ apiVersion: v1
 kind: Service
 metadata:
   labels:
+    # deprecated, could not be removed because of deployment match selector immutability
     control-plane: vm-operator
-    app.kubernetes.io/name: vm-operator
+
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/name: victoria-metrics-operator
     app.kubernetes.io/managed-by: kustomize
   name: operator-metrics-service
   namespace: vm
--- a/Makefile
+++ b/Makefile
@@ -219,10 +219,11 @@ publish:
 
 .PHONY: build-installer
 build-installer: manifests generate kustomize ## Generate a consolidated YAML with CRDs and deployment.
-	mkdir -p dist
+	mkdir -p dist && rm -rf dist/*
 	cd config/manager && $(KUSTOMIZE) edit set image manager=$(REGISTRY)/$(ORG)/$(REPO):$(TAG)
 	$(KUSTOMIZE) build config/base > dist/install-no-webhook.yaml
 	$(KUSTOMIZE) build config/base-with-webhook > dist/install-with-webhook.yaml
+	$(KUSTOMIZE) build config/crd/overlay > dist/crd.yaml
 
 olm: operator-sdk opm yq docs
 	rm -rf bundle*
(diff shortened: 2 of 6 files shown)


config/base/kustomization.yaml
```diff
@@ -1,4 +1,4 @@
 namespace: vm
-namePrefix: vm-
```
makasim: because it looks better:

```
$kubectl -n vm get pod 
NAME                             READY   STATUS              RESTARTS   AGE
vmagent-demo-5855fddf5d-wj54g    2/2     Running             0          14s
vmalert-demo-87c78595d-nqmgb     0/2     ContainerCreating   0          4s
vmalertmanager-demo-0            0/2     ContainerCreating   0          6s
vmoperator-585f4d9b7-fsft4       1/1     Running             0          34s
vmsingle-demo-744b9c44cb-5b567   1/1     Running             0          21s
```

makasim:
Its not backward compatible

config/default/metrics_service.yaml
```diff
@@ -2,8 +2,11 @@ apiVersion: v1
 kind: Service
 metadata:
   labels:
+    # deprecated, could not be removed because of deployment match selector immutability
     control-plane: vm-operator
-    app.kubernetes.io/name: vm-operator
+
+    app.kubernetes.io/instance: vmoperator
```
AndrewChubatiuk: `app.kubernetes.io/instance` - A unique name identifying the instance of an application
This labels is not relevant at all

makasim: Why helm sets this label then ?

AndrewChubatiuk: in helm value of this label is taken from release name, it means that you can deploy multiple instances of the same chart with different configurations, etc. This label is required to avoid service, scrapes and other selectors intersection, while multiple instances of the same chart are deployed

makasim: > in helm value of this label is taken from release name, it means that you can deploy multiple instances of the same chart with different configurations, etc. This label is required to avoid service, scrapes and other selectors intersection, while multiple instances of the same chart are deployed

The logic you've described could be applied to the operator installed via a tagged manifest. If you just download and apply it, it would be the default release.  Then you can run it via kustomize (change the instance label, + other stuff)  and deploy a new  operator release.

This way, the operators installed via manifest have a release label too, and are also consistent with the operator installed via helm. 

Also the label is not helm specific, it is one of the [well-known k8s labels](https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-instance)

I am not sure about the release name `vmoperator`. Maybe `default` is a better one. WDYT?

AndrewChubatiuk: it's not helm specific, but meaningless in context of plain manifests. set it to default

makasim: @AndrewChubatiuk fixed, please review

PR #1366, opened 2025-05-16
docs/changelog: Add update note to v0.58.0 release

makasim:
Related to https://github.com/VictoriaMetrics/VictoriaMetrics/pull/8941, https://github.com/VictoriaMetrics/VictoriaMetrics/pull/8945

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -17,6 +17,10 @@ aliases:
 
 **Release date:** 14 May 2025
 
+**Update Note 1:** This release by default deploys`vmagent` which contains a bug, see more details in [VictoriaMetrics#8941](https://github.com/VictoriaMetrics/VictoriaMetrics/pull/8941).
+We recommend skipping this release and waiting for newer release. 
+If you still want to upgrade, you can override the vmagent image version by setting the environment variable: `VM_VMAGENTDEFAULT_VERSION=v1.117.1`
+
 * Dependency: [vmoperator](https://docs.victoriametrics.com/operator/): Updated default versions for VM apps to [v1.117.0](https://github.com/VictoriaMetrics/VictoriaMetrics/releases/tag/v1.117.0) version
 
 * FEATURE: [operator](https://docs.victoriametrics.com/operator): introduce new flags for leader election configuration - `leader-elect-namespace`, `leader-elect-id`.


docs/CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ aliases:
 
 **Release date:** 14 May 2025
 
+**Update Note 1:** This release introduces a bug in the `vmagent`. The relabeling config is not properly reloaded, causing the relabeling logic to be skipped.
+We advise skipping this version and waiting for the fix in v0.58.1. See more details in [VictoriaMetrics#8941](https://github.com/VictoriaMetrics/VictoriaMetrics/pull/8941).
```
```suggestion
**Update Note 1:** This release by default deploys`vmagent` which contains a bug, see more details in [VictoriaMetrics#8941](https://github.com/VictoriaMetrics/VictoriaMetrics/pull/8941). 
We advise overriding the default vmagent image version by adding ENV `VM_VMAGENTDEFAULT_VERSION=v1.117.1`, or skipping this version and waiting for the fix in v0.58.1.
```

AndrewChubatiuk: agree with Haley's suggestion

makasim: Let me disagree. While this is a possible solution, we should avoid recommending hardcoding the version. It’s easy to forget to remove it later (People, in general, are not good at remembering stuff). It will lead to outdated versions and break automatic updates. It also introduces unnecessary manual effort for operators, especially at scale.

Since the latest release isn’t special and we ship updates regularly, there’s a high chance it could be skipped. So the simple recommendation to avoid the version is enough IMO.

WDYT @hagen1778  ?

AndrewChubatiuk: I'm not sure it's VM case. Lot's of VM operator users set VM version explicitly to not depend on operator default version changes

makasim: We can mention it without explicitly recommending it — would that work for you?

AndrewChubatiuk: Okay, but I'm strongly against this sentence `We advise skipping this version and waiting for the fix in v0.58.1`

Haleygo: >While this is a possible solution, we should avoid recommending hardcoding the version.

I think it depends on the users, to add the ENV and take care of it.
I'm ok with recommending to skip this vm-operator version as the first option, but it would be helpful to mention the workaround of overriding with the ENV for those wanting the new version. Since the issue stems from the VM component, informing users about the workaround on the VM-operator side seems reasonable.

AndrewChubatiuk (approved):
LGTM

Issue #1338, opened 2025-05-06
Release artefacts should provide operator only manifests.


makasim:
Currently, there are three artifact files (for example [v0.56.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.56.0)):

- crd.yaml – contains only the CRDs
- install-no-webhook.yaml – includes CRDs and operator resources
- install-with-webhook.yaml – includes CRDs, operator resources, and webhook-related resources

Each file is around 3,000 lines long, primarily due to the size of the CRDs.

It would be helpful to introduce a separate operator.yaml file that includes only the operator-related manifests. This would make it easier to review and modify operator resources without wading through large CRD definitions. It would also simplify studying and understanding the operator setup.

makasim:
It could help to improve examples in quick start (https://github.com/VictoriaMetrics/operator/pull/1328).

makasim:
Splitting resources and the operator is not practical, it adds more ways to break stuff during the installation. Closing as won't implement.

PR #1369, opened 2025-05-16
build operator without crds

makasim:
Fixes https://github.com/VictoriaMetrics/operator/issues/1338


makasim:
Splitting resources and the operator is not practical, it adds more ways to break stuff during the installation. Closing as won't implement.

PR #1374, opened 2025-05-21
docs/quick-start: Update next steps section

makasim:
This is the last one related to quick start page.


docs/quick-start.md
```diff
@@ -733,32 +733,31 @@ Make sure to use the same Host header you defined in your `VMAuth` config, or th
 
 ## Next steps
 
-That's it. We obtained a monitoring cluster corresponding to the target topology:
+Now that you have a working setup with [storage](https://docs.victoriametrics.com/operator/quick-start/#storage), [scraping](https://docs.victoriametrics.com/operator/quick-start/#scraping), [alerting](https://docs.victoriametrics.com/operator/quick-start/#alerting), and secure [access](https://docs.victoriametrics.com/operator/quick-start/#access),
+and a solid understanding of how everything fits together,
+you’re ready to take the next steps with VictoriaMetrics on Kubernetes.
 
-![Cluster Scheme](quick-start_cluster-scheme.webp)
```
hagen1778, line 738: I don't see this image to be re-used in the change. Shall we drop it from version control too?

docs/quick-start.md
```diff
@@ -733,32 +733,31 @@ Make sure to use the same Host header you defined in your `VMAuth` config, or th
...
 
-You have a full-stack monitoring cluster with VictoriaMetrics Operator.
+If you’re moving from Prometheus, VictoriaMetrics [makes the migration easy](https://docs.victoriametrics.com/operator/migration/).
+You can keep your current scrape configs and alert rules, and move your data with just a few changes.
 
-You can find information about these and other resources of operator on the [Custom resources page](https://docs.victoriametrics.com/operator/resources/).
+As you continue to explore, you might find the [resources](https://docs.victoriametrics.com/operator/resources/) and [their configuration reference](https://docs.victoriametrics.com/operator/api/) pages helpful.
 
-In addition, check out other sections of the documentation for VictoriaMetrics Operator:
+If you're expecting more traffic or need better uptime, consider switching from [VMSingle](https://docs.victoriametrics.com/operator/resources/vmsingle/) to [VMCluster](https://docs.victoriametrics.com/operator/resources/vmcluster/).
```
hagen1778: I'd make a last attempt to agitate for using single-node.

```
While [VMSingle](https://docs.victoriametrics.com/operator/resources/vmsingle/) has great vertical scalability capabilities and efficiency, [VMCluster](https://docs.victoriametrics.com/operator/resources/vmcluster/) allows scaling beyond capabilities of a single physical instance.
```

hagen1778 (dismissed):

docs/quick-start.md
makasim: I’ve decided to keep them, as they might be linked from external sources beyond our control. Do you think it's not worse it ?

docs/quick-start.md
makasim: Agree, I rewrote the text to stress that vmsingle is the recommended way.

PTAL

docs/quick-start.md
hagen1778: If our docs don't contain links on them - I'd kill the image.
It is unlikely someone store the link to image instead of copy of the image.

docs/quick-start.md
hagen1778: Thanks! It is much better than what I suggested!

hagen1778 (approved):

docs/quick-start.md
makasim: Removed!

Issue #1314, opened 2025-04-15
Support VictoraLogs Cluster mode


yuriydzobak:
Hi,
VictoriaLogs supports cluster mode starting from version 1.18.0. You can find more information here: https://docs.victoriametrics.com/victorialogs/cluster/.
It would be great to implement a new CRD, VLCluster, similar to VMCluster.
Thanks.

f41gh7:
It's on our road map. Currently, we're working on HA version for VLogsCluster.

valyala:
> It's on our road map. Currently, we're working on HA version for VLogsCluster.

It is better to make a separate simple CRD for VictoriaLogs cluster, which will allow to setup and run a regular VictoriaLogs cluster in Kubernetes with simple easy to understand configs. This CRD could be used by the majority of VictoriaLogs users who need horizontal scalability in a Kubernetes cluster with Kubernetes nodes running in a single AZ.

Later we could add an advanced CRD for multi-AZ HA VictoriaLogs setup. This must be a separate CRD in order to keep the simple CRD for VictoriaLogs cluster simple.

f41gh7:
Feature was included into [v0.59.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.59.0) and follow-up [v0.59.1](https://github.com/VictoriaMetrics/operator/releases/tag/v0.59.1) releases.

PR #1380, opened 2025-05-25
api/v1: introduce VLCluster component

f41gh7:
Fixes https://github.com/VictoriaMetrics/operator/issues/1314


api/operator/v1/vlcluster_types.go
```diff
@@ -0,0 +1,817 @@
...
+
+// GetAdditionalService returns AdditionalServiceSpec settings
+func (cr *VLInsert) GetAdditionalService() *vmv1beta1.AdditionalServiceSpec {
+	return cr.ServiceSpec
+}
+
+// VLInsertSyslogSpec defines syslog configuration for vlinsert
+type VLInsertSyslogSpec struct {
+	// ListenPort defines syslog server listen port
+	ListenPort int32 `json:"listenPort"`
```
Haleygo, line 206: victorialogs can accept syslog messages via multiple TCP and UDP ports with individual configurations, see https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#multiple-configs.
Also, since victorialogs separates the flags of TCP and UDP and allows setting them differently, should operator follow that?

api/operator/v1/vlcluster_types.go
```diff
@@ -0,0 +1,817 @@
...
+	*vmv1beta1.EmbeddedProbes `json:",inline"`
+	// Configures horizontal pod autoscaling.
+	// +optional
+	HPA *vmv1beta1.EmbeddedHPA `json:"hpa,omitempty"`
+	// RollingUpdateStrategy defines strategy for application updates
+	// Default is OnDelete, in this case operator handles update process
+	// Can be changed for RollingUpdate
+	// +optional
+	RollingUpdateStrategy appsv1.StatefulSetUpdateStrategyType `json:"rollingUpdateStrategy,omitempty"`
+	// ClaimTemplates allows adding additional VolumeClaimTemplates for StatefulSet
```
Haleygo: Why is vlselect created as a StatefulSet by default? Since it doesn't support result cache, is that in the roadmap?

api/operator/v1/vlcluster_types.go
```diff
@@ -0,0 +1,817 @@
...
+	// Log entries with timestamps bigger than now+futureRetention are rejected during data ingestion; see https://docs.victoriametrics.com/victorialogs/#retention
+	// +optional
+	// +kubebuilder:validation:Pattern:="^[0-9]+(h|d|y)?$"
+	FutureRetention string `json:"futureRetention,omitempty"`
+	// LogNewStreams Whether to log creation of new streams; this can be useful for debugging of high cardinality issues with log streams; see https://docs.victoriametrics.com/victorialogs/keyconcepts/#stream-fields
+	LogNewStreams bool `json:"logNewStreams,omitempty"`
+	// Whether to log all the ingested log entries; this can be useful for debugging of data ingestion; see https://docs.victoriametrics.com/victorialogs/data-ingestion/
+	LogIngestedRows bool `json:"logIngestedRows,omitempty"`
+
+	// PodMetadata configures Labels and Annotations which are propagated to the VLSelect pods.
```
Haleygo: Please clean the `VLSelect` under VLStorage.

api/operator/v1/vlcluster_types.go
```diff
@@ -0,0 +1,817 @@
...
+	// +optional
+	RollingUpdateStrategy appsv1.StatefulSetUpdateStrategyType `json:"rollingUpdateStrategy,omitempty"`
+	// ClaimTemplates allows adding additional VolumeClaimTemplates for StatefulSet
+	ClaimTemplates []corev1.PersistentVolumeClaim `json:"claimTemplates,omitempty"`
+
+	// StorageDataPath - path to storage data
+	// +optional
+	StorageDataPath string `json:"storageDataPath,omitempty"`
+	// Storage - add persistent volume for StorageDataPath
+	// its useful for persistent cache
```
Haleygo:
```suggestion
```

api/operator/v1/vlcluster_types.go
AndrewChubatiuk: also can it be reused in VLSingle?

f41gh7: > Also, since victorialogs separates the flags of TCP and UDP and allows setting them differently, should operator follow that?

 It's a good point. I think, it's worth to have different configuration of TCP and UDP.

internal/config/config.go
```diff
@@ -340,6 +340,55 @@ type BaseOperatorConf struct {
...
+			Port     string `default:"9428"`
+			Resource struct {
+				Limit struct {
+					Mem string `default:"1500Mi"`
+					Cpu string `default:"1000m"`
+				} `prefix:"LIMIT_"`
+				Request struct {
+					Mem string `default:"500Mi"`
+					Cpu string `default:"250m"`
+				} `prefix:"REQUEST_"`
```
Haleygo: I think it's better to set the default limit and request values the same, as we recommend to have guaranteed QoS for the storage node.

internal/config/config.go
```diff
@@ -340,6 +340,55 @@ type BaseOperatorConf struct {
 		ConfigReloaderMemory string `default:"25Mi" env:"CONFIGRELOADERMEMORY"`
 	} `prefix:"VMAUTHDEFAULT_"`
 
+	VLClusterDefault struct {
+		UseDefaultResources bool `default:"true" env:"USEDEFAULTRESOURCES"`
+		VLSelectDefault     struct {
+			Image    string `default:"victoriametrics/victoria-logs"`
+			Version  string `env:",expand" default:"${VM_LOGS_VERSION}-victorialogs"`
+			Port     string `default:"9428"`
```
Haleygo: I think we should assign different default ports for vlselect&vlinsert pods and services, to avoid possible port conflicts(hostnetwork mode...), and make those services more recognizable.

internal/controller/operator/factory/vlcluster/vlcluster.go
```diff
@@ -0,0 +1,254 @@
...
+				}
+			}
+			if ptr.Deref(vmst.DisableSelfServiceScrape, false) && !ptr.Deref(prevSt.DisableSelfServiceScrape, false) {
+				if err := finalize.SafeDeleteWithFinalizer(ctx, rclient, &vmv1beta1.VMServiceScrape{ObjectMeta: commonObjMeta}); err != nil {
+					return fmt.Errorf("cannot remove serviceScrape from prev storage: %w", err)
+				}
+			}
+			prevSvc, currSvc := prevSt.ServiceSpec, vmst.ServiceSpec
+			if err := reconcile.AdditionalServices(ctx, rclient, cr.GetVLStorageName(), cr.Namespace, prevSvc, currSvc); err != nil {
+				return fmt.Errorf("cannot remove vmstorage additional service: %w", err)
```
Haleygo:
```suggestion
				return fmt.Errorf("cannot remove vlstorage additional service: %w", err)
```

internal/controller/operator/factory/vlcluster/vlcluster.go
```diff
@@ -0,0 +1,254 @@
...
+				}
+			}
+			if ptr.Deref(vmse.DisableSelfServiceScrape, false) && !ptr.Deref(prevSe.DisableSelfServiceScrape, false) {
+				if err := finalize.SafeDeleteWithFinalizer(ctx, rclient, &vmv1beta1.VMServiceScrape{ObjectMeta: commonObjMeta}); err != nil {
+					return fmt.Errorf("cannot remove serviceScrape from prev select: %w", err)
+				}
+			}
+			prevSvc, currSvc := prevSe.ServiceSpec, vmse.ServiceSpec
+			if err := reconcile.AdditionalServices(ctx, rclient, cr.GetVLSelectName(), cr.Namespace, prevSvc, currSvc); err != nil {
+				return fmt.Errorf("cannot remove vmselect additional service: %w", err)
```
Haleygo:
```suggestion
				return fmt.Errorf("cannot remove vlselect additional service: %w", err)
```

internal/controller/operator/factory/vlcluster/vlcluster.go
```diff
@@ -0,0 +1,254 @@
...
+				}
+			}
+			if ptr.Deref(vmis.DisableSelfServiceScrape, false) && !ptr.Deref(prevIs.DisableSelfServiceScrape, false) {
+				if err := finalize.SafeDeleteWithFinalizer(ctx, rclient, &vmv1beta1.VMServiceScrape{ObjectMeta: commonObjMeta}); err != nil {
+					return fmt.Errorf("cannot remove serviceScrape from prev insert: %w", err)
+				}
+			}
+			prevSvc, currSvc := prevIs.ServiceSpec, vmis.ServiceSpec
+			if err := reconcile.AdditionalServices(ctx, rclient, cr.GetVLInsertName(), cr.Namespace, prevSvc, currSvc); err != nil {
+				return fmt.Errorf("cannot remove insert additional service: %w", err)
```
Haleygo, line 191:
```suggestion
				return fmt.Errorf("cannot remove vlinsert additional service: %w", err)
```

api/operator/v1/vlcluster_types.go
```diff
@@ -0,0 +1,817 @@
...
+func (cr *VLSelect) GetAdditionalService() *vmv1beta1.AdditionalServiceSpec {
+	return cr.ServiceSpec
+}
+
+// VLStorage defines vlstorage component configuration at victoria-logs cluster
+type VLStorage struct {
+	// RetentionPeriod for the stored logs
+	// https://docs.victoriametrics.com/victorialogs/#retention
+	// +optional
+	// +kubebuilder:validation:Pattern:="^[0-9]+(h|d|y)?$"
```
Haleygo: RetentionPeriod should also support `s (second), w (week)`, so does FutureRetention. Seems like we have the same incorrect limitations for VM.
>  -retentionPeriod value
    	Log entries with timestamps older than now-retentionPeriod are automatically deleted; log entries with timestamps outside the retention are also rejected during data ingestion; the minimum supported retention is 1d (one day); see https://docs.victoriametrics.com/victorialogs/#retention ; see also -retention.maxDiskSpaceUsageBytes
    	The following optional suffixes are supported: s (second), h (hour), d (day), w (week), y (year). If suffix isn't set, then the duration is counted in months (default 7d)

internal/controller/operator/factory/vlcluster/vlstorage.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	build.StatefulSetAddCommonParams(stsSpec, ptr.Deref(cr.Spec.VLStorage.UseStrictSecurity, false), &cr.Spec.VLStorage.CommonApplicationDeploymentParams)
+	storageSpec := cr.Spec.VLStorage.Storage
+	storageSpec.IntoSTSVolume(cr.Spec.VLStorage.GetStorageVolumeName(), &stsSpec.Spec)
+	stsSpec.Spec.VolumeClaimTemplates = append(stsSpec.Spec.VolumeClaimTemplates, cr.Spec.VLStorage.ClaimTemplates...)
+
+	return stsSpec, nil
+}
+
+func buildVLStoragePodSpec(cr *vmv1.VLCluster) (*corev1.PodTemplateSpec, error) {
```
Haleygo, line 166: Missing check for fields `RetentionMaxDiskSpaceUsageBytes`, `FutureRetention`, `LogNewStreams` and `LogIngestedRows`.

internal/config/config.go
f41gh7: It's better to keep ports in sync with our documentation and default values at `victoria-logs`. It uses `9428`, so lets keep it.

api/operator/v1/vlcluster_types.go
f41gh7: I doubt that `s` as `second` make any sense in case of retention. But it's worth to add `w` as `week`.

internal/config/config.go
f41gh7: This resource distribution allows to run components with minimal resource requirements.  And we recommend to have a CPU limit equal to the whole CORE. So I think it's fine.

api/operator/v1/vlcluster_types.go
```diff
@@ -0,0 +1,889 @@
...
+	StreamFields FieldsListString `json:"streamFields,omitempty"`
+	// IgnoreFields to ignore at logs
+	// +optional
+	IgnoreFields FieldsListString `json:"ingoreFields,omitempty"`
+	// DecolorizeFields to remove ANSI color codes across logs
+	// +optional
+	DecolorizeFields FieldsListString `json:"decolorizeFields,omitempty"`
+	// TenantID for logs ingested
+	// +optional
+	TenantID string `json:"tenantID,omitempty"`
```
Haleygo:
```suggestion
	// StreamFields to use as log stream labels, see https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#stream-fields
	// +optional
	StreamFields FieldsListString `json:"streamFields,omitempty"`
	// IgnoreFields to ignore at logs, see https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#dropping-fields
	// +optional
	IgnoreFields FieldsListString `json:"ignoreFields,omitempty"`
	// DecolorizeFields to remove ANSI color codes across logs, see https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#decolorizing-fields
	// +optional
	DecolorizeFields FieldsListString `json:"decolorizeFields,omitempty"`
	// TenantID for logs ingested, see https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#multiple-configs
	// +optional
	TenantID FieldsListString `json:"tenantID,omitempty"`
```

api/operator/v1/vlcluster_types.go
```diff
@@ -0,0 +1,889 @@
...
```
Haleygo:
```suggestion
	// StreamFields to use as log stream labels, see https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#stream-fields
	// +optional
	StreamFields FieldsListString `json:"streamFields,omitempty"`
	// IgnoreFields to ignore at logs, see https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#dropping-fields
	// +optional
	IgnoreFields FieldsListString `json:"ignoreFields,omitempty"`
	// DecolorizeFields to remove ANSI color codes across logs, see https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#decolorizing-fields
	// +optional
	DecolorizeFields FieldsListString `json:"decolorizeFields,omitempty"`
	// TenantID for logs ingested, see https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#multiple-configs
	// +optional
	TenantID FieldsListString `json:"tenantID,omitempty"`
```

api/operator/v1/vlcluster_types.go
```diff
@@ -0,0 +1,889 @@
...
+	UDPListeners []*SyslogUDPListener `json:"udpListeners,omitempty"`
+
+	// TLSMinVersion minimum TLS version that is acceptable.
+	// +optional
+	// +kubebuilder:validation:Enum=TLS10;TLS11;TLS12;TLS13
+	TLSMinVersion string `json:"tlsMinVersion,omitempty"`
+	// TLSCipherSuites defines list of supported cipher suites for TLS versions up to TLS 1.2
+	// https://golang.org/pkg/crypto/tls/#pkg-constants
+	// +optional
+	TLSCipherSuites []string `json:"tlsCipherSuites,omitempty"`
```
Haleygo: Why not put those two fields under `TCPListeners` as well, since they also work on `-syslog.listenAddr.tcp`?

internal/config/config.go
Haleygo: I still try to differentiate them when deploying, since they use the same image and binary name, which can sometimes cause confusion. We also provide [examples with different ports in the cluster setup](https://docs.victoriametrics.com/victorialogs/cluster/#quick-start), although the intention could be allowing running them all on one host. But that helps distinguish the services a bit.

internal/config/config.go
Haleygo: >And we recommend to have a CPU limit equal to the whole CORE.

I didn't know that:)

api/operator/v1/vlcluster_types.go
f41gh7: It's global configuration settings, not listener specific.

Haleygo: They're under `-syslog.xxx`, so only for `-syslog.listenAddr.tcp` https://docs.victoriametrics.com/victorialogs/#list-of-command-line-flags.
>  -syslog.tlsMinVersion string
    	The minimum TLS version to use for -syslog.listenAddr.tcp if -syslog.tls is set. Supported values: TLS10, TLS11, TLS12, TLS13. See https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#security (default "TLS13")

>  -syslog.tlsCipherSuites array
    	Optional list of TLS cipher suites for -syslog.listenAddr.tcp if -syslog.tls is set. See the list of supported cipher suites at https://pkg.go.dev/crypto/tls#pkg-constants . See also https://docs.victoriametrics.com/victorialogs/data-ingestion/syslog/#security
    	Supports an array of values separated by comma or specified via multiple flags.
    	Value can contain comma inside single-quoted or double-quoted string, {}, [] and () braces.

f41gh7: I think, it's a bad flags naming format. All syslog tcp listeneres shares the same values for these flags. It's not possible to use  `-tlsMinVersion=v1,v2,v3`, only `-tlsMinVersion=v1`.

internal/config/config.go
f41gh7: Okay, I checked and helm-chart already uses port numbers from this doc. So lets keep it consistent and use the same ports.

internal/config/config.go
f41gh7: https://github.com/VictoriaMetrics/VictoriaMetrics/commit/13c4324bb5c3cc8dcb5df5c938c73de9d911d324

It was changed recently.

api/operator/v1/vlcluster_types.go
Haleygo: I agree it has bad naming(having -syslog.xxx.tcp and -syslog.xxx.udp, then -syslog.xxx options which only makes sense to tcp) and value type, and I think putting both fields under `.TCPListeners` directly here can help a bit with the naming.

f41gh7: I just dropped those fields. If needed, it's possible to set it via extraArgs. I doubt that anyone will ever change it.

PR #1382, opened 2025-05-27
config/alerting: rm `BadObjects` rule as it is no-op

hagen1778:
The rule `BadObjects` uses `operator_controller_bad_objects_count` metric that was stopped from being exposed in https://github.com/VictoriaMetrics/operator/commit/6f35883c2d4766e66d899b432ed5acbc505545d9#diff-08d1eedc31f3c6403917e8f48ad89b21a0490d80bd3606c7b103028e90cc17fb


f41gh7:
Thanks for contribution!

PR #1392, opened 2025-05-30
fixed client generation

AndrewChubatiuk:
fixed client generation


f41gh7:
Thanks for contribution!

PR #1397, opened 2025-06-03
docs/configuration: Scrape operator metrics

makasim:


docs/configuration.md
```diff
@@ -183,6 +183,46 @@ kubectl get deployment -n vm vm-operator \
 # --leader-elect --health-probe-bind-address=:8081 --metrics-bind-address=:8080 -zap-log-level=debug
 ```
 
+## Scrape operator metrics
+
+To collect the operator metrics, you can create a [VMServiceScrape](https://docs.victoriametrics.com/operator/resources/vmservicescrape/) resource.
+Configure it to collects metrics from the pods that match the operator labels. 
+Apply it to the same namespace where the operator is running.
```
Haleygo: Under the same namespace is not mandatory, user can use `serviceScrapeSpec.namespaceSelector` to match them.

makasim: good point, I'll mention it

makasim: added, an example on how to configure namespace selector and changed a bit wording

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1399, opened 2025-06-05
VMAnomaly support

AndrewChubatiuk:
introduced VMAnomaly CR that manages statefulsets with ability to run in multiple shards and with replication support
initially supported CRs for multiple models and schedulers, but I would like to discuss current anomaly configuration structure, simplification of which can allow to introduce single model, scheduler, reader and writer CRs later


api/operator/v1beta1/vmextra_types.go
```diff
@@ -771,16 +771,16 @@ func (l *License) MaybeAddToArgs(args []string, secretMountDir string) []string
...
-		args = append(args, fmt.Sprintf("-license=%s", *l.Key))
+		args = append(args, fmt.Sprintf("--license=%s", *l.Key))
 	}
 	if l.KeyRef != nil {
-		args = append(args, fmt.Sprintf("-licenseFile=%s", path.Join(secretMountDir, l.KeyRef.Name, l.KeyRef.Key)))
+		args = append(args, fmt.Sprintf("--licenseFile=%s", path.Join(secretMountDir, l.KeyRef.Name, l.KeyRef.Key)))
 	}
 	if l.ForceOffline != nil {
-		args = append(args, fmt.Sprintf("-license.forceOffline=%v", *l.ForceOffline))
+		args = append(args, fmt.Sprintf("--license.forceOffline=%v", *l.ForceOffline))
```
f41gh7: This hidden change will trigger restart of all applications that uses license flag.

 Operator provides guarantees,  that version update will never trigger restart of VMSingle and VMCluster components. Please rollback this change. And make sure, that nothing else changed in the same manner.

api/operator/v1beta1/vmextra_types.go
```diff
@@ -766,7 +766,7 @@ func (l *License) IsProvided() bool {
 }
 
 // MaybeAddToArgs conditionally adds license commandline args into given args
-func (l *License) MaybeAddToArgs(args []string, secretMountDir string) []string {
+func (l *License) MaybeAddToArgs(args []string, secretMountDir string, doubleDash bool) []string {
```
Haleygo: Looks like doubleDash is only for VMAnomaly, could you add a comment for this?

internal/controller/operator/factory/build/deploy.go
```diff
@@ -0,0 +1,213 @@
+package build
```
Haleygo: Can we move this file to `internal/controller/operator/factory/vmanomaly` for now? Since it's only for vmanomaly, we can place it under build when it becomes publicly used in a future refactor.

AndrewChubatiuk: actually copied it from vmagent, which will reuse this functionality later as well

Haleygo: yeah, I saw that and understand the idea of reusing them for all components. However, since we're currently only using it for vmanomaly, I think it's better to keep them separate for clarity.

api/operator/v1/vmanomaly_types.go
```diff
@@ -0,0 +1,519 @@
...
+	ConfigSecret *corev1.SecretKeySelector `json:"configSecret,omitempty"`
+	// Metrics source for VMAnomaly
+	// See https://docs.victoriametrics.com/anomaly-detection/components/reader/
+	Readers *VMAnomalyReadersSpec `json:"readers,omitempty"`
+	// Metrics destination for VMAnomaly
+	// See https://docs.victoriametrics.com/anomaly-detection/components/writer/
+	Writers *VMAnomalyWritersSpec `json:"writers,omitempty"`
+	// StatefulMode enables StatefulSet for `VMAnomaly` instead of Deployment
+	// +optional
+	StatefulMode bool `json:"statefulMode,omitempty"`
```
f41gh7: Lets use `Statefulset`. VMAnomaly is state-full application. `Deployment` install type is only complicates overall code and configuration. There is no practical reason to use it.

api/operator/v1/vmanomaly_types.go
```diff
@@ -0,0 +1,519 @@
...
+}
+
+func (cr *VMAnomaly) GetShardCount() int {
+	if cr == nil || cr.Spec.ShardCount == nil {
+		return 0
+	}
+	return *cr.Spec.ShardCount
+}
+
+func (cr *VMAnomaly) AddShardSettings(dep runtime.Object, shardNum int) {
```
f41gh7: This function belongs to `build` or `factory/vmanomaly` package. It's not a part of `api`.

api/operator/v1/vmextra_types.go
```diff
@@ -0,0 +1,58 @@
...
+const (
+	TLSAssetKind     AssetKind = "tls"
+	ConfigAssetKind  AssetKind = "config"
+	DefaultAssetKind AssetKind = "default"
+)
+
+// +kubebuilder:object:generate=false
+// +kubebuilder:pruning:PreserveUnknownFields
+// +kubebuilder:validation:Schemaless
+type Args map[string]any
```
f41gh7: This struct is not used by `api` package. It also do not have proper comment. Consider to move it to the `build` or any other suitable place.

api/operator/v1beta1/vmextra_types_test.go
```diff
@@ -163,10 +164,25 @@ func TestLicense_MaybeAddToArgs(t *testing.T) {
...
+				secretMountDir: "/etc/secrets",
+				doubleDash:     true,
+			},
+			want: []string{"--licenseFile=/etc/secrets/license-secret/license-key"},
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got := tt.license.MaybeAddToArgs(tt.args.args, tt.args.secretMountDir)
+			got := tt.license.MaybeAddToArgs(tt.args.args, tt.args.secretMountDir, tt.args.doubleDash)
```
f41gh7: I think, it's also worth to move this args generation function into `build` package.

api/operator/v1/vmextra_types.go
AndrewChubatiuk: removed it, was used for vmanomaly model CRs

api/operator/v1beta1/vmextra_types_test.go
AndrewChubatiuk: done

api/operator/v1/vmanomaly_types.go
AndrewChubatiuk: removed deployment

api/operator/v1/vmanomaly_types.go
AndrewChubatiuk: updated

api/operator/v1beta1/vmextra_types.go
fred-navruzov: should be more convenient once single-dash arg support is added to vmanomaly - https://github.com/VictoriaMetrics/vmanomaly/pull/340

api/operator/v1beta1/vmextra_types.go
AndrewChubatiuk: fixed it

test/e2e/vmanomaly_test.go
```diff
@@ -0,0 +1,565 @@
...
+    fit_every: "4m"
+    fit_window: "1h"
+`
+
+//nolint:dupl,lll
+var _ = Describe("test vmanomaly Controller", Label("enterprise"), Ordered, func() {
+	ctx := context.Background()
+	namespace := "default"
+	BeforeAll(func() {
+		licenseKey := os.Getenv("LICENSE_KEY")
```
f41gh7: How contributor outside of VictoriaMetrics team should run vmanomaly test? 

 I suggest to add a fast path. If `licenseKey` is empty, print warning and exit earlier without any tests.

internal/controller/operator/factory/finalize/common.go
```diff
@@ -65,11 +65,11 @@ func RemoveFinalizer(ctx context.Context, rclient client.Client, instance client
...
 }
 
-func removeFinalizeObjByName(ctx context.Context, rclient client.Client, obj client.Object, name, ns string) error {
-	return removeFinalizeObjByNameWithOwnerReference(ctx, rclient, obj, name, ns, true)
+func RemoveFinalizeObjByName(ctx context.Context, rclient client.Client, obj client.Object, name, ns string) error {
+	return RemoveFinalizeObjByNameWithOwnerReference(ctx, rclient, obj, name, ns, true)
 }
 
-func removeFinalizeObjByNameWithOwnerReference(ctx context.Context, rclient client.Client, obj client.Object, name, ns string, keepOwnerReference bool) error {
+func RemoveFinalizeObjByNameWithOwnerReference(ctx context.Context, rclient client.Client, obj client.Object, name, ns string, keepOwnerReference bool) error {
```
f41gh7: Proper way of such refactoring is the following:
* introduce new public package method with proper comment
* keep exist private method and use it with public
* use new method for the new business logic ( keep previous logic as is).
* after changes merged - replace private method with new public at separate PR.

 This way keeps changes non-intrusive and easy to review.

internal/controller/operator/factory/build/build.go
```diff
@@ -18,7 +19,7 @@ func SetSkipRuntimeValidation(mustSkip bool) {
 	MustSkipRuntimeValidation = mustSkip
 }
 
-type builderOpts interface {
+type deployOpts interface {
```
f41gh7: please remove this change.

internal/controller/operator/factory/build/build.go
```diff
@@ -37,3 +43,34 @@ func PodDNSAddress(baseName string, podIndex int32, namespace string, portName s
 	}
 	return fmt.Sprintf("%s-%d.%s.%s.svc.%s:%s,", baseName, podIndex, baseName, namespace, domain, portName)
 }
+
+type ResourceKind string
```
f41gh7: What is resource kind? Is it Kubernetes `kind: Service` ?

internal/controller/operator/factory/build/cache.go
```diff
@@ -0,0 +1,538 @@
...
+}
+
+type HTTPClientCreds struct {
+	BasicAuth   *BasicAuthCreds
+	BearerToken string
+	OAuth2      *OAuth2Creds
+}
+
+// AssetsCache is a shared cache for all CR assets
+type AssetsCache struct {
```
f41gh7: This change doesn't look relevant to this PR. Please remove `cache.go` file.

internal/controller/operator/factory/build/deploy.go
f41gh7: I agree with Haley, please remove this change.

internal/controller/operator/factory/build/build.go
AndrewChubatiuk: it's predefined naming convention for each resource type: name for application resources, tls secrets, configuration secrets, streamaggr config secret

internal/controller/operator/factory/finalize/orphaned.go
```diff
@@ -95,3 +97,18 @@ func discoverSTSsByLabels(ctx context.Context, rclient client.Client, ns string,
 	}
 	return resp, nil
 }
+
+// RemoveOrphanedDaemonSet removes daemonset detached from given object
+func RemoveOrphanedDaemonSet(ctx context.Context, rclient client.Client, name, namespace string) error {
```
f41gh7: This change doesn't look relevant to this PR. Please remove it.

internal/controller/operator/factory/build/cache.go
AndrewChubatiuk: AssetsCache is used for TLS secrets, basic auth creds and bearer tokens loading in anomaly, which will be later used by other components

internal/controller/operator/factory/vmanomaly/config.go
```diff
@@ -0,0 +1,304 @@
...
+	Models     map[string]*model     `yaml:"models,omitempty"`
+	Reader     *reader               `yaml:"reader,omitempty"`
+	Writer     *writer               `yaml:"writer,omitempty"`
+	Monitoring *monitoring           `yaml:"monitoring,omitempty"`
+}
+
+func (c *config) override(cr *vmv1.VMAnomaly, ac *build.AssetsCache) error {
+	if cr.Spec.Readers != nil && len(cr.Spec.Readers.VM) > 0 {
+		vms := cr.Spec.Readers.VM
+		if len(vms) > 1 {
```
f41gh7: Should it be part of `crd.Validate`?

AndrewChubatiuk: thanks, moved it

internal/controller/operator/factory/build/build.go
AndrewChubatiuk: renamed it back

test/e2e/vmanomaly_test.go
AndrewChubatiuk: done

internal/controller/operator/factory/vmanomaly/deploy.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	if cr.IsOwnsServiceAccount() {
+		var prevSA *corev1.ServiceAccount
+		if prevCR != nil {
+			prevSA = build.ServiceAccount(prevCR)
+		}
+		if err := reconcile.ServiceAccount(ctx, rclient, build.ServiceAccount(cr), prevSA); err != nil {
+			return fmt.Errorf("failed create service account: %w", err)
+		}
+		if ptr.Deref(cr.Spec.UseVMConfigReloader, false) {
```
f41gh7: I think, config reloader is not needed, this configuration hash is written into annotations and it will trigger restart of application each time configuration is changed.

internal/controller/operator/factory/vmanomaly/models.go
```diff
@@ -0,0 +1,247 @@
...
+
+type modelDetectionDirection string
+
+const (
+	bothModelDirection          modelDetectionDirection = "both"           //nolint:unused
+	aboveExpectedModelDirection modelDetectionDirection = "above_expected" //nolint:unused
+	belowExpectedModelDirection modelDetectionDirection = "below_expected" //nolint:unused
+)
+
+type CommonModelParams struct {
```
f41gh7: Should it be private struct?

internal/controller/operator/factory/vmanomaly/models.go
```diff
@@ -0,0 +1,247 @@
+package vmanomaly
```
f41gh7: I think, it's better to move all configuration related files, such as `monitoring`, `writes`, `readers` into sub-package `vmanomaly/config` or keep it at `config.go` file.

 Also, those files have non-trivial parsing logic and I was not able to fine any tests for configuration building methods, neither links for configuration source file. 

 Could you please add tests in first place? 

 And if it's possible provide links to the configuration source. It'll simplify this code for the future readers.

internal/controller/operator/factory/vmanomaly/models.go
AndrewChubatiuk: done

internal/controller/operator/factory/vmanomaly/deploy.go
AndrewChubatiuk: removed it, used it in init only since anomaly doesn't support reloading, thought will be using it for decompression of config in k8s secret

internal/controller/operator/factory/vmanomaly/models.go
AndrewChubatiuk: moved it to separate package, will add tests soon

AndrewChubatiuk: added tests

internal/controller/operator/factory/finalize/orphaned.go
AndrewChubatiuk: removed

internal/controller/operator/factory/finalize/common.go
AndrewChubatiuk: reverted this change

internal/controller/operator/factory/build/deploy.go
AndrewChubatiuk: okay, removed

internal/controller/operator/factory/build/cache.go
Haleygo: To replace
https://github.com/VictoriaMetrics/operator/blob/d6b6873d8a735fdec30d811c3da791437afeba27/internal/controller/operator/factory/build/common.go#L49-L58?
why not extend the existing one?

internal/controller/operator/factory/build/cache.go
```diff
@@ -0,0 +1,538 @@
...
+		return "", fmt.Errorf("BUG, configmap key selector must be non nil in ns=%q", ns)
+	}
+	key := buildCacheKey(ns, cs.Name)
+	cm, ok := ac.configMaps[key]
+	if !ok {
+		cm = &corev1.ConfigMap{}
+		if err := ac.client.Get(ac.ctx, types.NamespacedName{Namespace: ns, Name: cs.Name}, cm); err != nil {
+			if k8serrors.IsNotFound(err) {
+				ac.configMaps[ns] = nil
+			}
```
Haleygo: Looks like the code should be `ac.configMaps[key] = nil`, but the configmap could be created next time, we shouldn't cache it if not found.

internal/controller/operator/factory/build/cache.go
```diff
@@ -0,0 +1,538 @@
...
+		return "", fmt.Errorf("BUG, secret key selector must be non nil in ns=%q", ns)
+	}
+	key := buildCacheKey(ns, ss.Name)
+	s, ok := ac.secrets[key]
+	if !ok {
+		s = &corev1.Secret{}
+		if err := ac.client.Get(ac.ctx, types.NamespacedName{Namespace: ns, Name: ss.Name}, s); err != nil {
+			if k8serrors.IsNotFound(err) {
+				ac.secrets[ns] = nil
+			}
```
Haleygo: same, we shouldn't cache it if not found.

AndrewChubatiuk:
> why not extend the existing one?

TLSConfig covers only TLS certs, AssetsCache instead covers, BearerTokens, BasicCreds and OAuth2Secrets as well. It allows to remove TLSConfig and also [this](https://github.com/VictoriaMetrics/operator/blob/d6b6873d8a735fdec30d811c3da791437afeba27/internal/controller/operator/factory/vmalert/vmalert.go#L558), [this](https://github.com/VictoriaMetrics/operator/blob/d6b6873d8a735fdec30d811c3da791437afeba27/internal/controller/operator/factory/vmagent/vmagent_scrapeconfig.go#L43) structures and [this](https://github.com/VictoriaMetrics/operator/blob/d6b6873d8a735fdec30d811c3da791437afeba27/internal/controller/operator/factory/k8stools/secret_content.go) file

internal/controller/operator/factory/build/cache.go
AndrewChubatiuk: thanks, pushed a change

internal/controller/operator/factory/build/cache.go
AndrewChubatiuk: TLSConfig covers only TLS certs, VMAnomaly also has basic and bearer auth. AssetsCache instead covers, BearerTokens, BasicCreds and OAuth2Secrets as well. It allows to remove TLSConfig and also [this](https://github.com/VictoriaMetrics/operator/blob/d6b6873d8a735fdec30d811c3da791437afeba27/internal/controller/operator/factory/vmalert/vmalert.go#L558), [this](https://github.com/VictoriaMetrics/operator/blob/d6b6873d8a735fdec30d811c3da791437afeba27/internal/controller/operator/factory/vmagent/vmagent_scrapeconfig.go#L43) structures and [this](https://github.com/VictoriaMetrics/operator/blob/d6b6873d8a735fdec30d811c3da791437afeba27/internal/controller/operator/factory/k8stools/secret_content.go) file

internal/controller/operator/factory/build/cache.go
Haleygo: why adding the key if the resource is not found? 
User might fix the issue afterwards, but operator will skip the check next time.

AndrewChubatiuk: this is not a permanent cache, it's scope is reconcile function execution and this nil is required to avoid calling API for absent secret multiple times within a single reconcile execution

AndrewChubatiuk: but actually in case of permanent cache it also makes sense to set to `nil`, but additionally ttl logic is required for this cache

api/operator/v1/vmanomaly_types.go
```diff
@@ -0,0 +1,473 @@
...
+	// Metrics destination for VMAnomaly
+	// See https://docs.victoriametrics.com/anomaly-detection/components/writer/
+	Writers *VMAnomalyWritersSpec `json:"writers,omitempty"`
+	// StatefulStorage configures storage for StatefulSet
+	// +optional
+	StatefulStorage *vmv1beta1.StorageSpec `json:"statefulStorage,omitempty"`
+	// StatefulRollingUpdateStrategy allows configuration for strategyType
+	// set it to RollingUpdate for disabling operator statefulSet rollingUpdate
+	// +optional
+	StatefulRollingUpdateStrategy appsv1.StatefulSetUpdateStrategyType `json:"statefulRollingUpdateStrategy,omitempty"`
```
Haleygo: Statefulset prefix looks unneeded.

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

f41gh7:
Feature was included into [v0.60.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.60.0) release

PR #1400, opened 2025-06-06
added entity selector, which combines namespace and object selector

AndrewChubatiuk:
- added entity selector, which combines namespace and object selector, which can be later used for vmagent, vmalertmanager and other parent CRs, which are collecting configuration from other CRs
- changed yaml library to one, which is still maintained
- renamed VisitObjectsForSelectorsAtNs to VisitSelected and signature now accepts EntitySelector instead of separate ns and object selectors
- moved selectNamespaces function tests to respective package and renamed selectNamespaces function to discoverNamespaces, since it also performs query to k8s API


api/operator/v1/vmextra_types.go
```diff
@@ -0,0 +1,21 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EntitySelectors namespace and object selector
+type EntitySelectors struct {
```
f41gh7: Any reason to have it at `api` package? It's not used by any api entities.

internal/controller/operator/factory/build/common.go
```diff
@@ -4,13 +4,43 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type keysSorter[T any] struct {
+	target []T
+	sorter []string
+}
+
+func OrderByKeys[T any](target []T, sorter []string) {
```
f41gh7, line 21: This function requires proper comment.

 Also, it's good practice to write comments for any public exported methods.

AndrewChubatiuk: added comment

api/operator/v1/vmextra_types.go
AndrewChubatiuk: initially was going to use it for vmanomaly, but then decided to not introduce separate CRD configurations due to complexity of configuration structure, also thought it would be nice to use it for scrapes, users and alertmanager config selectors in v1. removed it

go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/fsnotify/fsnotify v1.8.0
 	github.com/go-logr/logr v1.4.2
 	github.com/go-test/deep v1.1.1
+	github.com/goccy/go-yaml v1.18.0
```
f41gh7: looks like this import is not needed

AndrewChubatiuk: thanks, fixed

f41gh7:
Thanks for contribution

PR #1409, opened 2025-06-17
removed unneeded id from nodescrape target name

AndrewChubatiuk:
job name for scrapes has a following convention `<scrape-type>/<scrape-name>/<idx>` when multiple targets are available in scrape configuration and `<scrape-type>/<scrape-name>`, when a single target can be configured. nodescrape has a single target configurable but for some reason its job name uses`nodescrape/<scrape-name>/<idx>` naming. this PR removes confusing index from nodescrape job name


f41gh7:
Thanks for contribution!

f41gh7:
Change was included into [v0.60.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.60.0) release

PR #1410, opened 2025-06-18
refactoring: consistently name variables in method signatures

AndrewChubatiuk:
scrape CRs are named either `cr`, `m` or `sc`, renamed them all to `sc`
parent CRs were renamed to `cr`


f41gh7:
Thanks for contribution!

PR #1417, opened 2025-06-23
factory/vmanomaly/config: fix `latency_offset` and add `decay`

zekker6:
- Latency offset was previously incorrectly marshalled which caused an error
- Add support of decay field for online models.


f41gh7:
Thanks for contribution

f41gh7:
Bugfix was included into [v0.60.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.60.0) release

Issue #1136, opened 2024-10-30
Add vmanomaly CRD 


f41gh7:
[vmanomaly](https://docs.victoriametrics.com/anomaly-detection/) is a IA based tool for forecasting and prediction of events.

It should be deployed as kubernetes `Deployment` with optional persistent storage.

It should have flexible(unstructured if needed) configuration [params](https://docs.victoriametrics.com/anomaly-detection/overview/#config-file).

 It must require license or eula enabled param.

f41gh7:
@Haleygo @AndrewChubatiuk  I'll appreciate your input and suggestions.

f41gh7:
`VMAnomaly` must be `Statefulset`, since it will support sharding in the same way as `VMAgent` in helm charts. It also needs persistent store for models.

f41gh7:
@AndrewChubatiuk  Could you please add missing documentation and examples of usage.

AndrewChubatiuk:
VMAnomaly support was introduced in operator [v0.60.0](https://docs.victoriametrics.com/operator/changelog/#v0600)

PR #1423, opened 2025-06-24
docs: added vmanomaly CR doc

AndrewChubatiuk:
added doc for vmanomaly CR
fixes https://github.com/VictoriaMetrics/operator/issues/1136

### DIFF
--- a/docs/resources/vmanomaly.md
+++ b/docs/resources/vmanomaly.md
@@ -1,0 +1,437 @@
+---
+weight: 22
+title: VMAnomaly
+menu:
+  docs:
+    identifier: operator-cr-vmanomaly
+    parent: operator-cr
+    weight: 22
+aliases:
+  - /operator/resources/vmanomaly/
+  - /operator/resources/vmanomaly/index.html
+tags:
+  - kubernetes
+  - metrics
+  - logs
+  - anomaly
+---
+`VMAnomaly` - represents [VictoriaMetrics Anomaly Detection](https://docs.victoriametrics.com/anomaly-detection/) configuration {{% available_from "v0.60.0" "operator"%}}.
+
+The `VMAnomaly` CRD declaratively defines a desired Anomaly Detection setup to run in a Kubernetes cluster.
+It provides options to configure [sharding](https://docs.victoriametrics.com/anomaly-detection/scaling-vmanomaly/#horizontal-scalability), [replication](https://docs.victoriametrics.com/anomaly-detection/scaling-vmanomaly/#high-availability), persistent storage.
+
+For each shard of `VMAnomaly` resource, the Operator deploys a properly configured `StatefulSet` in the same namespace.
+Anomaly Detection pods can either use a configuration Secret created from the value specified in `spec.configRawYaml`, or use an existing secret specified in `spec.configSecret`.
+
+## Configuration
+
+The operator generates a configuration file for `VMAnomaly` based on user input at the definition of [CRD](https://docs.victoriametrics.com/operator/api/#vmanomaly).
+
+Generated config stored at `Secret` created by the operator, it has the following name template `config-vmanomaly-CRD_NAME`.
+
+This configuration file is mounted at `VMAnomaly` `Pod`.
+
+**VMAnomaly is enterprise only, license is required for this CRD**. [Trial license can be requested](https://victoriametrics.com/products/enterprise/trial/) for `VMAnomaly` CRD evaluation.
+
+In contrast to Anomaly Detection configuration, `VMAnomaly` CRD requires to define [`reader`](https://docs.victoriametrics.com/anomaly-detection/components/reader/), [`writer`](https://docs.victoriametrics.com/anomaly-detection/components/writer/) and [`monitoring`](https://docs.victoriametrics.com/anomaly-detection/components/monitoring/) sections in raw configuration (except `reader.queries`, which should be defined at `spec.configRawYaml` or `spec.configSecret`). `reader`, `writer` and `monitoring` have dedicated sections in `VMAnomaly` CRD:
+
+```yaml
+apiVersion: operator.victoriametrics.com/v1
+kind: VMAnomaly
+metadata:
+  name: example
+spec:
+  replicaCount: 2
+  license:
+    key: "xx"
+  configRawYaml: |
+    reader:
+      queries:
+        ingestion_rate:
+          expr: 'sum(rate(vm_rows_inserted_total[5m])) by (type) > 0'
+          step: '1m'
+    schedulers:
+      scheduler_periodic_1m:
+        class: "periodic"
+        # class: "periodic" # or class: "scheduler.periodic.PeriodicScheduler" until v1.13.0 with class alias support)
+        infer_every: "1m"
+        fit_every: "2m"
+        fit_window: "3h"
+    models:
+      model_univariate_1:
+        class: 'zscore'
+        z_threshold: 2.5
+  reader:
+    datasourceURL: http://vmsingle-read-example:8428
+    samplingPeriod: 10s
+  writer:
+    datasourceURL: http://vmsingle-write-example:8428
+  monitoring:
+    push:
+      url: http://vmsingle-monitoring-example:8428
+```
+
+### Reader, Writer and Monitoring
+
+While Anomaly Detection [models](https://docs.victoriametrics.com/anomaly-detection/components/models/), [schedulers](https://docs.victoriametrics.com/anomaly-detection/components/schedulers/) and [settings](https://docs.victoriametrics.com/anomaly-detection/components/settings/) should be defined in `spec.configRawYaml` or `spec.configSecret`, [reader](https://docs.victoriametrics.com/anomaly-detection/components/reader/), [writer](https://docs.victoriametrics.com/anomaly-detection/components/writer/) and [monitoring](https://docs.victoriametrics.com/anomaly-detection/components/monitoring/) sections should defined at `spec.reader`, `spec.writer` and `spec.monitoring` respectively.
+
+This was done to allow to use K8s secrets and configmaps as a source for TLS, basic and bearer secrets for reader, writer and monitoring endpoints. Also structure of this sections differ from Anomaly Detection configuration structure.
+
+Given below reader, writer and monitoring configuration
+
+```yaml
+reader:
+  class: vm
+  datasource_url: http://localhost:8428
+  sampling_period: 10s
+  bearer_token: token
+  queries:
+    vmb:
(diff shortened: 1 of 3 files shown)


makasim:
> removed resource kind prefix/suffix from examples names, since the same prefix is added by operator to all managed resources

It would be better to move these changes to a separate PR, as they are unrelated to the vmanomaly documentation.

makasim:
I suggest adding vmanomaly to the top-level resources at the beginning of the [Resources section](https://docs.victoriametrics.com/operator/resources/#).

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,433 @@
...
+  - /operator/resources/vmanomaly/index.html
+tags:
+  - kubernetes
+  - metrics
+  - logs
+---
+`VMAnomaly` - represents [VictoriaMetrics Anomaly Detection](https://docs.victoriametrics.com/anomaly-detection/) configuration.
+
+The `VMAnomaly` CRD declaratively defines a desired Anomaly Detection setup to run in a Kubernetes cluster.
+It provides options to configure sharding, replication, persistent storage.
```
makasim: Can we link to some pages where these concepts are explained? 

cc @fred-navruzov

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,433 @@
...
+
+For each shard of `VMAnomaly` resource, the Operator deploys a properly configured `StatefulSet` in the same namespace.
+Anomaly Detection pods can either manage configuration `Secret` creating it from `spec.configRawYaml` value or use existing one when `spec.configSecret` is defined.
```
makasim:
```suggestion
Anomaly Detection pods can either use a configuration Secret created from the value specified in `spec.configRawYaml`, or use an existing secret specified in `spec.configSecret`.
```

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,433 @@
...
+
+## Configuration
+
+The operator generates a configuration file for `VMAnomaly` based on user input at the definition of `CRD`.
```
makasim: add a link to CRD definition

AndrewChubatiuk:
> I suggest adding vmanomaly to the top-level resources at the beginning of the [Resources section](https://docs.victoriametrics.com/operator/resources/#).

thanks, just added

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,433 @@
...
+        vmb:
+          expr: avg(vm_blocks)
+          step: 10s
+          data_range: ['-inf', 'inf']
+          tz: UTC
+```
+
+### Using secret
+
+Basically, you can use the global configuration defined at manually created `Secret`. This `Secret` must be created before `VMAnomaly`.
```
makasim:
```suggestion
You can use a global configuration defined in a manually created `Secret`. This `Secret` must be created before `VMAnomaly` resource.
```

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,434 @@
...
+  - /operator/resources/vmanomaly/index.html
+tags:
+  - kubernetes
+  - metrics
+  - logs
+  - anomaly
+---
+`VMAnomaly` - represents [VictoriaMetrics Anomaly Detection](https://docs.victoriametrics.com/anomaly-detection/) configuration.
+
+The `VMAnomaly` CRD declaratively defines a desired Anomaly Detection setup to run in a Kubernetes cluster.
```
makasim, line 20: We should mention the operator version this CRD is available from

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,434 @@
...
+  writer:
+    datasourceURL: http://vmsingle-write-example:8428
+  configSecret:
+    name: anomaly-config
+    key: config.yaml
+```
+
+### Using inline raw config
+
+Also, if there is no secret data at configuration, or you just want to redefine some global variables for `anomaly`.
```
makasim: > redefine some global variables for `anomaly`

I do not understand this part, what does it mean ?

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,434 @@
...
+      model_univariate_1:
+        class: 'zscore'
+        z_threshold: 2.5
+```
+
+If both `configSecret` and `configRawYaml` are defined, only configuration from `configRawYaml` will be used. Values from `configSecret` will be ignored.
+
+## High Availability and Sharding
+
+Anomaly Detection can be deployed in HA mode and can be scaled horizontally. While HA is configurable via amount of replicas `spec.replicaCount`, scaling is implementing via [automatic splitting of whole configuration into shards](https://new.docs.victoriametrics.com/anomaly-detection/scaling-vmanomaly/#sub-configuration) using `spec.shardCount` parameter. Operator creates `StatefulSet` with a count, that is equal to `spec.shardCount` and amount of replicas in each shard equals to `spec.replicaCount`.
```
makasim:
```suggestion
Anomaly Detection can be deployed in HA mode and scaled horizontally. Horizontal scaling is achieved by [automatically splitting the configuration into shards](https://new.docs.victoriametrics.com/anomaly-detection/scaling-vmanomaly/#sub-configuration) using the `spec.shardCount` parameter. The `spec.replicaCount` parameter defines the number of replicas per shard. The Operator creates one StatefulSet per shard (equal to `spec.shardCount`), with each StatefulSet containing `spec.replicaCount` replicas.
```

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,434 @@
...
+        infer_every: "1m"
+        fit_every: "2m"
+        fit_window: "3h"
+    models:
+      model_univariate_1:
+        class: 'zscore'
+        z_threshold: 2.5
+```
+
+Please note that with sharding enabled amount of queries at `spec.configRawYaml.reader.queries` should be greater than a number of shards, otherwise setup fails.
```
makasim: > otherwise setup fails.

fails at runtime or kube apply ?

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,434 @@
...
+Field `resources` in `VMAnomaly` spec have higher priority than operator parameters.
+
+If you set `VM_VMANOMALY_USEDEFAULTRESOURCES` to `false` and don't specify `resources` in `VMAnomaly` CRD,
+then `VMAnomaly` pods will be created without resource requests and limits.
+
+Also, you can specify requests without limits - in this case default values for limits will not be used.
+
+## Examples
+
+```yaml
```
makasim, line 406: Add a short explanation of what this example is about

makasim (changes requested):

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,434 @@
...
+
+The operator generates a configuration file for `VMAnomaly` based on user input at the definition of `CRD`.
+
+Generated config stored at `Secret` created by the operator, it has the following name template `config-vmanomaly-CRD_NAME`.
+
+This configuration file is mounted at `VMAnomaly` `Pod`.
+
+**VMAnomaly is enterprise only, license is required for this CRD**. [Trial license can be requested](https://victoriametrics.com/products/enterprise/trial/) for `VMAnomaly` CRD evaluation.
+
+In contrast to Anomaly Detection configuration, `VMAnomaly` CRD requires to define [`reader`](https://docs.victoriametrics.com/anomaly-detection/components/reader/), [`writer`](https://docs.victoriametrics.com/anomaly-detection/components/writer/) and [`monitoring`](https://docs.victoriametrics.com/anomaly-detection/components/monitoring/) sections in raw configuration (except `reader.queries`, which should be defined at `spec.configRawYaml` or `spec.configSecret`). `reader`, `writer` and `monitoring` have dedicated sections in `VMAnomaly` CRD:
```
makasim, line 36: Just a side note (not something we need to address in this PR): this separation feels a bit awkward and inconsistent from a user’s perspective. I understand the reasoning behind it (e.g., enabling the use of secrets for the reader, etc.), but it still feels off.

docs/resources/vmanomaly.md
AndrewChubatiuk: logo with link to a changelog should be after a first sentence

docs/resources/vmanomaly.md
AndrewChubatiuk: rephrased it

docs/resources/vmanomaly.md
AndrewChubatiuk: initially idea was to do in the same manner how it's done in vmalertmanager, but vmanomaly config structure cannot be straightforwardly ported to operator CR

docs/resources/vmanomaly.md
AndrewChubatiuk: removed sentence as it's application specific and validation is present for this case

docs/resources/vmanomaly.md
AndrewChubatiuk: added links

docs/resources/vmanomaly.md
```diff
@@ -0,0 +1,437 @@
...
+It provides options to configure [sharding](https://docs.victoriametrics.com/anomaly-detection/scaling-vmanomaly/#horizontal-scalability), [replication](https://docs.victoriametrics.com/anomaly-detection/scaling-vmanomaly/#high-availability), persistent storage.
+
+For each shard of `VMAnomaly` resource, the Operator deploys a properly configured `StatefulSet` in the same namespace.
+Anomaly Detection pods can either use a configuration Secret created from the value specified in `spec.configRawYaml`, or use an existing secret specified in `spec.configSecret`.
+
+## Configuration
+
+The operator generates a configuration file for `VMAnomaly` based on user input at the definition of [CRD](https://docs.victoriametrics.com/operator/api/#vmanomaly).
+
+Generated config stored at `Secret` created by the operator, it has the following name template `config-vmanomaly-CRD_NAME`.
```
makasim:
```suggestion
The generated configuration is stored in a Secret created by the operator, using the following naming template: config-vmanomaly-<CRD_NAME>
```

docs/resources/vmanomaly.md
makasim: Are there plans to improve it in the next version?

makasim (approved):
LGTM

docs/resources/vmanomaly.md
AndrewChubatiuk: There'll be VMAnomalyModel and VMAnomalyScheduler later, but also with a structure that differs from VMAnomaly app config structure, also hot-reload support should be implemented in VMAnomaly to make, otherwise these CRs are useless

PR #1424, opened 2025-06-24
factory/vmanomaly/config: fix marshalling of writer.metricFormat

zekker6:
Previously, it was not possible to add custom labels for writer. Replaced struct to a map of values since it was not possible to embed both named fields and a map, that would cause a failure of api-gen.

Implemented migration of previously supported `name` to `__name__` field as vmanomaly expects a specific name.

See https://docs.victoriametrics.com/anomaly-detection/components/writer/#metrics-formatting for reference of vmanomaly config.


api/operator/v1/vmanomaly_types.go
```diff
@@ -111,20 +111,11 @@ type VMAnomalyWritersSpec struct {
 	DatasourceURL string `json:"datasourceURL" yaml:"datasource_url,omitempty"`
 	// Metrics to save the output (in metric names or labels)
 	// +optional
-	VMAnomalyVMWriterMetricFormatSpec `json:"metricFormat,omitempty" yaml:"metric_format,omitempty"`
```
AndrewChubatiuk, line 114: expected this to be fixed on anomaly side and add map for custom labels instead

f41gh7: I also think, it's better to have an additional API field `extraLabels` and keep name and for required.

zekker6: Added `extraLabels`in https://github.com/VictoriaMetrics/operator/pull/1424/commits/5aa2ba3b000f6f02550978f0b7ca6dc64115c866
PTAL

f41gh7 (approved):
LGTM

AndrewChubatiuk (approved):

f41gh7:
Thanks for contribution!

Issue #1426, opened 2025-06-24
vmanomaly: `reason` field is not cleared after resource status change


zekker6:
Step by step:
1. Create a vmanomaly resource with incorrect configuration.
2. Resource is transitioned to failed state with `reason` containing an error message.
3. Fix the reason and apply new configuration.
4. Resource status changes to `expanding`, but `reason` still contains an error message. For example:
```
  reason: |-
    failed to unmarshal anomaly configuration, name="example": yaml: unmarshal errors:
      line 1: field readers not found in type config.config
  updateStatus: expanding
```

f41gh7:
It'd be great to check observedGeneration and if it differs from status generation, transit status to expanding.

zekker6:
Here is a full status before:
```
status:
  observedGeneration: 1
  reason: 'failed to validate anomaly configuration, name="example": failed to validate
    model="mad_cpu_spikes": decay must be in range [0, 1], got 1.993000'
  shards: 2
  updateStatus: failed
```
And after:
```
status:
  observedGeneration: 2
  reason: 'failed to validate anomaly configuration, name="example": failed to validate
    model="mad_cpu_spikes": decay must be in range [0, 1], got 1.993000'
  shards: 2
  updateStatus: expanding
```

AndrewChubatiuk:
fix is available in operator release 0.60.2, operator chart release 0.50.3 and k8s stack release 0.55.2

PR #1431, opened 2025-06-25
factory/reconcile: reset reason when transitioning into expanding state

zekker6:
Previously, `reason` was only reset when resource was transitioning to `operational` state. That was confusing to the end user as the resource workflow was:
1. Deploy a broken version, observe `updateStatus: failed`, `reaason: {error}`
2. Deploy a fix, observe `updateStatus: expanding`, `reason: {error}` - this is confusing as it is expected that error no longer appears after fix.

See: https://github.com/VictoriaMetrics/operator/issues/1426

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -13,6 +13,8 @@ aliases:
 
 ## tip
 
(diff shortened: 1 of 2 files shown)


f41gh7:
Thanks for contribution!

PR #1432, opened 2025-06-25
docs: removed resource kind prefix from operator managed resources in examples

AndrewChubatiuk:
removed CR names prefixes and suffixes, which are duplicate ones that are already added by operator to all managed resources


makasim (approved):
LGTM.

However, I noticed that a change in GitHub Actions is missing from both this and the previous PR. Was that intentional?

AndrewChubatiuk:
yes, will add this separately

AndrewChubatiuk:
@makasim 
https://github.com/VictoriaMetrics/operator/pull/1433

PR #1439, opened 2025-06-27
olm: added absent examples for VLCluster and VLSingle

AndrewChubatiuk:
image

### DIFF
--- a/config/examples/kustomization.yaml
+++ b/config/examples/kustomization.yaml
@@ -18,3 +18,5 @@ resources:
 - vmstaticscrape.yaml
 - vmscrapeconfig.yaml
 - vlogs.yaml
+- vlsingle.yaml
+- vlcluster.yaml


config/examples/kustomization.yaml
```diff
@@ -18,3 +18,6 @@ resources:
 - vmstaticscrape.yaml
 - vmscrapeconfig.yaml
 - vlogs.yaml
+- vlsingle.yaml
+- vlcluster.yaml
+- vmanomaly.yaml
```
f41gh7: As far as I know, olm performs tests for given examples. And there is a problem, it's not possible to run vmanomaly example without License...

AndrewChubatiuk: custom image with anomaly mock in example could solve this issue

f41gh7: Could you please add link to this mock image? As far as I see, example doesnt use any mock https://github.com/VictoriaMetrics/operator/blob/master/config/examples/vmanomaly.yaml

AndrewChubatiuk: i mean i could add this image

f41gh7: I suggest to exclude vmanomaly, it's not possible to install it without external secret with license.

AndrewChubatiuk: excluded

f41gh7 (approved):
LGTM

Issue #1228, opened 2025-01-28
Add CRD for streaming aggregation rules


tiny-pangolin:
in order to make adding streaming aggregation easier to use in the operator there should be a resource similar to VMRule that allows users to define a streaming aggregation rule as a CRD.

f41gh7:
Related issue:
https://github.com/VictoriaMetrics/operator/issues/788

luistilingue:
It'll be great!

PR #1421, opened 2025-06-24
refactoring: share stream aggregation config, moved VM-specific flag generation to build package

AndrewChubatiuk:
this PR shares stream aggregation volume and cmd args generation. Since stream aggregation is also a part of vmagent remote write configuration which has another logic for cmd args generation, this flags generation logic was also moved to build package.

PR #1440, opened 2025-06-30
Introduce VMStreamAggrRule for VMSingle and VMAgent

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1228
fixes https://github.com/VictoriaMetrics/operator/issues/788

PR depends on https://github.com/VictoriaMetrics/operator/pull/1421


f41gh7:
@Haleygo  could you please take a look?

makasim:
We need documentation with some basic real-world usage examples.

Haleygo:
After discussion, we decided to hold this pull request because:
1. we don't see high demands for managing separate VMStreamAggrRule;
2. it's not clear how to resolve possible rule conflicts;

In general, streamaggregation should be managed by system admin, who can modify the vmagent yaml, to reduce system load from the higher level.
For end user like application developer, if they want to do aggregation to their metrics, it's better to it using the prometheus client sdk when exposing the metrics, which is more resource-efficient.

cc @AndrewChubatiuk @f41gh7

PR #1451, opened 2025-07-04
vmagent: set remoteWrite.maxDiskUsagePerURL automatically in stateful mode with volume claim template set

### LINKED ISSUE #1209: adjust `maxDiskUsagePerURL` based on `VMAgent` configuration.
Currently there is hard-coded value of `1GB`.  It'd be great to auto-magically change it based on the following conditions:
1) `statefulMode: true` 
2) `statefulStorage.volumeClaimTemplate` is defined.

 Use `storage.requests` from `volumeClaimTemplate` as default value for `maxDiskUsagePerURL` flag.

AndrewChubatiuk:
fixes #1209

in stateful mode with volume claim template set default value of `remoteWrite.maxDiskUsagePerURL` flag is set to `(storage requests - total size of all explicitly defined flags) / amount of remotewrites with undefined flag`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -13,6 +13,8 @@ aliases:
 
 ## tip
 
+* FEATURE: [vmagent](https://docs.victoriametrics.com/operator/resources/vmagent/): Modify default `remoteWrite.maxDiskUsagePerURL` when statefulStorage.volumeClaimTemplate is set. See [#1209](https://github.com/VictoriaMetrics/operator/issues/1209)
+
 ## [v0.60.2](https://github.com/VictoriaMetrics/operator/releases/tag/v0.60.2)
 
 **Release date:** 27 June 2025
--- a/internal/controller/operator/factory/vmagent/vmagent.go
+++ b/internal/controller/operator/factory/vmagent/vmagent.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/VictoriaMetrics/VictoriaMetrics/lib/flagutil"
 	"gopkg.in/yaml.v2"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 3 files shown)


Haleygo (commented):
We could append `-remoteWrite.maxDiskUsagePerURL` args in both [buildRemoteWrites](https://github.com/VictoriaMetrics/operator/blob/1b6a4411fd695e24d93eb25b3f3bc33194f190e4/internal/controller/operator/factory/vmagent/vmagent.go#L1101C6-L1101C23) and [buildRemoteWriteSettings](https://github.com/VictoriaMetrics/operator/blob/1b6a4411fd695e24d93eb25b3f3bc33194f190e4/internal/controller/operator/factory/vmagent/vmagent.go#L1003C6-L1003C30), and the check is incomplete in `buildRemoteWriteSettings`, which might cause conflicts. For example:
```
			RemoteWrite: []vmv1beta1.VMAgentRemoteWriteSpec{
				{
					URL: "localhost:8429",
				},
				{
					URL: "localhost:8431",
				},
				{
					URL: "localhost:8432",
				},
			},
			RemoteWriteSettings: &vmv1beta1.VMAgentRemoteWriteSettings{
				MaxBlockSize: ptr.To(int32(1000)),
			},
			StatefulMode: true,
			StatefulStorage: &vmv1beta1.StorageSpec{
				VolumeClaimTemplate: vmv1beta1.EmbeddedPersistentVolumeClaim{
					Spec: corev1.PersistentVolumeClaimSpec{
						StorageClassName: ptr.To("embed-sc"),
						Resources: corev1.VolumeResourceRequirements{
							Requests: map[corev1.ResourceName]resource.Quantity{
								corev1.ResourceStorage: resource.MustParse("10Gi"),
							},
						},
					},
				},
			},
```
Would be great if we can migrate the two functions.

AndrewChubatiuk:
thanks @Haleygo
updated code and added a test for this case

Issue #1179, opened 2024-12-04
vmagent: add support of global relabeling configuration


zekker6:
Starting from [v1.106.0](https://docs.victoriametrics.com/changelog/#v11060) vmagent and VictoriaMetrics single-node support `metric_relabel_configs` and `relabel_configs` as a part of the global configuration.

It would be great to have an option to provide global relabel configs via operator as well.

f41gh7:
@Haleygo  could you please take a look? Currently operator provides own per resource configuration for relabeling. It's more flexible, but maybe we should switch to the native `vmagent` implementation.

f41gh7:
The feature was included into [v0.61.0](https://docs.victoriametrics.com/operator/changelog/#v0610) release

PR #1453, opened 2025-07-04
vmagent: support all global scrape configuration parameters

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1179


api/operator/v1beta1/common_scrapeparams.go
```diff
@@ -17,6 +17,22 @@ type AttachMetadata struct {
 	Node *bool `json:"node,omitempty"`
 }
 
+type GlobalScrapeConfig struct {
```
f41gh7: VMAgent already exposes global scrape configuration params at `spec.` root - `ScrapeInterval`, `ScrapeTimeout` and `ExternalLabels`.

 I'd suggest to add `MetricRelabelConfigs` and `RelabelConfigs` with proper comments.

AndrewChubatiuk: I've copied ScrapeInterval, ScrapeTimeout, ExternalLabels and marked them as deprecated in VMAgentSpec

f41gh7: I'm not happy with this change. We cannot simply deprecate field without a good reason.

 Initial implementation aligns with Prometheus operator resource and provides similar experience for users.

AndrewChubatiuk: > We cannot simply deprecate field without a good reason

vmagentspec in this case will have RelabelConfig, InlineRelabelConfig and these new RelabelConfigs and MetricRelabelConfigs. Expecting multiple similar issues

f41gh7: I think it's fine, also `globalScrape` prefix should be enough - `globalScrapeRelabelConfigs` and `globalScrapeMetricRelabelConfigs`.

AndrewChubatiuk: done

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1455, opened 2025-07-05
added forceSubdomain option to have predictable pod urls

AndrewChubatiuk:
related issue https://github.com/VictoriaMetrics/helm-charts/issues/2283


f41gh7:
Sorry I don't follow how this option helps. subdomain and hostname pod properties should be relevant to stand-alone pods, not pod managed by statefulset/deployment/daemonset.

fmunteanu:
@f41gh7 I'm using an 8 nodes K3s cluster (3 cd + 5 workers) with VKMS helm chart. As explained into issue, as soon as I set the alert manager pods higher than 1, the following alarm fires:

image

PR #1460, opened 2025-07-08
refactoring: rewrote tests using f-test pattern

AndrewChubatiuk:
rewrote tests using f-test pattern

### DIFF
--- a/api/operator/v1beta1/vmagent_types_test.go
+++ b/api/operator/v1beta1/vmagent_types_test.go
@@ -5,42 +5,64 @@ import (
 )
 
 func TestVMAgent_Validate(t *testing.T) {
-
-	tests := []struct {
-		name    string
-		spec    VMAgentSpec
+	type opts struct {
+		cr      *VMAgent
 		wantErr bool
-	}{
-		{
-			name:    "wo remotewrite",
-			spec:    VMAgentSpec{},
-			wantErr: true,
+	}
+	f := func(opts opts) {
+		t.Helper()
+		if err := opts.cr.Validate(); (err != nil) != opts.wantErr {
+			t.Errorf("Validate() error = %v, wantErr %v", err, opts.wantErr)
+		}
+	}
+
+	// wo remotewrite
+	o := opts{
+		cr:      &VMAgent{},
+		wantErr: true,
+	}
+	f(o)
+
+	// rw empty url
+	o = opts{
+		cr: &VMAgent{
+			Spec: VMAgentSpec{
+				RemoteWrite: []VMAgentRemoteWriteSpec{
+					{},
+				},
+			},
 		},
-		{
-			name: "rw empty url",
-			spec: VMAgentSpec{RemoteWrite: []VMAgentRemoteWriteSpec{
-				{},
-			}},
-			wantErr: true,
-		},
-		{
-			name: "bad inline cfg",
-			spec: VMAgentSpec{
+		wantErr: true,
+	}
+	f(o)
+
+	// bad inline cfg
+	o = opts{
+		cr: &VMAgent{
+			Spec: VMAgentSpec{
 				RemoteWrite:        []VMAgentRemoteWriteSpec{{URL: "http://some-rw"}},
 				InlineScrapeConfig: "some; none yaml formatted string",
 			},
-			wantErr: true,
 		},
-		{
-			name: "valid inline cfg",
-			spec: VMAgentSpec{
+		wantErr: true,
+	}
+	f(o)
+
+	// valid inline cfg
+	o = opts{
+		cr: &VMAgent{
+			Spec: VMAgentSpec{
 				RemoteWrite:        []VMAgentRemoteWriteSpec{{URL: "http://some-rw"}},
 				InlineScrapeConfig: `key: value`,
 			},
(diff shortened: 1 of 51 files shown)


f41gh7:
It's better to rewrite tests with `f-tests` pattern. But currently, it's not worth it.

AndrewChubatiuk:
@f41gh7 rewrote to f-tests

f41gh7 (changes requested):
Sorry, It's impossible to review this PR. And I have no confidence to blindly merge it.

f41gh7:
@Haleygo  could you please take a look?

internal/controller/operator/converter/apis_test.go
```diff
@@ -14,259 +14,243 @@ import (
...
+		},
+		want: vmv1beta1.VMServiceScrape{
+			Spec: vmv1beta1.VMServiceScrapeSpec{
+				Endpoints: []vmv1beta1.Endpoint{
+					{
+						EndpointRelabelings: vmv1beta1.EndpointRelabelings{
+							MetricRelabelConfigs: []*vmv1beta1.RelabelConfig{
+								{
+									Action:       "drop",
+									SourceLabels: []string{"__meta__instance"},
```
Haleygo:
```suggestion
									SourceLabels: []promv1.LabelName{"__meta__instance"},
```
nit: we can replace them all

internal/controller/operator/converter/v1alpha1/apis_test.go
```diff
@@ -16,19 +16,24 @@ import (
 )
 
 func TestConvertAlertmanagerConfig(t *testing.T) {
-	f := func(name string, promCfg *promv1alpha1.AlertmanagerConfig, validate func(convertedAMCfg *vmv1beta1.VMAlertmanagerConfig) error) {
```
Haleygo: Why change this test, it's already f-test pattern, no? And I prefer listing all the parameters if can than having an extra opts struct.

Haleygo (commented):
I see two kinds of f test here:
1. creating an `opts{}` anyway;
2. listing the parameters of `f()` when they are limited and clear, like in TestParsingMatch.

I prefer 2 in most cases to avoid repeating `f(o)`, and make parameters clearer, especially for booleans(now they are omitted sometimes as the default false, which is more error-prone).
wdyt? Shall we stick to 2 unless parameters become too complex?

internal/controller/operator/converter/v1alpha1/apis_test.go
AndrewChubatiuk: with separate struct for test params it's obvious for each case which entity represents each parameter

Haleygo: It's obvious to me when there are less than 4 parameters, for example:
```
1.
	type opts struct {
		cr      *VMAgent
		wantErr bool
	}
       f := func(opts opts) {}

2.
   f := func(cr *VMAgent, wantErr bool) {}
```

AndrewChubatiuk: `internal/controller/operator/converter` package is not in diff already

internal/controller/operator/converter/apis_test.go
AndrewChubatiuk: this package is not in diff already

AndrewChubatiuk:
> Shall we stick to 2 unless parameters become too complex?

it's and approach, which i was following, for cases with 2 and more arguments

Haleygo:
> > Shall we stick to 2 unless parameters become too complex?
> 
> it's and approach, which i was following, for cases with 2 and more arguments

I mean shall we use p2 unless there are more than like 3 or 4 parameters?
For example
```
func TestVMAgent_Validate(t *testing.T) {
	type opts struct {
		cr      *VMAgent
		wantErr bool
	}
	f := func(opts opts) {
		t.Helper()
		if err := opts.cr.Validate(); (err != nil) != opts.wantErr {
			t.Errorf("Validate() error = %v, wantErr %v", err, opts.wantErr)
		}
	}

	// wo remotewrite
	o := opts{
		cr:      &VMAgent{},
		wantErr: true,
	}
	f(o)
```
I don't see the need of opts here.

AndrewChubatiuk:
but for the same test, when vmagent definition is huge, that boolean flag is less visible, as it may look like one of CR params

Haleygo:
> but for the same test, when vmagent definition is huge, that boolean flag is less visible, as it may look like one of CR params

```
	o = opts{
		cr: &VMAgent{
			Spec: VMAgentSpec{
				RemoteWrite: []VMAgentRemoteWriteSpec{
					{},
				},
			},
		},
		wantErr: true,
	}
```
I don't see how the opts improves the case, compare to 
```
f(&VMAgent{
		Spec: VMAgentSpec{
			RemoteWrite: []VMAgentRemoteWriteSpec{
				{},
			},
		},
	},
	true,
)
```

AndrewChubatiuk:
@Haleygo 
could you please take a look at PRs #1650 #1651 and #1652
added opts struct there only for tests with small number of arguments

AndrewChubatiuk:
closing as outdated

Issue #1365, opened 2025-05-16
Adjust requestsLoadBalancer image when x.xxx.x-enterprise-cluster clusterVersion is used.


makasim:
When setting clusterVersion: "v1.117.1-enterprise-cluster" in a VMCluster spec, the operator tries to use the same tag (v1.117.1-enterprise-cluster) for all components, including vmauth. However, there is no corresponding image tag v1.117.1-enterprise-cluster for the vmauth component. Only v1.117.1-enterprise is available.

Example (Does not work):
```
piVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
spec:
  clusterVersion: "v1.117.1-enterprise-cluster"

  # ...

  requestsLoadBalancer:
    enabled: true
```
This results in an image pull error for vmauth, as the specified tag does not exist.

Workaround (Works):
```

piVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
spec:
  clusterVersion: "v1.117.1-enterprise-cluster"

  # ...

  requestsLoadBalancer:
    enabled: true
    spec:
      image:
        tag: "v1.117.1-enterprise"
```

Suggested Improvement:

The operator should be able to detect when a component (e.g., vmauth) does not have a tag matching clusterVersion and fall back to a valid, compatible tag automatically (e.g., v1.117.1-enterprise). 

Alternatively, it could fail with validation error but I personally dont like it.

f41gh7:
Upcoming release will fix the issue with:

```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
spec:
  clusterVersion: "v1.117.1-enterprise-cluster"
  # ...
  requestsLoadBalancer:
    enabled: true
```

 When operator implicitly set `requestsLoadBalancer.spec.image.tag` to `v1.117.1-enterprise-cluster`. Now operator ignores value of `spec.clusterVersion` for `requestsLoadBalancer`. 

 The purpose of `clusterVersion` is management of cluster components version - `vmselect`, `vmstorage` and `vminsert`. It shouldn't change version of `vmbackupmanager` and `requestsLoadbalancer`. Since those components are supplemental and should be managed separately. The second reason is `VLCluster`, which also has `clusterVersion` spec and it's not possible to use it as `vmauth` version. So it also keeps behavior consistent.

f41gh7:
[...]

PR #1462, opened 2025-07-08
defaults: append -cluster and -enterprise image suffixes when it's needed. 

AndrewChubatiuk:
- removed `-cluster` and `-enterprise` suffixes from default versions and append them in defaults when it's needed
- removed impact of vlcluster cluster version on lb image, since it's versioning differs

related issue #1365

### DIFF
--- a/internal/controller/operator/factory/build/defaults.go
+++ b/internal/controller/operator/factory/build/defaults.go
@@ -1,6 +1,7 @@
 package build
 
 import (
+	"github.com/Masterminds/semver/v3"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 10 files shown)


internal/controller/operator/factory/build/defaults.go
```diff
@@ -372,17 +399,27 @@ func addVMClusterDefaults(objI any) {
 				}
 			}(c.VMBackup.Resource),
 		}
+		version := c.VMBackup.Version
+		if cr.Spec.ClusterVersion != "" {
```
f41gh7, line 403: I don't like the idea to implicitly change value of user-input. It's fine to changes operator defaults.

 If there is `cr.Spec.ClusterVersion: v1.120.0`, it must be added as `v1.120.0` to the pod images without modifications. All implicit conversions of `v1.120.0` to `v1.120.0-cluster` or `v1.120.0-enterpise-cluster` are hard to debug.

cc @Haleygo

AndrewChubatiuk: IMO `ClusterTag` could be added without modification, but since it's `ClusterVersion` conversion to respective tag is expected

f41gh7:
Thanks for initial implementation! I've created a commit based on it https://github.com/VictoriaMetrics/operator/commit/0d91e181114da80a6e44f3ba6f6f6235a1268152

 It doesn't change default env variable configuration values and checks for exist `-enterprise` suffix ( see tests section).

 I've added comment with example of it to the linked issue https://github.com/VictoriaMetrics/operator/issues/1365#issuecomment-3050097540

PR #1470, opened 2025-07-16
internal/controller/operator/factory/vmagent: fix tls config marshalling

zekker6:
A config marshalling will not add `tls_config` to store the config for scrape configurations at vmagent since 8df334cc.


f41gh7:
The bugfix was included into [v0.61.1](https://github.com/VictoriaMetrics/operator/releases/tag/v0.61.1) release

PR #1472, opened 2025-07-17
introduce VMAnomalyConfig

AndrewChubatiuk:
starting 1.25.0 anomaly supports hot-reload, and this PR introduces VMAnomalyConfig CR, that allows to add extra `models`, `schedulers` and `queries` for VMAnomaly dynamic configuration


f41gh7 (dismissed):
It's required to add documentation in first place.  Is it possible that model could interfere on each other? Is it required to config VMAnomaly itself somehow? What is a relations between VMAnomaly and VMAnomalyModel? 

Currently it's not possible to use this new API.

 Also, I suggest to add `-configCheckInterval` to the vmanomaly in the same way as other components have it. It solves configuration reload without need of external config reloaders.

AndrewChubatiuk:
removed config reloader since python watchdog, which implements hot reload listens to fs events

AndrewChubatiuk:
added scheduler CR, added docs

api/operator/v1/vmanomaly_types.go
vrutkovs, line 1: Do we want to run it through `v1beta1` for a few releases first?

docs/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ aliases:
 **Update note 2**: removed `operator_prometheus_converter_watch_events_total` metric since migration of Prometheus object watchers to controllers made this counter obsolete.
 **Update note 3**: made `controller.prometheusCRD.resyncPeriod` command line flag noop, which was relevant to Prometheus object watchers.
 **Update note 4**: `-eula` flag is not set by default anymore for VMBackup and VMRestore. To avoid VMCluster/VMSingle rollouts set `spec.vmstorage.vmBackup.acceptEula: true` for VMCluster and `spec.vmBackup.acceptEula: true"` for VMSingle and replace it with `spec.license` during VMSingle/VMCluster upgrade.
+**Update note 5**: due to introduction of new flags and sidecar VMAnomaly will be recreated.
```
```suggestion
**Update note 5**: new flags and config-reloader sidecar was introduced, so VMAnomaly pods will be recreated.
```

internal/controller/operator/factory/vmanomaly/config/config.go
```diff
@@ -1,17 +1,101 @@
...
+	var nsn []string
+	opts := &k8stools.SelectorOpts{
+		DefaultNamespace:  cr.Namespace,
+		SelectAll:         cr.Spec.SelectAllByDefault,
+		ObjectSelector:    cr.Spec.ConfigSelector,
+		NamespaceSelector: cr.Spec.ConfigNamespaceSelector,
+	}
+	if err := k8stools.VisitSelected(ctx, rclient, opts, func(list *vmv1.VMAnomalyConfigList) {
+		for i := range list.Items {
+			item := &list.Items[i]
```
vrutkovs, line 40:
```suggestion
		for _, item := range list.Items {
```
Since we're not mutating `item` here

internal/controller/operator/factory/vmanomaly/config/config.go
```diff
@@ -1,17 +1,101 @@
...
+	case cr.Spec.ConfigSecret != nil:
+		secret, err := ac.LoadKeyFromSecret(cr.Namespace, cr.Spec.ConfigSecret)
+		if err != nil {
+			return nil, fmt.Errorf("cannot fetch secret content for anomaly config secret, name=%q: %w", cr.Name, err)
+		}
+		data = []byte(secret)
+	case cr.Spec.ConfigRawYaml != "":
+		data = []byte(cr.Spec.ConfigRawYaml)
+	default:
+		return nil, fmt.Errorf(`either "configRawYaml" or "configSecret" are required`)
```
vrutkovs, line 71: We should do this check via validatingwebhook. Also, I'd expect raw yaml to take precedence when both are specified?

internal/controller/operator/factory/vmanomaly/pod.go
```diff
@@ -17,16 +18,25 @@ import (
...
+	configEnvsubstFilename = "vmanomaly.env.yaml"
 )
 
+func reloadSupported(cr *vmv1.VMAnomaly) bool {
+	anomalyVersion, err := semver.NewVersion(cr.Spec.Image.Tag)
+	if err == nil {
+		return anomalyVersion.GreaterThanEqual(reloadMinVersion)
+	}
+	return false
+}
```
vrutkovs:
```suggestion
func reloadSupported(cr *vmv1.VMAnomaly) bool {
	anomalyVersion, err := semver.NewVersion(cr.Spec.Image.Tag)
	if err != nil {
		return false
	}
	return anomalyVersion.GreaterThanEqual(reloadMinVersion)
}
```
A bit more go idiomatic.

internal/controller/operator/factory/vmanomaly/pod.go
```diff
@@ -17,16 +18,25 @@ import (
 	"github.com/VictoriaMetrics/operator/internal/controller/operator/factory/k8stools"
 )
 
+var reloadMinVersion = semver.MustParse("v1.25.0")
```
vrutkovs: Is this the version of VMAnomaly which supports current config-reloader configuration? Perhaps the operator could support previous versions too

internal/controller/operator/factory/vmanomaly/pod.go
```diff
@@ -17,16 +18,25 @@ import (
...
+	anomalyDir             = "/etc/vmanomaly"
+	confDir                = anomalyDir + "/config"
+	tlsAssetsDir           = anomalyDir + "/tls"
+	storageDir             = "/storage"
+	configVolumeName       = "config"
+	configEnvsubstFilename = "vmanomaly.env.yaml"
 )
 
+func reloadSupported(cr *vmv1.VMAnomaly) bool {
+	anomalyVersion, err := semver.NewVersion(cr.Spec.Image.Tag)
```
vrutkovs: This might break if customers use custom tags / RCs  - we should add a CHANGELOG entry for this

internal/controller/operator/vmanomalyconfig_controller.go
```diff
@@ -0,0 +1,136 @@
...
+				l.Error(err, "cannot match vmanomaly and vmanomalyconfig")
+				continue
+			}
+			if !match {
+				continue
+			}
+		}
+
+		if err := vmanomaly.CreateOrUpdateConfig(ctx, r, item, &instance); err != nil {
+			l.Error(err, "failed to update vmanomaly config")
```
vrutkovs, line 119:
```suggestion
			l.Error(err, "failed to update vmanomaly config")
			return
```

internal/controller/operator/vmanomalyconfig_controller_test.go
```diff
@@ -0,0 +1,83 @@
...
+		BeforeEach(func() {
+			By("creating the custom resource for the Kind VMAnomalyConfig")
+			err := k8sClient.Get(ctx, typeNamespacedName, vmanomalyconfig)
+			if err != nil && k8serrors.IsNotFound(err) {
+				resource := &vmv1.VMAnomalyConfig{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      resourceName,
+						Namespace: "default",
+					},
+					// TODO(user): Specify other spec details if needed.
```
vrutkovs, line 53: :)

internal/controller/operator/vmanomalyconfig_controller_test.go
```diff
@@ -0,0 +1,83 @@
...
+			controllerReconciler := &VMAnomalyConfigReconciler{
+				Client:       k8sClient,
+				OriginScheme: k8sClient.Scheme(),
+			}
+
+			_, err := controllerReconciler.Reconcile(ctx, reconcile.Request{
+				NamespacedName: typeNamespacedName,
+			})
+			Expect(err).NotTo(HaveOccurred())
+			// TODO(user): Add more specific assertions depending on your controller's reconciliation logic.
```
vrutkovs, line 79: We can extend the e2e tests later

test/e2e/upgrade/upgrade_test.go
vrutkovs:
```suggestion
			name: "VMAlert/VMAuth/VMAlertmanager",
```

vrutkovs (changes requested):

internal/controller/operator/vmanomalyconfig_controller_test.go
AndrewChubatiuk: this is autogenerated file

internal/controller/operator/vmanomalyconfig_controller.go
AndrewChubatiuk: problem with this change is that error can be caused by other vmanomalyconfig

docs/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ aliases:
 **Update note 2**: removed `operator_prometheus_converter_watch_events_total` metric since migration of Prometheus object watchers to controllers made this counter obsolete.
 **Update note 3**: made `controller.prometheusCRD.resyncPeriod` command line flag noop, which was relevant to Prometheus object watchers.
 **Update note 4**: `-eula` flag is not set by default anymore for VMBackup and VMRestore. To avoid VMCluster/VMSingle rollouts set `spec.vmstorage.vmBackup.acceptEula: true` for VMCluster and `spec.vmBackup.acceptEula: true"` for VMSingle and replace it with `spec.license` during VMSingle/VMCluster upgrade.
+**Update note 5**: new flags and config-reloader sidecar was introduced, so VMAnomaly pods will be recreated.
```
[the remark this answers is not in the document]
AndrewChubatiuk: actually PR introduces no config-reloader, using --watch flag instead, Nik was against it

internal/controller/operator/factory/vmanomaly/config/config.go
AndrewChubatiuk: vmalertmanager has the same properties and priorities are the same

internal/controller/operator/factory/vmanomaly/config/config.go
AndrewChubatiuk: we're mutating them later in ForEach loops

internal/controller/operator/factory/vmanomaly/pod.go
AndrewChubatiuk: rcs should be covered by this check, but custom builds - not, haven't seen Fred shares builds with nonstandard versions

internal/controller/operator/factory/vmanomaly/pod.go
AndrewChubatiuk: version, since which operator supports watch flag, there's no reloader in this PR, it means in case of configuration changes old versions will be restarted. Now i'm thinking either add config reloader or disallow pre 1.25.0 versions with operator at all

vrutkovs: oh gotcha, I was confused by sidecar mention previously. I think config reloader is good option (even in the latest versions too as it separates config management from the app)

internal/controller/operator/factory/vmanomaly/pod.go
vrutkovs: Okay, lets mention it in the changelog - there is a chance the customers may mirror the images in their org and having this mentioned may save us some time on issue troubleshooting

vrutkovs (approved):

docs/resources/vmanomaly.md
```diff
@@ -326,6 +326,39 @@ spec:
         z_threshold: 2.5
 ```
 
+## Dynamic configuration
+
+`VMAnomaly` supports discovering additional models, schedulers, and queries using [VMAnomalyConfig](https://docs.victoriametrics.com/operator/resources/vmanomalyconfig/).
+
+For filtering configuration objects, `VMAnomaly` uses `configNamespaceSelector` and `configSelector` selectors.
```
vrutkovs:
```suggestion
For filtering configuration objects, `VMAnomaly` uses `configNamespaceSelector` and `configSelector` selectors. This enables users to associate multiple VMAnomalyConfigs with multiple VMAnomaly instances (a many-to-many relationship).
```

Issue #1468, opened 2025-07-15
Refactoring broke selector discovery logic


f41gh7:
Bug introduced at commit https://github.com/VictoriaMetrics/operator/commit/a4aeda3fb7f64df370c5715d4d128b6824924c00

 Due to change from switch case

```
	switch {
	case len(watchNS) > 0:
		// perform match only for watched namespaces
		// filters by namespace is disabled, since operator cannot access cluster-wide APIs
		// this case could be improved to additionally filter by namespace name - metadata.name label
		namespaces = append(namespaces, watchNS...)
	case objectSelector != nil && nsSelector == nil:
		// in single namespace mode, return object ns
		namespaces = append(namespaces, objNamespace)
	case nsSelector != nil:
		// perform a cluster wide request for namespaces with given filters
		nsSelector, err := metav1.LabelSelectorAsSelector(nsSelector)
		if err != nil {
			return fmt.Errorf("cannot convert selector: %w", err)
		}
		namespaces, err = SelectNamespaces(ctx, rclient, nsSelector)
		if err != nil {
			return fmt.Errorf("cannot select namespaces for  match: %w", err)
		}
		// if nsSelector is specified and no match, return directly
		if len(namespaces) == 0 {
			return nil
		}
	}
```

 to else if:
```
	if s.NamespaceSelector == nil && s.ObjectSelector != nil {
		// in single namespace mode, return object ns
		if len(namespaces) == 0 || slices.Contains(namespaces, s.DefaultNamespace) {
			namespaces = append(namespaces[:0], s.DefaultNamespace)
		}
	} else if len(namespaces) == 0 {
		// perform a cluster wide request for namespaces with given filters
		// filters by namespace is disabled, when WATCH_NAMESPACES is defined,
		// since operator cannot access cluster-wide APIs this case could be improved
		// to additionally filter by namespace name - metadata.name label
		opts := &client.ListOptions{}
		if s.NamespaceSelector != nil {
			selector, err := metav1.LabelSelectorAsSelector(s.NamespaceSelector)
			if err != nil {
				return nil, fmt.Errorf("cannot convert selector: %w", err)
			}
			opts.LabelSelector = selector
		}
		l := &corev1.NamespaceList{}
		if err := rclient.List(ctx, l, opts); err != nil {
			return nil, fmt.Errorf("cannot select namespaces for  match: %w", err)
		}
		for _, n := range l.Items {
			namespaces = append(namespaces, n.Name)
		}
	}
```

[...]

PR #1474, opened 2025-07-17
k8stools/selectors: properly apply namespace selectors

f41gh7:
Previously, operator performed namespace discovery on defined, but
empty namespaceSelector. Also operator treat empty response for namespaceSelector ( such as non-namespace matched for given selectors) as cluster wide request. Which is not correct.

 This commit fixes both issues and adds corresponding tests for
selectors.

Fixes https://github.com/VictoriaMetrics/operator/issues/1468


AndrewChubatiuk (approved):

AndrewChubatiuk:
Although it's not obvious why nil and empty selectors should work differently

f41gh7:
> Although it's not obvious why nil and empty selectors should work differently

 It's a dark side of golang type system. Slice is not commonly checked for nil . I'll think how to better handle it.

AndrewChubatiuk:
could you also rewrite affected tests to f-tests?

Issue #1478, opened 2025-07-21
VMagent connot start


QuentinBtd:
Hello,

Since Operator 0.61.0, VMAgent pod stop after startup

```log
│ vmagent 2025-07-21T11:57:51.332Z    info    VictoriaMetrics/app/vmagent/main.go:171    started vmagent in 0.008 seconds                                                                                                                                                                                                                                                  │
│ vmagent 2025-07-21T11:57:51.332Z    info    VictoriaMetrics/lib/promscrape/scraper.go:118    reading scrape configs from "/etc/vmagent/config_out/vmagent.env.yaml"                                                                                                                                                                                                      │
│ vmagent 2025-07-21T11:57:51.334Z    info    VictoriaMetrics/lib/httpserver/httpserver.go:145    started server at http://0.0.0.0:8429/                                                                                                                                                                                                                                   │
│ vmagent 2025-07-21T11:57:51.334Z    info    VictoriaMetrics/lib/httpserver/httpserver.go:147    pprof handlers are exposed at http://0.0.0.0:8429/debug/pprof/                                                                                                                                                                                                           │
│ vmagent 2025-07-21T11:57:51.402Z    info    VictoriaMetrics/lib/promscrape/config.go:149    starting service discovery routines...                                                                                                                                                                                                                                       │
│ vmagent 2025-07-21T11:57:51.404Z    info    VictoriaMetrics/lib/promscrape/config.go:155    started 173 service discovery routines in 0.002 seconds                                                                                                                                                                                                                      │
│ vmagent 2025-07-21T11:57:51.408Z    info    VictoriaMetrics/lib/promscrape/scraper.go:439    static_configs: added targets: 68, removed targets: 0; total targets: 68                                                                                                                                                                                                    │
│ config-reloader ts=2025-07-21T11:57:51.233132685Z level=info caller=/workspace/cmd/prometheus-config-reloader/main.go:148 msg="Starting prometheus-config-reloader" version="(version=0.84.0, branch=, revision=8337f85)" build_context="(go=go1.24.5, platform=linux/arm64, user=, date=20250714-10:10:06, tags=unknown)"                                               │
│ config-reloader ts=2025-07-21T11:57:51.233519004Z level=info caller=/workspace/internal/goruntime/cpu.go:27 msg="Updating GOMAXPROCS=1: using minimum allowed GOMAXPROCS"                                                                                                                                                                                                │
[...]

PR #1481, opened 2025-07-22
config-reloader: fixed command lint arguments override

AndrewChubatiuk:
related issue https://github.com/VictoriaMetrics/operator/issues/1478

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@ aliases:
 ## tip
 
 * BUGFIX: [vmagent](https://docs.victoriametrics.com/operator/resources/vmagent/): properly add `attach_metadata` section to scrape configuration. See [#1476](https://github.com/VictoriaMetrics/operator/issues/1476).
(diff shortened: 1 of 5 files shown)


f41gh7 (changes requested):
It'd be great to have a tests to mitigate such issue in the futre

AndrewChubatiuk:
the only thing i can check here is that CR before and after a test is the same

f41gh7:
> the only thing i can check here is that CR before and after a test is the same

 I thought, there is an issue with incorrect assigned of new config-reloader args. If it's only refactoring and not an actual fix, it should marked as refactoring.

 It's safe to modify provided `CR` resource, since it deep-copied and allows mutations. 

 Please add more context to the commit message or issue.

AndrewChubatiuk:
`makeSpec` is called twice. During first invocation, which happens for assets cache, it drops overrides from ConfigReloaderExtraArgs and produces a valid output, which is ignored and during a second invocation it only adds additional arguments to config reloader with no overrides and this output is used for Deployment/StatefulSet spec generation

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

f41gh7:
> the only thing i can check here is that CR before and after a test is the same

 The proper test should check if `VMAgent.spec.configReloaderArgs` is set and it properly applied to the result `Deployment.spec`

Issue #1485, opened 2025-07-25
PrometheusRule conversion omits keep_firing_for from rule definitions


brokenjacobs:
When a PrometheusRule contains a keep_firing_for key, the resulting VMRule object created by the conversion process is missing the keep_firing_for in the rule body. No error is generated.

Making a VMRule with keep_firing_for included works fine though.

AndrewChubatiuk:
conversion of `keep_firing_for` PrometheusRule parameter will be available in the next release

AndrewChubatiuk:
fix for this issue was released in 0.62.0

PR #1487, opened 2025-07-26
converter: support more PrometheusRule properties conversion

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1485
added support given below PrometheusRule properties:
- spec.limit
- spec.query_offset -> spec.eval_offset
- spec.labels
- spec.group[*].keep_firing_for

added tests for PrometheusRule and rewrote all converted tests using f-tests pattern


f41gh7 (approved):
LGTM

PR #1489, opened 2025-07-26
vlsingle, vmsingle: do not mount emptyDir if `data` volume is present in `spec.volumes`

AndrewChubatiuk:
currently it's only possible to set external PVC for vlsingle and vmsingle only with crutches: storageDataPath should be set using `spec.extraArgs`, since default or one, which is controlled by `spec.storageDataPath` creates PVC from `spec.storage` or mounts emptyDir if `spec.storage` is not set. any volume mounts that are mounted to `spec.storageDataPath` lead to collision. Even with this crutch pod has unneeded volume mount, also in vmsingle backup functionality becomes unavailable, since it also relies on `spec.storageDataPath` value.

this PR moves method, which mounts storage volume to build package and does the following:
- adds volume mount that points to `storageDataPath` only if volume with expected name is not already present in `spec.volumeMounts`
- adds volume with managed PVC if `spec.storage` is defined
- adds emptyDir volume only storage volume is not found in `spec.volumes`


internal/controller/operator/factory/build/container_test.go
```diff
@@ -41,220 +41,206 @@ func (t testBuildProbeCR) ProbeNeedLiveness() bool {
 }
 
 func Test_buildProbe(t *testing.T) {
```
f41gh7, line 46: changes to the container_test.go file looks unrelated to this PR.

AndrewChubatiuk: method for volumes and mounts creation was added to container.go, added tests for it to container_test.go and rewrote other tests to f-tests

internal/controller/operator/factory/build/container_test.go
```diff
@@ -349,5 +348,180 @@ func TestAddSyslogArgsTo(t *testing.T) {
...
+					AWSElasticBlockStore: &corev1.AWSElasticBlockStoreVolumeSource{
+						VolumeID: "aws-volume",
+					},
+				},
+			},
+		},
+		expectedMounts: []corev1.VolumeMount{{
+			Name:      "test",
+			MountPath: "/test",
+		}},
```
vrutkovs, line 466: Which of those volumes will be mounted in the end - iiuc its AWS since its the last one?

test/e2e/vmsingle_test.go
```diff
@@ -245,8 +245,34 @@ var _ = Describe("test vmsingle Controller", Label("vm", "single"), func() {
...
+						Spec: vmv1beta1.VMSingleSpec{
+							CommonApplicationDeploymentParams: vmv1beta1.CommonApplicationDeploymentParams{
+								ReplicaCount: ptr.To[int32](1),
+							},
+							CommonDefaultableParams: vmv1beta1.CommonDefaultableParams{
+								UseStrictSecurity: ptr.To(false),
+							},
+							RetentionPeriod:      "1",
+							RemovePvcAfterDelete: true,
+							StorageDataPath:      "/tmp/",
```
vrutkovs, line 265: We should be checking that `StorageDataPath` is listed in one of the volumes

internal/controller/operator/factory/build/container_test.go
AndrewChubatiuk: as you can check removed logic in vmsingle, vtsingle and vlsingle it first added storage volume and then other extra volumes. This order was preserved to avoid unneeded pods recreation for users, who have explicitly defined image tag

test/e2e/vmsingle_test.go
AndrewChubatiuk: updated PR

vrutkovs: I was thinking about extending tests too:
```
Expect(ts.Volumes[0].EmptyDir).ToNot(BeNil())
Expect(ts.Containers[0].VolumeMounts[0].MountPath).To(Equal("/tmp/"))
```
and so on in other testcases too

f41gh7:
Lets be careful with this change. It potentially may break exist deployments of VMSingle instances, I'd like to keep it backward-compatible.

AndrewChubatiuk:
@f41gh7 @vrutkovs 
updated implementation and description

lexfrei:
Related to #1677

docs/CHANGELOG.md
```diff
@@ -25,6 +24,7 @@ SECURITY: upgrade Go builder from Go1.25.4 to Go1.25.5. See [the list of issues
...
 * BUGFIX: [vmanomaly](https://docs.victoriametrics.com/operator/resources/vmanomaly/): properly handle configuration which is missing `reader.queries` in either `configRawYaml` or `configSecret`. Previously, it would lead to panic.
 * BUGFIX: [vmanomaly](https://docs.victoriametrics.com/operator/resources/vmanomaly/): fix configuration parsing when running in [UI mode](https://docs.victoriametrics.com/anomaly-detection/ui/). Previously, configuration required to use `preset: ui:version` instead of `preset: ui`.
+* BUGFIX: [vmsingle](https://docs.victoriametrics.com/operator/resources/vmsingle/), [vlsingle](https://docs.victoriametrics.com/operator/resources/vlsingle/) and [vmalertmanager](https://docs.victoriametrics.com/operator/resources/vmalertmanager): do not mount emptydir if storage data volume is already present in volumes list. Before it was impossible to mount external PVC without overriding default storageDataPath using `spec.extraArgs` and without having unneeded emptydir listed among pod volumes. Related issues [#1477](https://github.com/VictoriaMetrics/operator/issues/1477).
```
```suggestion
* BUGFIX: [vmsingle](https://docs.victoriametrics.com/operator/resources/vmsingle/), [vlsingle](https://docs.victoriametrics.com/operator/resources/vlsingle/), [vtsingle](https://docs.victoriametrics.com/operator/resources/vtsingle/) and [vmalertmanager](https://docs.victoriametrics.com/operator/resources/vmalertmanager): do not mount emptydir if storage data volume is already present in volumes list. Before it was impossible to mount external PVC without overriding default storageDataPath using `spec.extraArgs` and without having unneeded emptydir listed among pod volumes. Related issues [#1477](https://github.com/VictoriaMetrics/operator/issues/1477).
```

AndrewChubatiuk: updated

PR #1504, opened 2025-08-05
Zap-log-level command line flag docs update for missing warn level

possibull:
Updates docs to highlight reason for missing warn level to zap-log-level command line flag

### DIFF
--- a/docs/flags.md
+++ b/docs/flags.md
@@ -76,6 +76,8 @@ Usage of bin/operator:
     	Zap log encoding (one of 'json' or 'console')
   -zap-log-level value
     	Zap Level to configure the verbosity of logging. Can be one of 'debug', 'info', 'error', 'panic'or any integer value > 0 which corresponds to custom debug levels of increasing verbosity
+        Note: warn is missing by design due to warn level not being supported by `controller-runtime`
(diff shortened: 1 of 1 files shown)


makasim:
@possibull The `docs/flags.md` file is auto-generated when running `make docs` see (Makefile](https://github.com/VictoriaMetrics/operator/blob/00a56de298560fdc0895523a3a86610c76288768/Makefile#L120)).

To update it properly, modify the flag description in the source code, then run make docs and commit

possibull:
> @possibull The `docs/flags.md` file is auto-generated when running `make docs` see (Makefile](
> 
> https://github.com/VictoriaMetrics/operator/blob/00a56de298560fdc0895523a3a86610c76288768/Makefile#L120
> 
> )).
> To update it properly, modify the flag description in the source code, then run make docs and commit

Thanks @makasim appreciate you taking the time to explain this.  Will update.

possibull:
The challenge here is the `zap-log-level` flag description is coming from zap itself.

makasim:
I see, then I think the easiest way would be to add this info somewhere in this section https://docs.victoriametrics.com/operator/configuration/#flags

possibull:
I think I could append to `f.Usage` before it's displayed to user

possibull:
Working on why value is set to `controller runtime`. Will update load fix - and reduce the nested ifs

possibull:
This should be ready for review.  Thanks for the helpful comments.

PR #1510, opened 2025-08-15
controller/operator/factory: run PollUntilContextTimeout with immediate flag

vrutkovs:
When `immediate` flag is unset, this function waits for `interval` first and only afterward runs the function the first time. This commit changes all function calls to use `immediate` true to make it run sooner.

With [tracing](https://github.com/vrutkovs/operator/tree/tracing) the effects of these flags is most obvious.
Before:
image
`reconcile.waitDeploymentReady` doesn't do anything for a second and then runs `client.Get`

After:
image
Now `reconcile.waitDeploymentReady` takes 25 nanoseconds and runs `client.Get` immediately

Unfortunately I didn't get a chance to test this PR properly, as it affects a lot of areas in operators, but I can't think of a reason why would it cause regressions.


f41gh7:
Thanks for contribution!

f41gh7:
Feature was included into [v0.62.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.62.0) release

PR #1512, opened 2025-08-18
internal/config: update VictoriaMetrics components to v1.124.0

zekker6:


f41gh7:
Thanks for contribution!

Issue #379, opened 2021-11-13
VMUser `targetRefs.crd[]` uses path wildcard, preventing "single-tenant view" (usage of `VMUser/vminsert` and `VMUser/vmselect`)


flokli:
I created the following `VMUser` resource:

```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMUser
metadata:
  name: foo-vmalert
  namespace: default
spec:
  generatePassword: true
  targetRefs:
  - crd:
      kind: VMCluster/vminsert
      name: vmcluster
      namespace: default
    target_path_suffix: /insert/30
  - crd:
      kind: VMCluster/vmselect
      name: vmcluster
      namespace: default
    target_path_suffix: /select/30
  username: foo-vmalert
```

The idea was that this will look like a single-tenant vminsert and vmselect endpoint.

However, the rendered config looks like this:

```yaml
- url_map:
  - url_prefix: http://vminsert-vmcluster.default.svc:8480/insert/30
    src_paths:
    - /.*
  - url_prefix: http://vmselect-vmcluster.default.svc:8481/select/30
    src_paths:
    - /.*
  username: foo-vmalert
  password: XXX
```

… and vmselect queries mistakenly end up in the vminsert endpoint, and vice-versa.

Is there a way vminsert/vmselect could configure more specific `src_paths` only, those that are configured in `vminsert`/`vmselect`?

f41gh7:
I think, in this case vmselect must serve all unmatched paths - `/.*` and vminsert only specific urls for data ingestion. It changes less frequent and doesn't have any complicated paths. 

something like: 

```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMUser
metadata:
  name: foo-vmalert
  namespace: default
spec:
  generatePassword: true
  targetRefs:
  - crd:
      kind: VMCluster/vminsert
      name: vmcluster
      namespace: default
    target_path_suffix: /insert/30
    paths:
     - /prometheus/api/v1/write
     - /prometheus/api/v1/import
     - /prometheus/api/v1/import/csv
     - /prometheus/api/v1/import/prometheus
     - /prometheus/api/v1/import/native
     - /influx/.*
     - /datadog/.*
  - crd:
      kind: VMCluster/vmselect
      name: vmcluster
      namespace: default
    target_path_suffix: /select/30
  username: foo-vmalert
```

[...]

PR #1519, opened 2025-09-03
vmuser: fix default paths when using VMCluster/vminsert and VMCluster…

Haleygo:
…/vmselect crds as target


f41gh7 (approved):
LGTM

f41gh7:
I added compatibility with https://github.com/VictoriaMetrics/operator/issues/379

f41gh7 (approved):
LGTM

f41gh7:
I think, it's not what is expected to simply convert `/insert/15` into match `/insert/.*`, since it could be a single tenant access case.

f41gh7:
Thanks for contribution!

f41gh7:
The feature was included into [v0.63.0](https://docs.victoriametrics.com/operator/changelog/#v0630) release

PR #1522, opened 2025-09-10
reconcile/service: properly handle loadbalancerClass field change

f41gh7:
Previously, operator produced runtime error:

```
cannot reconcile additional service for CRD: Service "CRD-NAME" is invalid: spec.loadBalancerClass: Invalid value: "null": may not change once set
```

 This commit adds `Service` re-creation on field change. It's expected
that `Service` `IP` will be changed. And it may interrupt communication with application.


f41gh7:
The bugfix was included into [v0.63.0](https://docs.victoriametrics.com/operator/changelog/#v0630) release

Issue #1480, opened 2025-07-21
VMOperator: Make top-level routing configurable instead of "namespace" label in VMAlertManager


robincw-gr:
### Context
VMOperator merges routes in user-provided VMAlertManagerConfig into the routing tree in a shared VMAlertManager.
When a new top-level route for a new VMAlertManagerConfig is built, the `namespace` label of VMAlertManagerConfig is used to make a label matcher for incoming alerts. This requires that alerts have a `namespace` label that refers to the namespace of a VMAlertManagerConfig that contains desired routes.
- There is an option to disable this dynamic routing rule: `disableNamespaceMatcher`
- There is an option to add enforced matchers that are _static_ at time of VMAlertMnager creation: `enforcedTopRouteMatchers`
- There are no options to customise the dynamic matcher.

### Problem
The `namespace` label is [hard-coded](https://github.com/VictoriaMetrics/operator/blob/master/internal/controller/operator/factory/vmalertmanager/config.go#L233) in VMOperator.
This works if the user-owned alert rules set the `namespace` label to the namespace of a VMAlertManagerConfig that contains desired routes.
**But**, the `namespace` label name is (ironically) not namespaced, and is widely used in other contexts (like Kubernetes), therefore it is ambiguous and overloaded. It can be set by users and often appears in source metrics (used in alert rules) to mean other types of namespaces in the infrastructure or business domains. This means alert routing does not work, or the `namespace` label from source telemetry has to be overwritten.

### Proposal
The `namespace` label in this context _should_ mean "the namespace of the VMAlertManagerConfig to which the alert should be routed", not "the namespace provided by the alert expression". We need to avoid conflicts with labels from source metrics.
**~Option 1~:** A better label name would be something like "alertmanagerconfig_namespace"... but backwards compatibility is needed.
**Option 2:** Making it configurable in the operator would be much better, with the default being `namespace` for compatibility.
**Option 3:** Allow an override for the label name in the VMAlertManagerConfig. For example, a field called `topRouteNamespaceLabel` could be provided, with the default being `namespace` for compatibility.

f41gh7:
I'd like an idea to make it configurable at `VMAlertmanger` level. The problem with `VMAlertManagerConfig` is security. The main purpose of `VMAlertManagerConfig` sub-resource is to provide an API for building a SaaS-like monitoring solution on top of Kubernetes cluster. And it's required to have isolation at namespace level. 

[...]

PR #1525, opened 2025-09-11
vmalertmanager: add `customizedNamespaceKey` field to allow changing …

Haleygo:
…the top route namespace matcher key for VMAlertmanagerConfig

fix https://github.com/VictoriaMetrics/operator/issues/1480


api/operator/v1beta1/vmalert_types.go
```diff
@@ -366,11 +366,17 @@ func (cr *VMAlert) Validate() error {
 		if cr.Spec.Notifier.URL == "" && cr.Spec.Notifier.Selector == nil {
 			return fmt.Errorf("spec.notifier.url and spec.notifier.selector cannot be empty at the same time, provide at least one setting")
 		}
+		if cr.Spec.NotifierConfigRef != nil {
```
f41gh7: Could you please extract this change into separate PR?

api/operator/v1beta1/vmalertmanager_types.go
```diff
@@ -163,11 +163,16 @@ type VMAlertmanagerSpec struct {
...
-	// DisableNamespaceMatcher disables top route namespace label matcher for VMAlertmanagerConfig
+	// DisableNamespaceMatcher disables adding top route label matcher "namespace = <VMAlertmanagerConfig.namespace>" for VMAlertmanagerConfig
 	// It may be useful if alert doesn't have namespace label for some reason
 	// +optional
 	DisableNamespaceMatcher bool `json:"disableNamespaceMatcher,omitempty"`
 
+	// CustomizeNamespaceMatcherKey defines the namespace label key for top route matcher for VMAlertmanagerConfig
+	// Default is "namespace"
+	// +optional
+	CustomizeNamespaceMatcherKey string `json:"customizedNamespaceKey,omitempty"`
```
f41gh7: WDYT of `EnforcedNamespaceLabel` name instead of `CustomizeNamespaceMatcherKey`?

 It'll make configuration consistent:

```
	// EnforcedNamespaceLabel enforces adding a namespace label of origin for each alert
	// and metric that is user created. The label value will always be the namespace of the object that is
	// being created.
```

api/operator/v1beta1/vmalert_types.go
Haleygo: Done, please see https://github.com/VictoriaMetrics/operator/pull/1526

f41gh7:
Thanks for contribution!

Issue #1511, opened 2025-08-15
vmalert fails to start if notifierConfigRef and notifiers.selector both set


robincw-gr:
### Describe the bug

Using the CRD to set both `notifierConfigRef` and `notifiers.selector` results in vmalert unable to start.
The notifierConfigRef in the VMAlert CRD provides vmalert with a `-notifier.config` flag to vmalert, set to the file from the specified secret which can contain many useful options. But vmalert does not start if that is provided as a flag in addition to `-notifier.url`, which the operator sets if `notifiers.selector` in the CRD is set for k8s discovery of Alertmanager. 

### To Reproduce

Example:
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAlert
metadata:
  name: test
  namespace: test
spec:

  # sets a -notifier.url flag for each alertmanager found:
  notifiers:
    - selector:
        labelSelector:
          matchLabels:
            notifier: test

  # sets the -notifier.config flag:
  notifierConfigRef:
    name: test-notifierconfig-secret
```

### Version

`vmselect-20241129-171755-tags-v1.107.0-cluster-0-g27e11b6f11`
I don't have access to the vmalert binary right now

### Logs

`only one of -notifier.config or -notifier.url flags must be specified`

### Additional information

The notifier config file has many useful options that are not supported in the CRD. Improving the CRD should be considered.

Haleygo:
Hi,
The error message is expected on vmalert side, the [VMAlert CR validation check](https://github.com/VictoriaMetrics/operator/blob/346398caa8c620e138b4d9e59c3b2fcdd809bf5a/api/operator/v1beta1/vmalert_types.go#L355) needs to be improved, see https://github.com/VictoriaMetrics/operator/issues/1480#issuecomment-3195061508.

>The notifier config file has many useful options that are not supported in the CRD. Improving the CRD should be considered.

Users should be able to place the entire notifier config file inside the notifierConfigRef; which option is not currently supported?

I'm transferring this issue to the operator repo as an enhancement for the VMAlert CRD, since no VMAlert-side fixes are required.

f41gh7:
The bugfix was included into [v0.63.0](https://docs.victoriametrics.com/operator/changelog/#v0630) release

PR #1526, opened 2025-09-12
vmalert: add mutually exclusive checks for notifier configs in admiss…

Haleygo:
…ion webhook validation

fix https://github.com/VictoriaMetrics/operator/issues/1511


f41gh7:
Thanks for contribution!

PR #1529, opened 2025-09-15
generate specless crd

AndrewChubatiuk:
- generate specless crds for smaller helm release secret size
- generate compressed crds for upgrade job


f41gh7 (approved):
LGTM

PR #1532, opened 2025-09-19
Fix: Errors don't get logged for a number of controllers

rrhubenov:
# Issue
Due to returning empty result & nil error at the end of each reconciliation loop, the values for the `result` and `err` variables is `ctrl.Result{}` and `nil` respectively when calling the deferred function. This leads to the Reconciliation loop always returning those values.

# How to reproduce 
Anything that would cause an error to occur during the reconciliation logic will lead to this issue.
In my case, a CRD that was necessary to be installed for the `VLSingle` controller to work properly, was not installed. This lead to a silent failure (no logs).

I'm not the most experienced Go dev, so the combination of named returns + deferred function was a bit jarring to see. I understand the benefits, but the problem was not really obvious at a glance due to this.


f41gh7 (approved):
LGTM

f41gh7:
Could you please mention bugfix at changelog?

rrhubenov:
Thank you for the quick reviews, @f41gh7 I've added the bugfix to the changelog.

f41gh7:
Thanks for contribution!

Issue #1538, opened 2025-09-28
vmanomaly: allow "empty" configs for ui preset


fred-navruzov:
### Description

As a part of forthcoming release of vmanomaly GUI, we updated our presets with `ui` mode, to enfore all no-op components except for server that handles GUI requests on vmanomaly configuration backtesting, to reduce resource utilization and context switching.
The main purpose is to spawn vmanomaly is "ui" mode and share it with internal users to test and validate their configurations before they go to production.

UI preset would require from users as little (valid) config as 
```yaml
preset: 'ui'   # or preset: 'ui:latest'

# and *optional* settings level, like n_workers, restore_state or logger_levels
# settings:
#   restore_state: true
#   n_workers: 4
```
to be internally merged with preset-defined part which implements no-op components
e.g. 
```yaml
schedulers:
  noop:
    class: noop
reader:
  class: noop
writer:
  class: noop
models:
  placeholder:
    class: 'zscore'  # actually does nothing and is not used in production-based schedule (other components are noop)
```
to start a vmanomaly in "ui server mode"

Currently, there is an enforcement to have non-empty sections for schedulers/reader/models ... but they can be skipped in preset mode (as the main config validation happens after all the parts - user-driven and preset-premade - are united into a single config that (already) has all the required sections present (with noop components) ), so operator falls with null pointer exception (based on Zakhar's tests, also adding him as a participant)

@AndrewChubatiuk , can you please have a look? Image 'v1.26.0-alpha4' already has everything to test the potential fix

P.s. @zekker6 , JFYI, can you provide more details of what you faced and how you (temporarily) fixed that issue?

AndrewChubatiuk:
@fred-navruzov 
with fix on operator side suggested config above doesn't work, since each model requires a list of schedulers.
updated it to 

```
schedulers:
  noop:
    class: noop
reader:
  class: noop
writer:
  class: noop
models:
  placeholder:
    schedulers: ['noop']
    class: 'zscore'
```
but this doesn't work still, since [this condition](https://github.com/VictoriaMetrics/vmanomaly/blob/4b537757457093ef53e455f95f5d8149755fa9c7/src/config/decomposition.py#L482) is not passing, could you please update it to `len(entiries) > 0 and len(entities) <= shard`?

f41gh7:
The feature was included into [v0.64.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.64.0) release

PR #1539, opened 2025-09-29
vmanomaly: support ui preset mode

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1538
allow providing empty reader, writer, models and schedulers sections, when spec.configRawYaml.preset is equal to ui

### DIFF
--- a/internal/controller/operator/factory/vmanomaly/config/config.go
+++ b/internal/controller/operator/factory/vmanomaly/config/config.go
@@ -29,6 +29,7 @@ type config struct {
 	Monitoring *monitoring           `yaml:"monitoring,omitempty"`
 	Preset     string                `yaml:"preset,omitempty"`
 	Settings   *settings             `yaml:"settings,omitempty"`
+	Server     *server               `yaml:"server,omitempty"`
 }
 
 type settings struct {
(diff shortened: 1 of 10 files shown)


internal/controller/operator/factory/vmanomaly/config/readers.go
```diff
@@ -22,7 +24,10 @@ type reader struct {
 }
 
 func (r *reader) validate() error {
-	if r.Class != "reader.vm.VmReader" && r.Class != "vm" && r.Class != "reader.synthetic.SyntheticVmReader" && r.Class != "synthetic_vm" {
+	if strings.ToLower(r.Class) == "noop" {
+		return nil
+	}
+	if !slices.Contains([]string{"reader.vm.VmReader", "vm", "reader.synthetic.SyntheticVmReader", "synthetic_vm"}, r.Class) {
```
fred-navruzov: @AndrewChubatiuk , can you please also include (as a part of the same PR) support of new reader type (alias "vlogs", full class name "reader.vlogs.VLogsReader") which is also a part of v1.26.0?
Otherwise, I believe it will raise an error "reader class is not supported"?)

internal/controller/operator/factory/vmanomaly/config/config.go
```diff
@@ -38,6 +39,42 @@ type settings struct {
 }
 
 func (c *config) override(cr *vmv1.VMAnomaly, ac *build.AssetsCache) error {
+	if strings.ToLower(c.Preset) == "ui" {
```
fred-navruzov: can it be replaced with `strings.HasPrefix(strings.ToLower(c.Preset), "ui")` ? 
As long as valid ui preset names (we support preset versioning) are "ui", "ui:latest", "ui:1.0", "ui:2.0", etc.

internal/controller/operator/factory/vmanomaly/config/writers.go
```diff
@@ -14,7 +15,10 @@ type writer struct {
 }
 
 func (w *writer) validate() error {
-	if w.Class != "writer.vm.VmReader" && w.Class != "vm" {
+	if strings.ToLower(w.Class) == "noop" {
+		return nil
+	}
+	if !slices.Contains([]string{"writer.vm.VmReader", "vm"}, w.Class) {
```
fred-navruzov: probably a typo here, shouldn't it be "writer.vm.VmWriter" for writer section check?

PR #1542, opened 2025-09-30
Fix: conserve custom labels on services

### LINKED ISSUE #1533: Custom labels on services are deleted upon reconciling
Consider the following steps:

```
kubectl annotate svc vmselect-mycluster myannotation=myvalue
kubectl label svc vmselect-mycluster mylabel=myvalue
kubectl rollout restart deployment victoria-metrics-operator
```

this will trigger a reconcile of all objects managed by the VictoriaMetrics operator. I find that the custom label is deleted, but the annotation remains. Custom labels may be introduced by mutating webhooks or other controllers and dropping them is undesirable. In fact, I believe, this behavior is unintentional, as a [now removed comment](https://github.com/VictoriaMetrics/operator/blob/658983a542d943ecb162b3836236e77ee794e328/internal/controller/operator/factory/reconcile/service.go#L55) would seem to imply. It appears that the annotations are [correctly conserved](https://github.com/VictoriaMetrics/operator/blob/658983a542d943ecb162b3836236e77ee794e328/internal/controller/operator/factory/reconcile/service.go#L108), but the labels have been forgotten about. I think this is a bug and should be fixed.

lllamnyp:
When reconciling k8s services managed by the operator, custom labels, such as those added by mutating webhooks, the user, or third-party controllers are clobbered. Resolves #1533.

### DIFF
--- a/internal/controller/operator/factory/reconcile/service.go
+++ b/internal/controller/operator/factory/reconcile/service.go
@@ -114,6 +114,11 @@ func reconcileService(ctx context.Context, rclient client.Client, newService, pr
 		prevAnnotations = prevService.Annotations
 	}
 
+	var prevLabels map[string]string
+	if prevService != nil {
+		prevLabels = prevService.Labels
+	}
+
 	rclient.Scheme().Default(newService)
 	isEqual := equality.Semantic.DeepDerivative(newService.Spec, currentService.Spec)
 	if isEqual &&
@@ -125,6 +130,7 @@ func reconcileService(ctx context.Context, rclient client.Client, newService, pr
 
 	vmv1beta1.AddFinalizer(newService, currentService)
 	newService.Annotations = mergeAnnotations(currentService.Annotations, newService.Annotations, prevAnnotations)
+	newService.Labels = mergeAnnotations(currentService.Labels, newService.Labels, prevLabels)
 	cloneSignificantMetadata(newService, currentService)
 
 	logMsg := fmt.Sprintf("updating service %s configuration, is_current_equal=%v, is_prev_equal=%v, is_prev_nil=%v",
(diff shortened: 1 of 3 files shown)


lllamnyp:
JFYI, the test [here](https://github.com/VictoriaMetrics/operator/blob/e2f542bf2a032e958b6b6da20bdf475ec49e11c7/test/e2e/vmcluster_test.go#L1282) is now broken, but I can't figure out where the test data is being sourced from. I'd appreciate a pointer, so I could fix this as well.

f41gh7:
> JFYI, the test [here](https://github.com/VictoriaMetrics/operator/blob/e2f542bf2a032e958b6b6da20bdf475ec49e11c7/test/e2e/vmcluster_test.go#L1282) is now broken, but I can't figure out where the test data is being sourced from. I'd appreciate a pointer, so I could fix this as well.

Looks like this test spotted some discrepancy at `Service` builder function for vmauth-lb. It could be easily fixed by:

```diff
diff --git a/internal/controller/operator/factory/vmcluster/vmcluster.go b/internal/controller/operator/factory/vmcluster/vmcluster.go
index 71f0285e..dd99ccd4 100644
--- a/internal/controller/operator/factory/vmcluster/vmcluster.go
+++ b/internal/controller/operator/factory/vmcluster/vmcluster.go
@@ -1579,6 +1579,7 @@ func createOrUpdateVMAuthLBService(ctx context.Context, rclient client.Client, c
 	var prevSvc *corev1.Service
 	if prevCR != nil && prevCR.Spec.RequestsLoadBalancer.Enabled {
 		t.VMCluster = prevCR
+		t.finalLabels = prevCR.FinalLabels(lbls)
 		t.additionalService = prevCR.Spec.RequestsLoadBalancer.Spec.AdditionalServiceSpec
 		prevSvc = build.Service(t, prevCR.Spec.RequestsLoadBalancer.Spec.Port, nil)
 	}
```

lllamnyp:
@f41gh7 thanks, that fixed the tests. I've updated the PR.

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1544, opened 2025-10-07
Update VictoriaMetrics version to v1.127.0

rtm0:
Not sure why new file is added - I just ran `make docs` per release instructions.


AndrewChubatiuk:
New specless CRD was neither committed nor added to gitignore

PR #1545, opened 2025-10-07
update victoriatraces and anomaly versions

AndrewChubatiuk:
update traces and anomaly default versions


f41gh7:
Please add changelog entries

PR #1546, opened 2025-10-08
config: removed unused manifests

AndrewChubatiuk:
removed manifests, which are not used


f41gh7:
Thanks for contribution!

PR #1553, opened 2025-10-17
docs: fixed example in quick start guide

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/VictoriaMetrics/issues/9873

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -295,9 +295,6 @@ spec:
       containers:
         - name: main
           image: docker.io/victoriametrics/demo-app:1.2
-          ports:
-            - name: http
-              containerPort: 8080
 ---
 apiVersion: v1
 kind: Service


docs/quick-start.md
```diff
@@ -295,6 +295,9 @@ spec:
       containers:
         - name: main
           image: docker.io/victoriametrics/demo-app:1.2
+          ports:
```
makasim: It should not be required. `port: http` in service is enough

source https://github.com/VictoriaMetrics/debug-notes/blob/main/operator/quick-start/30-create-demo-app-manifest.sh#L40

docs/quick-start.md
```diff
@@ -295,6 +295,9 @@ spec:
...
+            - name: http
+              containerPort: 8080
```
AndrewChubatiuk:
```suggestion
```

makasim (approved):

Issue #1548, opened 2025-10-14
PDB issue with vmAgent


mdm183:
I am are trying to implement a Pod Disruption budget for vmAgent across multiple shards.

Example

I have 5 shards with 3x replicas each. I want a policy to ensure each shard can have max 1 pod unavailable in a budgeting scenario.

Using label selectors to achieve this in our deployment: 

```
podDisruptionBudget:
    maxUnavailable: 1
    selectorLabels:
      app.kubernetes.io/component: monitoring
      app.kubernetes.io/instance: vmagent
      app.kubernetes.io/name: vmagent
      managed-by: vm-operator
      shard-num: '%SHARD_NUM%'
```

Issue arises with the last line  `shard-num: '%SHARD_NUM%'`

The logs show this error

```
error":"cannot update pod disruption budget for vmagent: PodDisruptionBudget.policy \"vmagent-vmagent\" is invalid: spec.selector.matchLabels: Invalid value: \"%SHARD_NUM%\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')"}
```

The SHARD_NUM is not getting processed by the operator. 

Due to this I am struggling to get vmagent scheduled.

makasim:
The issue has been transferred to the  operator repo.

AndrewChubatiuk:
operator now creates PDB with a spec as is without template rendering since it creates a single PDB despite amount of shards vmagent has.
solution here is to create PDB per replica and in this case pair `shard-num: '%SHARD_NUM%'` should be added automatically by operator

AndrewChubatiuk:
starting operator [v0.66.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.66.0) PDB is created per each vmagent and vmanomaly shard

PR #1554, opened 2025-10-19
vmagent and vmanomaly: create PDB per shard

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1548

### DIFF
--- a/internal/controller/operator/factory/finalize/orphaned_test.go
+++ b/internal/controller/operator/factory/finalize/orphaned_test.go
@@ -15,101 +15,113 @@ import (
 )
 
 func TestRemoveOrphanedDeployments(t *testing.T) {
-	type args struct {
-		ctx             context.Context
-		cr              orphanedCRD
-		keepDeployments map[string]struct{}
-	}
-	tests := []struct {
-		name              string
-		args              args
+	type opts struct {
+		cr                orphanedCRD
+		keepDeployments   map[string]struct{}
 		wantDepCount      int
 		wantErr           bool
 		predefinedObjects []runtime.Object
-	}{
-		{
-			name: "remove nothing",
-			args: args{
-				ctx: context.TODO(),
-				cr: &vmv1beta1.VMAgent{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      "base",
-						Namespace: "default",
-					},
-				},
-				keepDeployments: map[string]struct{}{"base": {}},
-			},
-			predefinedObjects: []runtime.Object{
-				&appsv1.Deployment{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      "base",
-						Namespace: "default",
-						Labels: map[string]string{
-							"app.kubernetes.io/name":      "vmagent",
-							"app.kubernetes.io/instance":  "base",
-							"app.kubernetes.io/component": "monitoring",
-							"managed-by":                  "vm-operator",
-						},
-					},
-				},
-			},
-			wantDepCount: 1,
-		},
-		{
-			name: "remove 1 orphaned",
-			args: args{
-				ctx: context.TODO(),
-				cr: &vmv1beta1.VMAgent{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      "base",
-						Namespace: "default",
-					},
-				},
-				keepDeployments: map[string]struct{}{"base-0": {}},
-			},
-			predefinedObjects: []runtime.Object{
-				&appsv1.Deployment{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      "base-0",
-						Namespace: "default",
-						Labels: map[string]string{
-							"app.kubernetes.io/name":      "vmagent",
-							"app.kubernetes.io/instance":  "base",
-							"app.kubernetes.io/component": "monitoring",
-							"managed-by":                  "vm-operator",
-						},
-					},
-				},
-				&appsv1.Deployment{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      "base-1",
-						Namespace: "default",
-						Labels: map[string]string{
-							"app.kubernetes.io/name":      "vmagent",
-							"app.kubernetes.io/instance":  "base",
-							"app.kubernetes.io/component": "monitoring",
-							"managed-by":                  "vm-operator",
-						},
-					},
-				},
-			},
-			wantDepCount: 1,
-		},
 	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			cl := k8stools.GetTestClientWithObjects(tt.predefinedObjects)
-			if err := RemoveOrphanedDeployments(tt.args.ctx, cl, tt.args.cr, tt.args.keepDeployments); (err != nil) != tt.wantErr {
-				t.Errorf("RemoveOrphanedDeployments() error = %v, wantErr %v", err, tt.wantErr)
-			}
-			var existDep appsv1.DeploymentList
-			opts := client.ListOptions{Namespace: tt.args.cr.GetNamespace(), LabelSelector: labels.SelectorFromSet(tt.args.cr.SelectorLabels())}
-			if err := cl.List(tt.args.ctx, &existDep, &opts); err != nil {
-				t.Fatalf("unexpected error: %v", err)
-			}
-			if tt.wantDepCount != len(existDep.Items) {
-				t.Fatalf("unexpected count of deployments, got:%v, want: %v", len(existDep.Items), tt.wantDepCount)
-			}
-		})
+	f := func(o opts) {
+		t.Helper()
+		cl := k8stools.GetTestClientWithObjects(o.predefinedObjects)
+		ctx := context.TODO()
+		if err := RemoveOrphanedDeployments(ctx, cl, o.cr, o.keepDeployments); (err != nil) != o.wantErr {
+			t.Errorf("RemoveOrphanedDeployments() error = %v, wantErr %v", err, o.wantErr)
+		}
+		var existDep appsv1.DeploymentList
+		lo := client.ListOptions{Namespace: o.cr.GetNamespace(), LabelSelector: labels.SelectorFromSet(o.cr.SelectorLabels())}
+		if err := cl.List(ctx, &existDep, &lo); err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+		if o.wantDepCount != len(existDep.Items) {
+			t.Fatalf("unexpected count of deployments, got:%v, want: %v", len(existDep.Items), o.wantDepCount)
+		}
 	}
+
+	// remove nothing
+	f(opts{
+		cr: &vmv1beta1.VMAgent{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "base",
+				Namespace: "default",
+			},
+		},
+		keepDeployments: map[string]struct{}{"base": {}},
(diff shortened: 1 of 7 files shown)


f41gh7:
cc @vrutkovs

AndrewChubatiuk:
this PR doesn't add templating support for PDB, it creates PDB per shard with additional `shard-num` label. 
> why it's not supported by ServiceSpec, ServiceScrapeSpec, HPA

PDB applies policy for pods, that match provided selector, and in case of multiple deployments there's no guarantee that pods will be rotated properly during upgrade. HPA is not supported by VMAgent, one ServiceScrapeSpec is enough to scrape metrics properly from all pods. For Service maybe it makes sense to have one per shard for scenario, when vmagent shards are configured as remote write destination, but in our case sharding is needed for scraping first of all and Service is mostly used for scraping as well

AndrewChubatiuk:
> 1. it's intrusive change with a lot of refactoring and not necessary changes. It's hard to read.
> 2. commit message doesn't have any motivation of this change.

1. i can split it into several PR's, maybe can apply some changes, that could improve readability for you, but overall idea of "not necessary changes" is to have similar logic of places, where shards are configured.
2. updated

internal/controller/operator/factory/vmagent/vmagent_test.go
```diff
@@ -714,6 +697,23 @@ func TestCreateOrUpdate(t *testing.T) {
...
+									Namespace: "default",
+									Name:      fmt.Sprintf("vmagent-%s-%d", tt.cr.Name, i),
+								}, &pdb); err != nil {
+
+									return false, nil
+								}
+								return true, nil
+							})
+							if err != nil {
+								t.Errorf("cannot wait pdb ready: %s", err)
```
vrutkovs:
```suggestion
								t.Errorf("pdb %s didn't get created: %w",  fmt.Sprintf("vmagent-%s-%d", tt.cr.Name, i), err)
```

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -146,123 +143,38 @@ func CreateOrUpdate(ctx context.Context, cr *vmv1beta1.VMAgent, rclient client.C
...
+	shardCount := cr.GetShardCount()
+	prevShardCount := prevCR.GetShardCount()
+	logger.WithContext(ctx).Info(fmt.Sprintf("using cluster version of VMAgent with shards count=%d", shardCount))
 
 	isUpscaling := false
-	if prevCR != nil && prevCR.Spec.ShardCount != nil {
-		if *prevCR.Spec.ShardCount < shardsCount {
-			logger.WithContext(ctx).Info(fmt.Sprintf("VMAgent shard upscaling from=%d to=%d", *prevCR.Spec.ShardCount, shardsCount))
+	if prevShardCount > 1 {
+		if prevShardCount < shardCount {
```
vrutkovs:
```suggestion
	if prevShardCount > 1  && prevShardCount < shardCount {
```

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -275,88 +187,98 @@ func createOrUpdateShardedDeploy(ctx context.Context, rclient client.Client, cr,
...
+		defer func() {
+			rtCh <- &rv
+			wg.Done()
+		}()
+
+		if cr.Spec.PodDisruptionBudget != nil && !cr.Spec.DaemonSetMode {
+			pdb := build.PodDisruptionBudget(cr, cr.Spec.PodDisruptionBudget)
+			if num != nil {
+				pdb.Name = fmt.Sprintf("%s-%d", pdb.Name, *num)
+				pdb.Spec.Selector.MatchLabels["shard-num"] = strconv.Itoa(*num)
```
vrutkovs: I'd recommend adding more labels (like `app: vmagent` and such) so that we wouldn't accidentally match some other app. It seems to be repeated a few times, so lest have a separate function setting `MatchLabels`

AndrewChubatiuk: `build.PodDisruptionBudget` already configures default selector, here only per shard labels are added

AndrewChubatiuk:
updated PR

internal/controller/operator/factory/finalize/orphaned_test.go
```diff
@@ -15,101 +15,91 @@ import (
...
-					},
-				},
-				keepDeployments: map[string]struct{}{"base": {}},
+	}
+	f := func(o opts) {
+		t.Helper()
+		cl := k8stools.GetTestClientWithObjects(o.predefinedObjects)
+		ctx := context.TODO()
+		if err := RemoveOrphanedDeployments(ctx, cl, o.cr, o.keepDeployments); (err != nil) != o.wantErr {
+			t.Errorf("RemoveOrphanedDeployments() error = %v, wantErr %v", err, o.wantErr)
```
vrutkovs, line 30:
```suggestion
		err := RemoveOrphanedDeployments(ctx, cl, o.cr, o.keepDeployments)
		Expect(err).To(MatchError(o.wantErr), "RemoveOrphanedDeployments() failure")
```
Nitpick: gomega matchers would give us better diff

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -146,125 +142,38 @@ func CreateOrUpdate(ctx context.Context, cr *vmv1beta1.VMAgent, rclient client.C
...
-		if prevObjectSpec != nil {
-			prevAppObject, ok := prevObjectSpec.(*appsv1.DaemonSet)
-			if ok {
-				prevDeploy = prevAppObject
-			}
+			return fmt.Errorf("cannot build new deploy for vmagent: %w", err)
 		}
-		if err := reconcile.DaemonSet(ctx, rclient, newDeploy, prevDeploy); err != nil {
-			return err
+		if reflect.TypeOf(newAppTpl) != reflect.TypeOf(prevAppTpl) {
```
vrutkovs:
```suggestion
		if prevAppTpl.GetObjectKind() != newAppTpl.GetObjectKind() {
```
Nitpick: this would remove `reflect` import and would probably be bit faster

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -146,125 +142,38 @@ func CreateOrUpdate(ctx context.Context, cr *vmv1beta1.VMAgent, rclient client.C
...
-	deploymentNames := make(map[string]struct{})
-	stsNames := make(map[string]struct{})
-	shardsCount := *cr.Spec.ShardCount
-	logger.WithContext(ctx).Info(fmt.Sprintf("using cluster version of VMAgent with shards count=%d", shardsCount))
+	deploymentToKeep := make(map[string]struct{})
+	stsToKeep := make(map[string]struct{})
+	pdbToKeep := make(map[string]struct{})
+	shardCount := cr.GetShardCount()
+	prevShardCount := prevCR.GetShardCount()
+	logger.WithContext(ctx).Info(fmt.Sprintf("using cluster version of VMAgent with shards count=%d", shardCount))
```
vrutkovs:
```suggestion
	logger.WithContext(ctx).Info("using cluster version of VMAgent", "shardCount", shardCount))
```
nitpick: structured logging will be useful later

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -146,125 +142,38 @@ func CreateOrUpdate(ctx context.Context, cr *vmv1beta1.VMAgent, rclient client.C
...
 
 	isUpscaling := false
-	if prevCR != nil && prevCR.Spec.ShardCount != nil {
-		if *prevCR.Spec.ShardCount < shardsCount {
-			logger.WithContext(ctx).Info(fmt.Sprintf("VMAgent shard upscaling from=%d to=%d", *prevCR.Spec.ShardCount, shardsCount))
-			isUpscaling = true
-		}
+	if prevShardCount > 1 && prevShardCount < shardCount {
+		logger.WithContext(ctx).Info(fmt.Sprintf("VMAgent shard upscaling from=%d to=%d", prevShardCount, shardCount))
```
vrutkovs:
```suggestion
		logger.WithContext(ctx).Info(fmt.Sprintf("VMAgent shard upscaling", "from", prevShardCount, "to", shardCount))
```
nitpick: structured logging

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -275,88 +184,90 @@ func createOrUpdateShardedDeploy(ctx context.Context, rclient client.Client, cr,
...
+			}
+			if err = reconcile.Deployment(ctx, rclient, newApp, prevApp, false); err != nil {
+				rv.err = fmt.Errorf("cannot reconcile deployment for sharded vmagent(%d): %w", num, err)
+				return
+			}
+			rv.deploymentName = newApp.Name
+		case *appsv1.StatefulSet:
+			newApp, err := build.RenderShard(newAppTpl, nil, num)
+			if err != nil {
+				rv.err = err
```
vrutkovs:
```suggestion
				rv.err = fmt.Errorf("cannot render new Statefulset shard: %w", err)
```
nitpick: let's wrap errors earlier so that we can spot a precise point where it happened

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -275,88 +184,90 @@ func createOrUpdateShardedDeploy(ctx context.Context, rclient client.Client, cr,
...
+				return
+			}
+			patchShardContainers(newApp.Spec.Template.Spec.Containers, num, shardCount)
+			var prevApp *appsv1.StatefulSet
+			if prevAppTpl != nil {
+				prevApp, err = build.RenderShard(prevAppTpl.(*appsv1.StatefulSet), nil, num)
 				if err != nil {
-					rv.err = fmt.Errorf("cannot fill placeholders for sts in sharded vmagent(%d): %w", shardNum, err)
+					rv.err = err
```
vrutkovs:
```suggestion
					rv.err = fmt.Errorf("cannot render previous Statefulset shard: %w", err)
```

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -1390,19 +1279,9 @@ func deletePrevStateResources(ctx context.Context, rclient client.Client, cr, pr
 		return fmt.Errorf("cannot remove additional service: %w", err)
 	}
 	objMeta := metav1.ObjectMeta{Name: cr.PrefixedName(), Namespace: cr.Namespace}
-	if cr.Spec.PodDisruptionBudget == nil && prevCR.Spec.PodDisruptionBudget != nil {
```
vrutkovs, line 1315: Not sure why we remove this - we handle PDB somewhere else?

vrutkovs (approved):
LGTM, few suggestions

internal/controller/operator/factory/finalize/orphaned_test.go
AndrewChubatiuk: currently we're using gomega matchers in tests at `test/` folder only

internal/controller/operator/factory/vmagent/vmagent.go
AndrewChubatiuk: please check `updateShard` function, PDB management should be inside

vrutkovs: oh, I see, thanks

PR #1562, opened 2025-10-22
e2e: adaptive concurrency

AndrewChubatiuk:
adjust concurrency up to environment resources
currently vmanomaly tests fail frequently in CI, but always run without issues locally


f41gh7 (approved):
LGTM

f41gh7:
But it seems, that tests fail anyway

PR #1564, opened 2025-10-22
reconcile: do not rely on prev spec during cleanup for (vm|vt|vl)cluster CRs

AndrewChubatiuk:
- removed impact of previous state during dangling resources removal for cluster CRs
- removed reconcile.AdditionalServices function, that is not relevant to reconcile package and depends on previous state, replaced it with function for orphaned services removal
- e2e: do not wrap CR update into Eventually since update should be applied immediately and repetitive calls may lead to conflicts

fixes https://github.com/VictoriaMetrics/operator/issues/1580


internal/controller/operator/factory/finalize/cluster.go
```diff
@@ -0,0 +1,200 @@
...
+	owner := cr.AsOwner()
+	for _, objToRemove := range objsToRemove {
+		if err := SafeDeleteWithFinalizer(ctx, rclient, objToRemove, &owner); err != nil {
+			return fmt.Errorf("failed to remove object=%s: %w", objToRemove.GetObjectKind().GroupVersionKind(), err)
+		}
+	}
+	return nil
+}
+
+// OnInsertDelete removes all objects related to insert component
```
f41gh7: select component

internal/controller/operator/factory/finalize/cluster.go
```diff
@@ -0,0 +1,200 @@
...
+	owner := cr.AsOwner()
+	for _, objToRemove := range objsToRemove {
+		if err := SafeDeleteWithFinalizer(ctx, rclient, objToRemove, &owner); err != nil {
+			return fmt.Errorf("failed to remove lb object=%s: %w", objToRemove.GetObjectKind().GroupVersionKind(), err)
+		}
+	}
+	return nil
+}
+
+type ChildCleaner struct {
```
f41gh7, line 152: It'd be great to have comments on public exported Structs and methods

internal/controller/operator/factory/finalize/cluster.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func (cc *ChildCleaner) KeepScrape(v string) {
+	cc.scrapes[v] = struct{}{}
+}
+
+func (cc *ChildCleaner) RemoveOrphaned(ctx context.Context, rclient client.Client, cr build.ParentOpts) error {
+	b := build.NewChildBuilder(cr, vmv1beta1.ClusterComponentRoot)
+	ls := b.SelectorLabels()
+	delete(ls, "app.kubernetes.io/name")
```
f41gh7, line 195: Any reason why this label should be deleted? Add comment please to clarify this logic

AndrewChubatiuk: dependent cluster components don't have the same value of `app.kubernetes.io/name`, that's why it's removed to have selector that matches all dependent resources

AndrewChubatiuk:
thanks for review
addressed your comments

PR #1565, opened 2025-10-23
use existing github action for runner cleanup

AndrewChubatiuk:
replace custom runner cleanup scripts with github action


f41gh7:
FYI, this change was reverted. It takes too much CI time to perform clean job.

PR #1566, opened 2025-10-23
e2e: use custom timeouts for anomaly, change e2e registry to quay.io

AndrewChubatiuk:
- e2e tests are hitting rate limit on hub.docker.io from time to time, switching to quay.io
- use already defined anomaly specific timeouts on cases that are failing frequently enough


f41gh7 (approved):
LGTM

PR #1567, opened 2025-10-23
properly handle reconcile errors

AndrewChubatiuk:
- consistently reconcile across all custom resources
- some controllers do update reconcile result if error is present, which is not allowed
- do not propagate error if it's of type `ErrWaitReady`
- additionally wrap into `ErrWaitReady` error such errors as:
  - `cannot find any pod for selector`
  - `sts wasn't yet removed`

### DIFF
--- a/internal/controller/operator/controllers.go
+++ b/internal/controller/operator/controllers.go
@@ -398,6 +398,7 @@ func reconcileAndTrackStatus[T client.Object, ST reconcile.StatusWithMetadata[ST
 		desiredStatus := vmv1beta1.UpdateStatusFailed
 		if reconcile.IsErrorWaitTimeout(err) {
 			desiredStatus = vmv1beta1.UpdateStatusExpanding
+			err = nil
 		}
 		if updateErr := reconcile.UpdateObjectStatus(ctx, c, object, desiredStatus, err); updateErr != nil {
 			resultErr = fmt.Errorf("failed to update object status: %q, origin err: %w", updateErr, err)
--- a/internal/controller/operator/factory/reconcile/deploy.go
+++ b/internal/controller/operator/factory/reconcile/deploy.go
@@ -160,5 +160,7 @@ func reportFirstNotReadyPodOnError(ctx context.Context, rclient client.Client, o
 		}
 		return podStatusesToError(origin, &dp)
 	}
-	return fmt.Errorf("cannot find any pod for selector=%q, check kubernetes events, origin err: %w", selector.String(), origin)
+	return &errWaitReady{
+		origin: fmt.Errorf("cannot find any pod for selector=%q, check kubernetes events, origin err: %w", selector.String(), origin),
+	}
 }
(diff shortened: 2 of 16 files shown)


internal/controller/operator/factory/finalize/common.go
```diff
@@ -213,5 +214,28 @@ func FreeIfNeeded(ctx context.Context, rclient client.Client, object client.Obje
...
 	}
-	return fmt.Errorf("deletionTimestamp is not zero=%q for object=%s/%s kind=%s, recreating it at next reconcile loop. Warning never delete object manually", object.GetDeletionTimestamp(), object.GetNamespace(), object.GetName(), object.GetObjectKind().GroupVersionKind())
+	return &ErrWaitReady{
+		Err: fmt.Errorf("deletionTimestamp is not zero=%q for object=%s/%s kind=%s, recreating it at next reconcile loop. Warning never delete object manually", object.GetDeletionTimestamp(), object.GetNamespace(), object.GetName(), object.GetObjectKind().GroupVersionKind()),
+	}
+}
+
+// IsErrorWaitTimeout determines if the err is an error which indicates that timeout for app
+// transition into Ready state reached and should be continued at the next reconcile loop
+func IsErrorWaitTimeout(err error) bool {
```
f41gh7: For me, this error must belong to reconcile package.

AndrewChubatiuk: maybe, but using this error in finalize.RemoveFinalizer and cannot import `reconcile` inside `finalize` due to import cycle

internal/controller/operator/factory/finalize/common.go
```diff
@@ -213,5 +214,28 @@ func FreeIfNeeded(ctx context.Context, rclient client.Client, object client.Obje
 	if err := RemoveFinalizer(ctx, rclient, object); err != nil {
 		return fmt.Errorf("cannot remove finalizer from object=%s/%s, kind=%q: %w", object.GetNamespace(), object.GetName(), object.GetObjectKind().GroupVersionKind(), err)
 	}
-	return fmt.Errorf("deletionTimestamp is not zero=%q for object=%s/%s kind=%s, recreating it at next reconcile loop. Warning never delete object manually", object.GetDeletionTimestamp(), object.GetNamespace(), object.GetName(), object.GetObjectKind().GroupVersionKind())
+	return &ErrWaitReady{
```
f41gh7: This error must be returned to the user and it cannot be skipped. It indicates, that something is wrong and reconcile loop must be started again. 

 I think, instead of returning some wrapped error, it's better to check for exist finalizers and poll for object deletion:

```
	if object.GetDeletionTimestamp().IsZero() {
		// fast path
		return nil
	}
    if len(object.GetFinalizers) == 0 {
      wait.PollUntil(func{
           GetObject() == err.NotFound
     })
       return nil
   }

```c

AndrewChubatiuk:
@f41gh7 updated PR

f41gh7:
Thanks for contribution!

PR #1568, opened 2025-10-23
upgrade jimmidyson reloader, allow preverve original registry for config reloader if using global registry

AndrewChubatiuk:
- upgraded jimmidyson config reloader, new versions are present on ghcr.io
- jimmidyson has no correspondent image on quay.io, which causes errors for alertmanager e2e tests. added preserveregistry config reloader option to preserve original registry if global one is set


f41gh7 (changes requested):
> jimmidyson has no correspondent image on quay.io, which causes errors for alertmanager e2e tests. added preserveregistry config reloader option to preserve original registry if global one is set

 If it causes issues e2e tests - set container image explicitly.  There is no need at new configuration option, which is kinda confusing and hard to use.

 The same applies to the other apps - if there is no such image at registry, it's either must be added or changed for particular resource.


 Long term solution will be completely deprecate 3rd party config-reloaders and use only own implementation by default.

PR #1569, opened 2025-10-23
upgrade jimmidyson reloader

AndrewChubatiuk:
upgrade jimmidyson reloaders, take into account in registry replacement function also docker.io and ghcr.io


f41gh7:
This change must be replaced by https://github.com/VictoriaMetrics/operator/pull/1571

PR #1570, opened 2025-10-23
vmalert: rewrite tests using f-test

AndrewChubatiuk:
also add test case for initial concern from issue, that is linked to https://github.com/VictoriaMetrics/operator/pull/703


f41gh7:
Thanks for contribution!

PR #1575, opened 2025-10-24
do not update prev spec annotation in case of CR expanding state

AndrewChubatiuk:
- update prev spec in CR annotations after successful or failed reconcile. currently annotation is updated before reconcile, which may lead to dangling resources if one of reconcile loops finished with an error, that doesn't lead to CR failed state. this may explain e2e flaky tests in CI
- e2e: print resources that were not removed
- e2e: do not enclose CR update into `Eventually` since update should be applied immediately and repetitive calls may lead to conflicts
- e2e: consistently use `nsn` instead of `nss` for `type.NamespacedName` variables
- added predicate filter to exclude change in last-applied-spec annotation from reconcile

### DIFF
--- a/api/operator/v1beta1/vmextra_types.go
+++ b/api/operator/v1beta1/vmextra_types.go
@@ -49,8 +49,8 @@ const (
 	SkipValidationValue      = "true"
 	AdditionalServiceLabel   = "operator.victoriametrics.com/additional-service"
 	// PVCExpandableLabel controls checks for storageClass
-	PVCExpandableLabel            = "operator.victoriametrics.com/pvc-allow-volume-expansion"
-	lastAppliedSpecAnnotationName = "operator.victoriametrics/last-applied-spec"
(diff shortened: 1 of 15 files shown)


f41gh7:
This was already implemented, but replaced with current version, please see commit and issues history

AndrewChubatiuk:
before [this commit](https://github.com/VictoriaMetrics/operator/commit/471f183f1eed6eaae60c86c9a96731f3c00f5d16) prev spec annotation was updated after successful reconcile which led to infinite loops for CRs that are always failing and after this commit prev spec annotation is updated before reconcile but it disabled dangling resources cleanup if CR update was not able to complete within a single reconcile execution (create deployment and timed out during first reconcile, which ended with `expanding` state and second reconcile, which switched state to `operational`)

AndrewChubatiuk:
after discussion implemented another solution for this issue in https://github.com/VictoriaMetrics/operator/pull/1564

Issue #1581, opened 2025-10-29
Config reloader doesn't support IPv6


dhess:
We run the operator, along with the `k8s-stack` Helm chart, in an IPv6 cluster. With the latest release of the operator Helm chart, which enables the operator's config reloader, all of our pods break due to liveness check failures on port `8435`, I believe because the `config-reloader` only listens on IPv4:

```
config-reloader 2025-10-29T18:33:40.952Z    info    /go/pkg/mod/github.com/!victoria!metrics/!victoria!metrics@v1.116.0/lib/httpserver/httpserver.go:170    started server at http://0.0.0.0:8435/
```

I've added the following config:

```yaml
  configReloaderExtraArgs:
    reload-url: http://[::1]:8429/-/reload
    http.listenAddr: "[::]:8435"
```

and the `config-reloader` accepts the args, but it doesn't make any difference, as it still listens on `0.0.0.0:8435`.

Other components of VictoriaMetrics also have this issue, but those are solved by setting `enableTCP6` on the affected components. Presumably the config reloader needs `enableTCP6` support, as well.

f41gh7:
Sorry for breaking deployments. It was not expected and I think, we should make `-enableTCP6` as a default arg for the new config-reloader container.


 Currently, it could be solved with `spec.configReloaderExtraArgs: enableTCP6: "true"`. This option must be supported, since it uses standard VictoriaMetrics httpserver.

f41gh7:
Also, maybe it's better to introduce a new config option to the operator, which will enable `ipv6` by default for the all managed components?

benben:
It should be the default for the config-reloader container when you have it set on `VMAuth` et al already.

AndrewChubatiuk:
In operator [v0.66.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.66.0) `VM_ENABLETCP6` global parameter was added, which automatically enables TCP6 on managed resources.

PR #1582, opened 2025-10-29
add VM_ENABLETCP6 env variable to enable IPv6 by default

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1581
- add VM_ENABLETCP6 env variable that enables support of IPv6 across all managed CRs
- consistently set defaults from global config


f41gh7:
WDYT of making a global configuration for operator, which will automatically add `enableTCP6` arg to the all components? And change `127.0.0.1` config-reloader urls to the `localhost`. 

It should simplify management of `IPV6` only installations.

AndrewChubatiuk:
> And change 127.0.0.1 config-reloader urls to the localhost

we already have `localhost` everywhere

f41gh7:
Thanks for contribution!

Issue #1583, opened 2025-10-30
vlcluster vlinsert storageNode argument missing commas.


elee1766:
since updating to the latest version of k8s-stack (and therefore the operator), my vlinsert nodes fail to connect to vlstorage since the -storageNode just puts all the nodes together without commas in between.

i think this is a result of this commit: https://github.com/VictoriaMetrics/operator/commit/ff722eb3ba4e72765548b4353b5f370b42d143f7#diff-935a27d82c61eccfc9fb9fd33ad10a1e4166983d4589bb61486816e997c99cd5 

which broke it (see build.go changes)

finomen:
same issue after update to v0.64.0, revert to 0.63.0 fixed it

AndrewChubatiuk:
thanks for spotting this issue, new option for `storageNode` flag generation was not properly ported to all components, that are using it, fixed in #1584

AndrewChubatiuk:
fix released in 0.64.1, operator chart 0.55.3 and k8s-stack 0.62.2

PR #1584, opened 2025-10-30
vlcluster: fix vlinsert storageNode argument generation

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1583


f41gh7:
Thanks for contribution!

PR #1585, opened 2025-10-30
converter: support ServiceMonitor's serviceDiscoveryRole

AndrewChubatiuk:
adds ability to convert ServiceMonitor's ServiceDiscoveryRole

To support this option it was required to upgrade some packages, which led to inability to use fake client as it complains while operating with uint64 numbers, that are used in `SeriesLimit` and `SampleLimit` properties. Also noted, type of these limits in operator and VM repo differs - in operator they are uint64 (most likely to be compatible with prometheus operator) and in VM - int. In this PR I've changed type of these limits to int to be compatible with VM and not setting them during conversion if prometheus CRs uint64 limit overflow int.

@f41gh7 @Haleygo 
do you think we need to change types of limits to int in operator or to uint64 in VM?

### DIFF
--- a/internal/controller/operator/converter/apis.go
+++ b/internal/controller/operator/converter/apis.go
@@ -1,6 +1,7 @@
 package converter
 
 import (
+	"math"
 	"strings"
 
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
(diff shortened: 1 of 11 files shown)


internal/controller/operator/converter/apis.go
```diff
@@ -156,6 +158,20 @@ func ReplacePromDirPath(origin string) string {
 	return origin
 }
 
+func convertDiscoveryRole(src *promv1.ServiceDiscoveryRole) string {
+	role := "endpoints"
+	if src == nil {
+		return role
```
f41gh7, line 164: I think, it's better to keep it empty. It'll make possible to use global configuration for discoveryRole `ENFORCEENDPOINTSLICES`

f41gh7:
> do you think we need to change types of limits to int in operator or to uint64 in VM?

 Let's keep it as int. Practically, it's extremely rare case, when some limit could be greater than uint32.

internal/controller/operator/converter/apis.go
AndrewChubatiuk: updated

Issue #1579, opened 2025-10-29
feature request: configurable rollout strategy for vmcluster


dctrwatson:
I recently upgraded from v0.60.2 to v0.63.0 and the operator now seems to do something between minimum and no time downtime strategy when performing vmcluster changes.

It will update the vmstorage and vmselect STS in parallel. And then once the vmselect STS rollout is complete, it will update the vminsert deployment. This makes for a very noisy rollout and cluster availability that gets worse the larger the cluster is.

I much prefer the old behavior of doing the equivalent to [No Downtime Strategy](https://docs.victoriametrics.com/victoriametrics/cluster-victoriametrics/#no-downtime-strategy). We have tuned many aspects of the VictoriaMetrics components and k8s to have little or no interruptions when rollouts are performed in that manner.

f41gh7:
> I recently upgraded from v0.60.2 to v0.63.0 and the operator now seems to do something between minimum and no time downtime strategy when performing vmcluster changes.

 It shouldn't. Operator performs new parallel updates only if `rollingUpdateStrategyBehavior` is set. Otherwise, it performs old no downtime strategy with updating pods one by one.

 Could you please provide an example of `VMCluster` yaml configuration?

dctrwatson:
```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
metadata:
  name: canary
  namespace: monitoring
spec:
  replicationFactor: 2
  retentionPeriod: "3"
  vminsert:
    extraEnvs:
      - name: GOMAXPROCS
        value: "2"
    hpa:
      maxReplicas: 12
      metrics:
        - pods:
            metric:
              name: vm_concurrent_insert_utilization
            target:
              averageValue: 800m
              type: AverageValue
          type: Pods
      minReplicas: 6
    image:
      tag: v1.128.0-cluster
    minReadySeconds: 180
    podDisruptionBudget:
      maxUnavailable: 1
    port: "8480"
    priorityClassName: monitoring-canary
    replicaCount: 6
    resources:
      limits:
        memory: 2Gi
      requests:
        cpu: 500m
    rollingUpdate:
      maxSurge: 0
      maxUnavailable: 1
    serviceSpec:
      metadata:
        annotations:
          service.kubernetes.io/topology-mode: Auto
        name: vminsert-canary-az-aware
      spec: {}
    terminationGracePeriodSeconds: 300
    topologySpreadConstraints:
      - maxSkew: 1
[...]

PR #1588, opened 2025-10-30
reconcile: properly check deployment and statefulset ready status

f41gh7:
Previously, due to race condition between local client cache and actual
 Kubernetes api-server, operator may incorrectly track ready status for
 Deployment and StatefulSet.

 client.Update(object) call increments `Generation` field of the update
object if there was any change. While, client.Get(object) may not yet have this change.

 So this commit explicitly checks if updateObject.Generation is greater
 than getObject.Generation, to prevent such case.

Fixes https://github.com/VictoriaMetrics/operator/issues/1579

### DIFF
--- a/internal/controller/operator/factory/reconcile/deploy.go
+++ b/internal/controller/operator/factory/reconcile/deploy.go
@@ -102,7 +102,10 @@ func waitDeploymentReady(ctx context.Context, rclient client.Client, dep *appsv1
 		// (https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment)
 		// this function uses the deployment readiness detection algorithm from `kubectl rollout status` command
 		// (https://github.com/kubernetes/kubectl/blob/6e4fe32a45fdcbf61e5c30ebdc511d75e7242432/pkg/polymorphichelpers/rollout_status.go#L76)
-		if actualDeploy.Generation > actualDeploy.Status.ObservedGeneration {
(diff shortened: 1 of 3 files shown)


internal/controller/operator/factory/reconcile/deploy.go
```diff
@@ -102,7 +102,10 @@ func waitDeploymentReady(ctx context.Context, rclient client.Client, dep *appsv1
 		// (https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment)
 		// this function uses the deployment readiness detection algorithm from `kubectl rollout status` command
 		// (https://github.com/kubernetes/kubectl/blob/6e4fe32a45fdcbf61e5c30ebdc511d75e7242432/pkg/polymorphichelpers/rollout_status.go#L76)
-		if actualDeploy.Generation > actualDeploy.Status.ObservedGeneration {
+		if actualDeploy.Generation > actualDeploy.Status.ObservedGeneration ||
+			// special case to prevent possible race condition between updated object and local cache
+			// See this issue https://github.com/VictoriaMetrics/operator/issues/1579
+			dep.Generation > actualDeploy.Generation {
```
AndrewChubatiuk, line 108:
```suggestion
		if max(actualDeploy.Generation, dep.Generation) > actualDeploy.Status.ObservedGeneration {
```

f41gh7: I'd prefer to keep it simple, this version looks too smart for me.

Issue #1589, opened 2025-10-30
VMAgent stuck in reconcile loop


micke:
I just deployed my first `VMAgent` via the operator running in helm chart `0.55.3` but it seem to be stuck in a reconcile loop:

```json5
{"level":"info","ts":"2025-10-30T23:37:47Z","logger":"controller.VMAgent","msg":"object has changes with previous state, applying changes","vmagent":"victoria-metrics","namespace":"victoria-metrics"}
[...]

PR #1590, opened 2025-10-31
config-reloader: properly set default for useVMConfigReloader value

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1589


f41gh7:
Thanks for contribution!

Issue #1592, opened 2025-11-03
Operator is not setting VMAuth loadBalancingPolicy  commandline flag


thejuan:
From the [docs](https://docs.victoriametrics.com/victoriametrics/vmauth/#load-balancing)
```
By default, vmauth uses least_loaded policy to spread the incoming requests among available backends. The policy can be changed to first_available via -loadBalancingPolicy command-line flag.
```

Setting [this property in the CR](https://docs.victoriametrics.com/operator/api/#vmauthspec-load_balancing_policy) has no effect. The commandline flag is not set.

Operator version v0.63.0

```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAuth

spec:
...
  load_balancing_policy: first_available
...
status:
  observedGeneration: 25
  updateStatus: operational

```

Haleygo:
Hi @thejuan ,
Sorry for the unclear document! The `load_balancing_policy`(or other VMUserConfigOptions under vmauthSpec) is for `unauthorized_user` defined in `vmauthSpec.unauthorizedAccessConfig`, not for vmauth.
To fix this confusion, we introduced `vmauthSpec.unauthorizedUserAccessSpec` to replace `vmauthSpec.unauthorizedAccessConfig` and include the options in [v0.51.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.51.0), and we will removed those fields in v1.0.
>Update note 2: VMAuth.spec.unauthorizedAccessConfig is deprecated in favour of VMAuth.spec.unauthorizedUserAccessSpec. Operator still serves deprecated fields until v1.0 release.

> The commandline flag is not set.

If you want set `-loadBalancingPolicy` command line flag, please add it in `vmauthSpec.extraArgs` as
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAuth

spec:
  extraArgs:
    loadBalancingPolicy: first_available
```

AndrewChubatiuk:
closing as answered

PR #1593, opened 2025-11-03
docs: clarify vmauth VMUserConfigOptions and update docs

Haleygo:
add some clarifications for https://github.com/VictoriaMetrics/operator/issues/1592 in code, unfortunately it can't be reflected in the generated doc


f41gh7:
Thanks for contribution!

Issue #1594, opened 2025-11-02
bug: HPA for VMAgent triggers unnecessary restarts


caryjpaepen:
**Chart name and version**
chart: victoria-metrics-distributed
version: v 0.21.1

**Describe the bug**

The operator sets sharding properties (`promscrape.cluster.membersCount`) on VMAgent even when scraping/streaming aggregation is disabled. In the distributed chart—where agents act only as ingress buffers—this causes unnecessary restarts of all agents during HPA-driven scaling.

Because the Helm chart does not currently support creating an HPA directly, we apply a separate HPA manifest.

E.g. (VMAgent pods labeled `zone=az1`)
```
spec:
  maxReplicas: 100
  minReplicas: 10
  - external:
      metric:
        name: xxx
        selector:
          matchLabels:
            zone: az1
      target:
        type: Value
        value: "10"
    type: External
  scaleTargetRef:
    apiVersion: operator.victoriametrics.com/v1beta1
    kind: VMAgent
    name: az1
```

As the HPA scales from 10 to 100 replicas, VMAgent pods are recycled on every step.

Haleygo:
Hi,
vm-operator shouldn't add `promscrape.cluster.membersCount` unless `VMAgentSpec.ShardCount` is specified. 
Is the vmagent per-zone or an extra vmagent? Could you share its spec?

caryjpaepen:
Hey Haleygo,

This is the per-zone vmagent. 

Yes, we had to use ShardCount in order to use HPA. The `subresources.scale.specReplicasPath` specified by `vmagents.operator.victoriametrics.com` is indeed `.spec.shardCount`. This dictates what we could use.

A snippet of our VMAgent resource from one of the zones:

```
spec:
  affinity: {}
  extraArgs:
    loggerFormat: json
    maxConcurrentInserts: "40"
  image:
    tag: v1.125.1
  imagePullSecrets:
  - name: use-docker
  port: "8429"
  remoteWrite:
  - url: http://vminsert-az1.victoriametrics-store.svc.cluster.local.:8480/insert/0/prometheus/api/v1/write
  - url: http://vminsert-az2.victoriametrics-store.svc.cluster.local.:8480/insert/0/prometheus/api/v1/write
  remoteWriteSettings:
    showURL: true
  shardCount: 10
  statefulMode: true
  statefulStorage:
    volumeClaimTemplate:
      spec:
        resources:
          requests:
            storage: 2Ti
```

[...]

PR #1596, opened 2025-11-04
vmagent: do not set promscrape.cluster.* flags in ingestOnly mode

AndrewChubatiuk:
related issue #1594

`promescrape.cluster.*` flags are causing restarts while scaling are not needed in ingestOnly mode. Removing them for ingestOnly mode should prevent all pods from restarting while scaling


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #1580, opened 2025-10-29
bug: operator may leave dangling resources after CR update


AndrewChubatiuk:
each VM operator CR has `operator.victoriametrics/last-applied-spec` annotation, where operator stores last applied CR spec as JSON. During each reconcile it unmarshals this value from that annotation into CR spec and depending on a diff with a current spec it performs dependent resources creation / update / removal. Value of this annotation is updated before each reconcile, which usually doesn't produce issues, but sometimes (e.g: takes much time for pod to become ready) it may take more than one Reconcile execution till CR reaches `operational` state and since annotation after first reconcile keeps value of current spec it cannot produce correct diff to remove dangling resources. This should not be a serious issue, since resources still have owner reference and should be removed as soon as parent CR is removed, but this is a reason of frequent test failures, which may hide other operator issues.

possible solution is to try to retrieve using k8s resources which could be created for certain CR and remove if they are present and have relevant owner reference and finalizer. but in some cases it's required to list all resources filtered by CR labels, like `Service` that is managed by `serviceSpec`, as they may have custom name, which is not distinguishable from CR spec, where `serviceSpec` is not set

AndrewChubatiuk:
starting operator [v0.66.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.66.0) dangling resources are properly removed

PR #1597, opened 2025-11-04
operator: change AsOwner return type from slice to struct

AndrewChubatiuk:
changing return type of AsOwner allows to simplify lookup logic during dangling resources deletion
related issue https://github.com/VictoriaMetrics/operator/issues/1580


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1600, opened 2025-11-05
vol 1: extracted subset of changes from #1564

AndrewChubatiuk:
PR contains subset of straightforward changes from https://github.com/VictoriaMetrics/operator/pull/1564 that should be easy to review

- simplify LastAppliedChangesAsPatch signature, object metadata is not needed there
- added predicate to disable reconcile for `operator.victoriametrics/last-applied-spec` annotation change
- add owner reference for vmpodscrape 
- add retryOnConflict function that retries if conflict happens or if resource is being deleted (errRecreate was thrown)
- throw errRecreate for resources that have non-zero deletion timestamp


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1603, opened 2025-11-06
use ubi image for olm

AndrewChubatiuk:


Makefile
```diff
@@ -237,6 +238,7 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 
 publish:
 	TAG=$(TAG) ROOT=./cmd $(MAKE) docker-buildx
+	TAG=$(TAG)-ubi BASEIMAGE=redhat/ubi8-minimal ROOT=./cmd $(MAKE) docker-buildx
```
vrutkovs:
```suggestion
	TAG=$(TAG)-ubi BASEIMAGE=redhat/ubi9-micro ROOT=./cmd $(MAKE) docker-buildx
```
RHEL8 is too old, let's use at least 9? "old" here means it would be out of support in mid-2029 though.

`minimal` variant has a package manager, which we don't need - so `micro` is a better idea.

vrutkovs:
Could you also pull in a few more required changes from https://github.com/VictoriaMetrics/operator/pull/1604?

Things to discuss:
* RHEL 10 vs RHEL 9 vs RHEL 8. RHEL 10 requires x86-64-v3, RHEL 8 will soon be out of support
* Extra binaries installed in the image - coreutils, bash. Are we fine with this, or we rather release a separate, `scratch` based version?
* Maybe let's not use the `scratch` version at all? Cost is image growth - 84 MB -> 124 MB.

Makefile
AndrewChubatiuk: updated

PR #1605, opened 2025-11-07
fix: use new variable for rendered shards instead of copy

AndrewChubatiuk:
removed unneeded CRDs copying since new copy should be rendered from templates


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1606, opened 2025-11-07
follow up for #c280a4e5

AndrewChubatiuk:
use retrieved object from API for deletion timestamp check instead of crafted one


f41gh7 (approved):
LGTM

PR #1610, opened 2025-11-08
vol 2: extracted subset of changes from #1564

AndrewChubatiuk:
this makes refactoring of cluster components (select, insert, lb, storage) creation/removal for this purpose it does the following:
- use common `ChildBuilder` interface as a replacement for per-module optsBuilder to unify and simplify cluster components removal/creation and should help to simplify implementation in https://github.com/VictoriaMetrics/operator/pull/1564
- to unify child builder creation (optsBuilder is some places calls newOptsBuilder sometimes not) replaced per component functions like `Get*Name`, `*PodLabels`, `*PodMetadata`, `*SelectorLabels`, `*URL` with single function that returns output depending on component kind argument

### DIFF
--- a/api/operator/v1/vlcluster_types.go
+++ b/api/operator/v1/vlcluster_types.go
@@ -109,6 +109,8 @@ func (cr *VLCluster) PodMetadata(kind vmv1beta1.ClusterComponent) *vmv1beta1.Emb
 		return cr.Spec.VLStorage.PodMetadata
 	case vmv1beta1.ClusterComponentBalancer:
 		return cr.Spec.RequestsLoadBalancer.Spec.PodMetadata
+	default:
+		panic("BUG unsupported cluster kind=%s", string(kind))
 	}
 	return nil
 }
@@ -136,6 +138,8 @@ func (cr *VLCluster) GetAdditionalService(kind vmv1beta1.ClusterComponent) *vmv1
 		return cr.Spec.VLStorage.ServiceSpec
 	case vmv1beta1.ClusterComponentBalancer:
 		return cr.Spec.RequestsLoadBalancer.Spec.AdditionalServiceSpec
+	default:
+		panic("BUG unsupported cluster kind=%s", string(kind))
 	}
 	return nil
 }
@@ -820,6 +824,8 @@ func (cr *VLCluster) AsURL(kind vmv1beta1.ClusterComponent) string {
 			}
 		}
 		extraArgs = cr.Spec.VLStorage.ExtraArgs
+	default:
+		panic("BUG unsupported cluster kind=%s", string(kind))
 	}
 	return fmt.Sprintf("%s://%s.%s.svc:%s", vmv1beta1.HTTPProtoFromFlags(extraArgs), cr.PrefixedName(kind), cr.Namespace, port)
 }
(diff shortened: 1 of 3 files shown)


api/operator/v1/vlcluster_types.go
```diff
@@ -81,6 +81,112 @@ type VLClusterSpec struct {
...
+		if cr.Spec.VLSelect == nil {
+			return nil
+		}
+		return cr.Spec.VLSelect.PodMetadata
+	case vmv1beta1.ClusterComponentStorage:
+		if cr.Spec.VLStorage == nil {
+			return nil
+		}
+		return cr.Spec.VLStorage.PodMetadata
+	case vmv1beta1.ClusterComponentBalancer:
```
f41gh7, line 110: default panic is missing. It's a BUG if default case was met

api/operator/v1/vlcluster_types.go
```diff
@@ -81,6 +81,112 @@ type VLClusterSpec struct {
...
+		if cr.Spec.VLSelect == nil {
+			return nil
+		}
+		return cr.Spec.VLSelect.ServiceSpec
+	case vmv1beta1.ClusterComponentStorage:
+		if cr.Spec.VLStorage == nil {
+			return nil
+		}
+		return cr.Spec.VLStorage.ServiceSpec
+	case vmv1beta1.ClusterComponentBalancer:
```
f41gh7, line 139: The same default is bug

api/operator/v1/vlcluster_types.go
```diff
@@ -829,81 +766,62 @@ func (cr *VLCluster) IsOwnsServiceAccount() bool {
...
+			port = "9491"
+		}
+		if cr.Spec.VLStorage.ServiceSpec != nil && cr.Spec.VLStorage.ServiceSpec.UseAsDefault {
+			for _, svcPort := range cr.Spec.VLStorage.ServiceSpec.Spec.Ports {
+				if svcPort.Name == "http" {
+					port = fmt.Sprintf("%d", svcPort.Port)
+				}
 			}
 		}
+		extraArgs = cr.Spec.VLStorage.ExtraArgs
```
f41gh7, line 826: default case

f41gh7 (approved):
LGTM

 please see minor comments

api/operator/v1/vlcluster_types.go
```diff
@@ -81,6 +81,116 @@ type VLClusterSpec struct {
...
+		return cr.Spec.VLSelect.PodMetadata
+	case vmv1beta1.ClusterComponentStorage:
+		if cr.Spec.VLStorage == nil {
+			return nil
+		}
+		return cr.Spec.VLStorage.PodMetadata
+	case vmv1beta1.ClusterComponentBalancer:
+		return cr.Spec.RequestsLoadBalancer.Spec.PodMetadata
+	default:
+		panic("BUG unsupported cluster kind=%s", string(kind))
```
vrutkovs, line 113: This makes my compiler unhappy:
```
invalid operation: too many arguments for panic("BUG unsupported cluster kind=%s", string(kind)) (expected 1, found 2)
```

AndrewChubatiuk: yes, i'm aware of it, fix is [here](https://github.com/VictoriaMetrics/operator/pull/1615)

PR #1614, opened 2025-11-11
config-reloader: set config reloader image equal to operator image by default

AndrewChubatiuk:
related PR https://github.com/VictoriaMetrics/helm-charts/pull/2562
- removed validation for container-reloader image as this doesn't take allow to use custom images
- use same version of config reloader if operator's buildinfo.ShortVersion is not empty and fallback to static version


f41gh7 (changes requested):
This change complicates local development and significantly slows development process. I was not able to run operator after this change.

 Possible solutions:
1) during release process, variable must be updated by maintainer accordingly to the release name. It should be reflected at release-guide doc. 
2) CI should modify source code and commit back changes. Possible security risks, I don't like it.

 I prefer 1 personally.

AndrewChubatiuk:
updated PR, for release version is set to `buildinfo.ShortVersion()`, for test builds it will be equal to static version

f41gh7 (approved):
LGTM

PR #1616, opened 2025-11-11
reconcile: treat more errors as expanding

AndrewChubatiuk:
- removed quotes from custom reconcile error messages, which makes them unreadable
- treat more errors as CR's expanding state: k8serrors.IsConflict, isErrorWaitTimeout(err) and isRecreate(err)

### DIFF
--- a/internal/controller/operator/controllers.go
+++ b/internal/controller/operator/controllers.go
@@ -365,14 +365,21 @@ func reconcileAndTrackStatus[T client.Object, ST reconcile.StatusWithMetadata[ST
 		resultErr = fmt.Errorf("cannot parse exist spec changes")
 		return
 	}
-	var diffPatch client.Patch
-	if specChanged {
-		diffPatch, err = object.LastAppliedSpecAsPatch()
-		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/controller/operator/controllers.go
```diff
@@ -365,22 +365,29 @@ func reconcileAndTrackStatus[T client.Object, ST reconcile.StatusWithMetadata[ST
...
 			return
 		}
+	}()
 
+	if specChanged {
 		if err := reconcile.UpdateObjectStatus(ctx, c, object, vmv1beta1.UpdateStatusExpanding, nil); err != nil {
 			resultErr = fmt.Errorf("failed to update object status: %w", err)
 			return
 		}
+		newPatch, err := object.LastAppliedSpecAsPatch()
```
f41gh7, line 378: call above to the `reconcile.UpdateObjectStatus` will update `object` with the latest state from Kubernetes API, which may have new field values. That's why `diffPatch, err = object.LastAppliedSpecAsPatch()` is called earlier.

AndrewChubatiuk: thanks for review, updated

PR #1622, opened 2025-11-16
docs: fixed links to old vm versions

AndrewChubatiuk:


AndrewChubatiuk:
docs.victoriametrics.com doesn't contain VM changelog before version 1.42.0
image

Issue #1625, opened 2025-11-17
feat: Add Gateway API support for vmauth


hkmdxlftjf:
Hello, I would like to add support for Gateway API in vmauth, so that it can recognize and work with Gateway and HTTPRoute resources, similar to how it currently handles Ingress. 

I am interested in implementing this feature and contributing the changes to vmauth.

makasim:
@f41gh7 @Haleygo @AndrewChubatiuk wdyt ?

AndrewChubatiuk:
suppose it's a good feature, but not sure we need to manage gateway resource as well, httproute should be enough

hkmdxlftjf:
> suppose it's a good feature, but not sure we need to manage gateway resource as well, httproute should be enough

Okay, I will only add management for httproute.

f41gh7:
In my understanding, it only makes sense to introduce simplified version of `HTTPRoute`.

 Something like:

```
kind: VMAuth
spec:
  httpRoute:
    parentRefs:
    - name: my-gateway
    hostnames:
    - my-vmauth.example.com    
```


 Operator will render it and attach single rules targeting vmauth.

AndrewChubatiuk:
HTTPRoute support for VMAuth was added in operator [v0.66.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.66.0) and released in helm chart [0.57.0](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-operator-0.57.0)

PR #1628, opened 2025-11-20
feat: Add httproute support for vmauth.

hkmdxlftjf:
feat: https://github.com/VictoriaMetrics/operator/issues/1625

Hello,

I’ve extended the vmspec.spec to include an httpRoute definition, for example:

```
kind: VMAuth
metadata:
  name: vmauth-sample
spec:
  selectAllByDefault: true
  httpRoute:
    spec:
      parentRefs:
      - group: gateway.networking.k8s.io
        kind: Gateway
        name: eg
      rules:
      - matches:
        - path:
            type: PathPrefix
            value: /
```

This configuration creates an HTTPRoute resource that binds the VMAuth instance to the specified Gateway.

Additionally, the httpRoute spec looks like this:

```
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  creationTimestamp: "2025-11-20T03:55:34Z"
  finalizers:
  - apps.victoriametrics.com/finalizer
  generation: 1
  labels:
    app.kubernetes.io/component: monitoring
    app.kubernetes.io/instance: vmauth-sample
    app.kubernetes.io/name: vmauth
    managed-by: vm-operator
  name: vmauth-vmauth-sample
  namespace: default
  ownerReferences:
  - apiVersion: operator.victoriametrics.com/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: VMAuth
    name: vmauth-sample
    uid: fe949e6b-7f59-4115-8d44-0e2af8192779
  resourceVersion: "8185085"
  uid: f25fa425-b16c-4004-af8f-c87935a8173b
spec:
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: eg
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: vmauth-vmauth-sample
      port: 8427
      weight: 1
    matches:
    - path:
        type: PathPrefix
        value: /
```


internal/controller/operator/factory/vmauth/vmauth.go
```diff
@@ -117,6 +123,26 @@ func CreateOrUpdate(ctx context.Context, cr *vmv1beta1.VMAuth, rclient client.Cl
...
+		prevHTTPRoute, err = build.HTTPRoute(cr, cr.Spec.Port, prevCr.Spec.HTTPRoute)
+		if err != nil {
+			return err
+		}
+	}
+
+	return reconcile.HTTPRoute(ctx, rclient, newHTTPRoute, prevHTTPRoute)
+}
+
 func newDeployForVMAuth(cr *vmv1beta1.VMAuth) (*appsv1.Deployment, error) {
```
AndrewChubatiuk, line 149: could you please also add removal of stale http route to `deletePrevStateResources`?

api/operator/v1beta1/vmauth_types.go
```diff
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
```
AndrewChubatiuk, line 17: please add this alias to .golangci.yml

AndrewChubatiuk:
overall looks good, please add unit (internal/controller/operator/factory/vmauth/vmauth_test.go) and e2e (test/e2e/vmauth_test.go) tests for this feature

api/operator/v1beta1/vmauth_types.go
hkmdxlftjf: I appreciate the review. This is now resolved.

internal/controller/operator/factory/vmauth/vmauth.go
hkmdxlftjf: I appreciate the review. This is now resolved.

hkmdxlftjf:
> overall looks good, please add unit (internal/controller/operator/factory/vmauth/vmauth_test.go) and e2e (test/e2e/vmauth_test.go) tests for this feature

Thanks! I've already added the tests.

hkmdxlftjf:
I'm not sure if this is an issue. When I add the following to vmauth_controller:

```
// +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=httproutes,verbs=get;list;watch;create;update;patch;delete
```

and then run `make manifests`, it doesn't seem to update anything under `config/rbac/`.
I also tried deleting all files under `config/rbac/*.yaml` and running make manifests again, but the result is the same.

AndrewChubatiuk:
> and then run make manifests, it doesn't seem to update anything under config/rbac/.

it doesn't add manifests to config/rbac/, please add it manually
also operator should check if gateway api CRDs are installed in a cluster to allow using this functionality, they should not be provided by VM operator

hkmdxlftjf:
> > and then run make manifests, it doesn't seem to update anything under config/rbac/.
> 
> it doesn't add manifests to config/rbac/, please add it manually also operator should check if gateway api CRDs are installed in a cluster to allow using this functionality, they should not be provided by VM operator

I have added the RBAC manifests to config/rbac/.

Regarding the Gateway API, I added a check for the HTTPRoute CRD. Currently, if the CRD is missing, the operator simply skips creating the HTTPRoute. However, should the operator update the Status to inform the user that their requested HTTPRoute was not created due to the missing CRD?

AndrewChubatiuk:
if CreateOrUpdate returns error it will automatically set status to Failed. Just need to write proper log message and text in error message

internal/controller/operator/factory/vmauth/vmauth.go
```diff
@@ -119,6 +125,34 @@ func CreateOrUpdate(ctx context.Context, cr *vmv1beta1.VMAuth, rclient client.Cl
 	return nil
 }
 
+func createOrUpdateHTTPRoute(ctx context.Context, rclient client.Client, cr, prevCr *vmv1beta1.VMAuth) error {
+	if cr.Spec.HTTPRoute == nil {
+		return nil
+	}
+
+	if err := k8stools.GetCRD(ctx, rclient, "httproutes.gateway.networking.k8s.io"); err != nil {
+		if k8serrors.IsNotFound(err) {
```
hkmdxlftjf: > if CreateOrUpdate returns error it will automatically set status to Failed. Just need to write proper log message and text in error message

@AndrewChubatiuk So if vmauth.spec.HTTPRoute is not nil but the HTTPRoute CRD is missing, we should return an error in this case, correct?

hkmdxlftjf:
Please let me know if there’s anything else I should update.

internal/controller/operator/factory/vmauth/vmauth.go
```diff
@@ -131,9 +131,6 @@ func createOrUpdateHTTPRoute(ctx context.Context, rclient client.Client, cr, pre
 	}
 
 	if err := k8stools.GetCRD(ctx, rclient, "httproutes.gateway.networking.k8s.io"); err != nil {
-		if k8serrors.IsNotFound(err) {
-			return nil
-		}
 		return err
```
AndrewChubatiuk: could you please wrap error with advice to install Gateway CRDs

hkmdxlftjf: ok

config/manager/kustomization.yaml
```diff
@@ -1,2 +1,3 @@
 resources:
 - manager.yaml
+- ../rbac
```
AndrewChubatiuk:
```suggestion
```
it's not needed here

hkmdxlftjf: However, when running make deploy, the RBAC resources are not created, so the operator’s ReplicaSet will not be able to create any Pods.

```
 Warning  FailedCreate  5s (x13 over 26s)  replicaset-controller  Error creating: pods "operator-74fd76f858-" is forbidden: error looking up service account vm/operator: serviceaccount "operator" not found
 ```

Comment (AndrewChubatiuk) [reply config/manager/kustomization.yaml]: try `OVERLAY=config/base-with-webhook make deploy` instead
usually I'm using `make deploy-kind`, which spins up local kind cluster, builds operator and deploys it there

**Comment by wilmardo** [comment]
Great work! I am interested in this feature as well so thanks for the effort <3

Quick points I noticed during review:
* It needs a reference in the docs for the [VMAuthSpec](https://github.com/hkmdxlftjf/operator/blob/67b790e5bb14caaf6825be0c4ecac332f04e8b11/docs/api.md#vmauthspec-ingress)
* In the above docs the `EmbeddedIngress` is currently marked as `required` I am not sure if this is enforced in the code somewhere (needs checking). But for sure the docs needs to be updated that Ingress is optional when an HTTPRoute is present. Or make both optional

**Comment by hkmdxlftjf** [comment]
Thank you for the feedback! 

> * It needs a reference in the docs for the [VMAuthSpec](https://github.com/hkmdxlftjf/operator/blob/67b790e5bb14caaf6825be0c4ecac332f04e8b11/docs/api.md#vmauthspec-ingress)

I’ll add the missing reference to VMAuthSpec in the documentation.

> * In the above docs the `EmbeddedIngress` is currently marked as `required` I am not sure if this is enforced in the code somewhere (needs checking). But for sure the docs needs to be updated that Ingress is optional when an HTTPRoute is present. Or make both optional

It looks like this is a documentation mistake. @AndrewChubatiuk Should I update it in this PR?

**Comment by hkmdxlftjf** [comment]
> Great work! I am interested in this feature as well so thanks for the effort <3
> 
> Quick points I noticed during review:
> 
> * It needs a reference in the docs for the [VMAuthSpec](https://github.com/hkmdxlftjf/operator/blob/67b790e5bb14caaf6825be0c4ecac332f04e8b11/docs/api.md#vmauthspec-ingress)
> * In the above docs the `EmbeddedIngress` is currently marked as `required` I am not sure if this is enforced in the code somewhere (needs checking). But for sure the docs needs to be updated that Ingress is optional when an HTTPRoute is present. Or make both optional


Thank you for the feedback! 

> * It needs a reference in the docs for the [VMAuthSpec](https://github.com/hkmdxlftjf/operator/blob/67b790e5bb14caaf6825be0c4ecac332f04e8b11/docs/api.md#vmauthspec-ingress)

I’ll add the missing reference to VMAuthSpec in the documentation.

> * In the above docs the `EmbeddedIngress` is currently marked as `required` I am not sure if this is enforced in the code somewhere (needs checking). But for sure the docs needs to be updated that Ingress is optional when an HTTPRoute is present. Or make both optional

It looks like this is a documentation mistake. @AndrewChubatiuk Should I update it in this PR?

**Comment by AndrewChubatiuk** [comment]
> It looks like this is a documentation mistake. @AndrewChubatiuk Should I update it in this PR?

right, looks like a documentation issue. no need to do it

**Comment by hkmdxlftjf** [comment]
Hi, may I ask if this is ready to be merged? 
Please let me know if anything else is needed from my side.

**Comment by AndrewChubatiuk** [comment]
@Haleygo @f41gh7 @vrutkovs 
could you please take a look as well?

File: internal/controller/operator/factory/build/httproute.go [hunk 6]
```diff
@@ -0,0 +1,53 @@
...
+// HTTPRoute creates HTTPRoute object
+func HTTPRoute(cr builderOpts, port string, httpRoute *vmv1beta1.EmbeddedHTTPRoute) (*gwapiv1.HTTPRoute, error) {
+	var (
+		defaultPort gwapiv1.PortNumber
+	)
+	if defaultPortVal, err := strconv.Atoi(port); err != nil {
+		return nil, err
+	} else {
+		defaultPort = gwapiv1.PortNumber(defaultPortVal)
+	}
```
vrutkovs:
```suggestion
	defaultPortVal, err := strconv.Atoi(port)
	if err != nil {
		return nil, err
	}
	defaultPort = gwapiv1.PortNumber(defaultPortVal)
```
nitpick: IMO this makes the flow clear

internal/controller/operator/factory/build/httproute.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	}
+
+	lbls := labels.Merge(httpRoute.Labels, cr.SelectorLabels())
+	spec := httpRoute.Spec
+	for i := range spec.Rules {
+		if spec.Rules[i].BackendRefs == nil {
+			spec.Rules[i].BackendRefs = []gwapiv1.HTTPBackendRef{}
+		}
+		spec.Rules[i].BackendRefs = append(spec.Rules[i].BackendRefs, defaultBackendRef)
```
vrutkovs: Do we need to turn this into a set first, so that `defaultBackendRef` would not appear multiple times?

internal/controller/operator/factory/reconcile/httproute.go
```diff
@@ -0,0 +1,51 @@
...
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	"github.com/VictoriaMetrics/operator/internal/controller/operator/factory/finalize"
+	"github.com/VictoriaMetrics/operator/internal/controller/operator/factory/logger"
+)
+
+// HTTPRoute creates or update HTTPRoute object
```
vrutkovs:
```suggestion
// HTTPRoute creates or updates HTTPRoute object
```

internal/controller/operator/factory/reconcile/httproute.go
```diff
@@ -0,0 +1,51 @@
...
+func HTTPRoute(ctx context.Context, rclient client.Client, newHTTPRoute, prevHTTPRoute *gwapiv1.HTTPRoute) error {
+	return retryOnConflict(func() error {
+		var curHTTPRoute gwapiv1.HTTPRoute
+		if err := rclient.Get(ctx, types.NamespacedName{Name: newHTTPRoute.GetName(), Namespace: newHTTPRoute.GetNamespace()}, &curHTTPRoute); err != nil {
+			if k8serrors.IsNotFound(err) {
+				logger.WithContext(ctx).Info(fmt.Sprintf("creating HTTPRoute %s configuration", newHTTPRoute.Name))
+				return rclient.Create(ctx, newHTTPRoute)
+			}
+			return fmt.Errorf("cannot get exist HTTPRoute object: %w", err)
```
vrutkovs:
```suggestion
			return fmt.Errorf("cannot get existing HTTPRoute object: %w", err)
```

internal/controller/operator/factory/reconcile/httproute.go
```diff
@@ -0,0 +1,51 @@
...
+			return &errRecreate{
+				origin: fmt.Errorf("waiting for HTTPRoute %q to be removed", newHTTPRoute.Name),
+			}
+		}
+		if err := finalize.FreeIfNeeded(ctx, rclient, &curHTTPRoute); err != nil {
+			return err
+		}
+
+		if equality.Semantic.DeepEqual(newHTTPRoute.Spec, curHTTPRoute.Spec) &&
+			equality.Semantic.DeepEqual(newHTTPRoute.Labels, curHTTPRoute.Labels) &&
```
vrutkovs: Do we need this - `isObjectMetaEqual` would also check the labels?

internal/controller/operator/factory/build/httproute.go
hkmdxlftjf: Yes, duplication is indeed a problem here. When vmauth.spec.httproute.spec.Rules[i].BackendRefs is not empty, this code can cause defaultBackendRef to appear multiple times.

My idea is to simplify the logic to:

```
spec := httpRoute.Spec
for i := range spec.Rules {
    spec.Rules[i].BackendRefs = []gwapiv1.HTTPBackendRef{defaultBackendRef}
}
```


This ensures that each rule always has a single, deterministic BackendRef, avoiding duplicates entirely.

Do you have a better suggestion or preferred pattern here?

internal/controller/operator/factory/reconcile/httproute.go
hkmdxlftjf: Yes, this is redundant — isObjectMetaEqual already checks the labels. I’ll remove the extra comparison.

api/operator/v1beta1/vmauth_types.go
```diff
@@ -469,6 +472,15 @@ func (cr *VMAuthSpec) UnmarshalJSON(src []byte) error {
 	return nil
 }
 
+// EmbeddedHTTPRoute describes httproute configuration options.
+type EmbeddedHTTPRoute struct {
+	//  EmbeddedObjectMetadata adds labels and annotations for object.
+	EmbeddedObjectMetadata `json:",inline"`
+	// Spec describes the attributes that a user creates on a httproute.
+	// More info: https://gateway-api.sigs.k8s.io/api-types/httproute/
+	Spec gwapiv1.HTTPRouteSpec `json:"spec"`
```
f41gh7: It'd be great to switch it into schemaless. `HTTPRouteSpec` adds too much yaml lines into the CRD.

internal/controller/operator/factory/build/httproute.go
```diff
@@ -0,0 +1,50 @@
...
+				Name: gwapiv1.ObjectName(cr.PrefixedName()),
+				Port: &defaultPort,
+			},
+		},
+	}
+
+	lbls := labels.Merge(httpRoute.Labels, cr.SelectorLabels())
+	spec := httpRoute.Spec
+	for i := range spec.Rules {
+		if spec.Rules[i].BackendRefs == nil {
```
f41gh7: I think, operator should only fill required fields and simplify configuration management. 

 There is no need to embed whole HTTPRoute struct, it has a lot of different rare used options.

 The only really useful fields are:
* `parentRefs`
* `Hostnames`

 Rules with default backend could be added by operator. 

 As an alternative, we could also add `extraRules` with user defined unmanaged rules.

internal/controller/operator/factory/build/httproute.go
vrutkovs: Yeah, that would work. Sometimes we use [this](https://github.com/VictoriaMetrics/operator/blob/master/internal/controller/operator/factory/build/vmservicescrape.go#L127-L141) method, but it looks overcomplicated.

internal/controller/operator/factory/build/httproute.go
```diff
@@ -0,0 +1,95 @@
...
+						Path: &gwapiv1.HTTPPathMatch{
+							Type:  ptr.To(gwapiv1.PathMatchPathPrefix),
+							Value: ptr.To("/"),
+						},
+					},
+				},
+				BackendRefs: refs,
+			},
+		}, nil
+	}
```
vrutkovs, line 90:
```suggestion
	var rules []gwapiv1.HTTPRouteRule
	if httpRoute.ExtraRules != nil {
		var r gwapiv1.HTTPRouteRule
		for _, ext := range httpRoute.ExtraRules {
			if len(ext.Raw) == 0 {
				continue
			}
			if err := json.Unmarshal(ext.Raw, &r); err != nil {
				return nil, fmt.Errorf("failed to decode extraRule: %w", err)
			}
			r.BackendRefs = append([]gwapiv1.HTTPBackendRef{}, refs...)
			rules = append(rules, r)
		}
	} else {
		rules = []gwapiv1.HTTPRouteRule{
			{
				Matches: []gwapiv1.HTTPRouteMatch{
					{
						Path: &gwapiv1.HTTPPathMatch{
							Type:  ptr.To(gwapiv1.PathMatchPathPrefix),
							Value: ptr.To("/"),
						},
					},
				},
				BackendRefs: refs,
			},
		}
	}
	return rules, nil
```

Lets not allocate in the loop and ensure only one `return` is used

internal/controller/operator/factory/reconcile/httproute.go
```diff
@@ -0,0 +1,51 @@
...
```
vrutkovs:
```suggestion
```

internal/controller/operator/factory/vmauth/vmauth.go
```diff
@@ -119,6 +125,36 @@ func CreateOrUpdate(ctx context.Context, cr *vmv1beta1.VMAuth, rclient client.Cl
 	return nil
 }
 
+func createOrUpdateHTTPRoute(ctx context.Context, rclient client.Client, cr, prevCr *vmv1beta1.VMAuth) error {
+	if cr.Spec.HTTPRoute == nil {
+		return nil
+	}
+
+	if err := k8stools.GetCRD(ctx, rclient, "httproutes.gateway.networking.k8s.io"); err != nil {
```
f41gh7: I think, it's fine to skip this check and produce runtime error during httpRoute create request. GetCRD request requires extra RBAC access and cannot be used with per namespace mode, so it's better to do not execute it for simplicity.

f41gh7:
Thanks for contribution!

PR #1630, opened 2025-11-20
vol 3: extracted subset of changes from #1564

AndrewChubatiuk:
extracted another portion of changes from https://github.com/VictoriaMetrics/operator/pull/1564
- removed impact of previous CR state for all non-cluster CRs
- check finalizer and ownerreference during resource removal


f41gh7:
Thanks for contribution!

PR #1635, opened 2025-11-24
ci: added openshift certified operators repo to release job

AndrewChubatiuk:
added openshift certified operators repo to release job


vrutkovs:
iiuc our operator is not yet conformant with Red Hat Certified Operators collection, but once we update it this PR would be very useful. Lets keep it open for some time

Makefile
```diff
@@ -256,19 +256,21 @@ build-installer: manifests generate kustomize ## Generate a consolidated YAML wi
 	$(KUSTOMIZE) build config/crd/overlay > dist/crd.yaml
 
 olm: operator-sdk opm yq docs
+	$(eval DIGEST = $(shell $(CONTAINER_TOOL) buildx imagetools inspect $(REGISTRY)/$(ORG)/$(REPO):$(TAG)-ubi --format "{{print .Manifest.Digest}}"))
```
vrutkovs, line 259: We should make sure it returns the manifest list digest instead of arch-specific manifest

AndrewChubatiuk: it does return list digest, you can recheck if you want

PR #1373, opened 2025-05-21
Change Error log to Info for PVC size increase

bertinm-gc:
addresses #1361

Issue #1191, opened 2024-12-17
Inconsistent logging format


emilbillberg:
Hi,

I am running victoria metrics operator v0.50.0; and I have noticed the log format can be a bit inconsistent.

I have noticed these log fields:

- namespace
- controller
- desiredVersion
- podMustRecreate
- sts.name
- vmcluster
- vmalert
- vmalertmanager
- rules
- secret_for
- len
- ' len ' (contains spaces)
- 'map names'
- 'invalid rules'
- 'invalid configs'
- 'desired version'

Some of them perhaps should stay, some of them should most likely be merged with the `msg` field instead. The spaces should also be removed from the keys as log collectors usually add a `_` or `.` and it can be a bit weird.

I would expect `msg`, `level` and `ts` to exist as well as a few other consistent log fields which helps to understand the context of the log.

PR #1636, opened 2025-11-25
Fix odd number of key-value pairs in info log about pvc size difference

plkokanov:
After https://github.com/VictoriaMetrics/operator/pull/1373, which changed the log severity from `Error` to `Info` of the log describing that a PVC cannot be decreased, because it already has a larger size than what is currently configured, the following error can be observed, because the `Info()` call did not contain an even number of key-value pairs after the main log message: 

```
{"level":"info","ts":"2025-11-25T12:00:46Z","logger":"controller.VLSingle","msg":"cannot decrease PVC=vlsingle-example size from=30Gi to=40Gi","vlsingle":"example","namespace":"default"}
{"level":"dpanic","ts":"2025-11-25T12:01:47Z","logger":"controller.VLSingle","msg":"odd number of arguments passed as key-value pairs for logging","vlsingle":"example","namespace":"default","ignored key":"cannot decrease PVC size, please check VolumeClaimTemplate configuration"}
```

This PR moves the extra string `"please check VolumeClaimTemplate configuration"` to the `fmt.Srpintf(...)` call and removes the unnecessary `cannot decrease PVC size` text as it would just be duplicated.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -21,6 +21,7 @@ aliases:
 * FEATURE: [vmalertmanagerconfig](https://docs.victoriametrics.com/operator/resources/vmalertmanagerconfig/): support incident.io receiver. See [#1637](https://github.com/VictoriaMetrics/operator/issues/1637).
 
 * BUGFIX: [vmalertmanager](https://docs.victoriametrics.com/operator/resources/vmalertmanager/): check `mute_time_intervals` in subroutes: See [#1618](https://github.com/VictoriaMetrics/operator/issues/1618).
+* BUGFIX: [vmoperator](https://docs.victoriametrics.com/operator/): remove incorrect key argument in structured log for when the actual PVC storage size is larger than the currently configured size and properly indicate which is the new and which is the existing size: See PR [#1636](https://github.com/VictoriaMetrics/operator/pull/1636) for details.
 
 * Dependency: [vmoperator](https://docs.victoriametrics.com/operator/): Updated default versions for VT apps to [v0.5.0](https://github.com/VictoriaMetrics/VictoriaTraces/releases/tag/v0.5.0) version.
 


internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
```diff
@@ -222,8 +222,7 @@ func mayGrow(ctx context.Context, pvc *corev1.PersistentVolumeClaim, newSize, ex
 		// do no return error
 		// probably, user updated pvc manually
 		// without applying this changes to the configuration.
-		logger.WithContext(ctx).Info(fmt.Sprintf("cannot decrease PVC=%s size from=%s to=%s", pvc.Name, newSize.String(), existSize.String()),
-			"cannot decrease PVC size, please check VolumeClaimTemplate configuration")
+		logger.WithContext(ctx).Info(fmt.Sprintf("cannot decrease PVC=%s size from=%s to=%s, please check VolumeClaimTemplate configuration", pvc.Name, newSize.String(), existSize.String()))
```
AndrewChubatiuk: what do you think about doing it like this?
```suggestion
		logger.WithContext(ctx).Info("cannot decrease PVC size, please check VolumeClaimTemplate configuration", "pvc", pvc.Name, "from", existSize.String(), "to", newSize.String())
```

plkokanov: Thanks for the suggestion, I would also prefer it that way. 
Just saw that other places were still using `fmt.Sprintf()` for structured logs 😅

plkokanov: done

plkokanov:
Added the changelog entry. Hopefully, it's the right place and format (this is my first PR here 😊 )

internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
```diff
@@ -222,8 +222,7 @@ func mayGrow(ctx context.Context, pvc *corev1.PersistentVolumeClaim, newSize, ex
 		// do no return error
 		// probably, user updated pvc manually
 		// without applying this changes to the configuration.
-		logger.WithContext(ctx).Info(fmt.Sprintf("cannot decrease PVC=%s size from=%s to=%s", pvc.Name, newSize.String(), existSize.String()),
-			"cannot decrease PVC size, please check VolumeClaimTemplate configuration")
+		logger.WithContext(ctx).Info("cannot decrease PVC size, please check VolumeClaimTemplate configuration", "pvc", pvc.Name, "from", existSize.String(), "to", newSize.String())
```
f41gh7: It's better to use `fmt.Sprintf` instead of each field logging.

The problem with per field logging is read-ability.  Log line:
`cannot decrease PVC size, please check VolumeClaimTemplate configuration` doesn't have any context with out sizes and PVC name. It also adds an additional pressure on remote logging systems, since each field needs an additional index.

Latter we're going to replace logger implementation.

 See this issue for more context https://github.com/VictoriaMetrics/operator/issues/1191

f41gh7: I suggest to apply the following fix:

```
logger.WithContext(ctx).Info(fmt.Sprintf("cannot decrease PVC=%s size from=%s to=%s, please check VolumeClaimTemplate configuration", pvc.Name,existSize.String(), newSize.String()))
```

AndrewChubatiuk: looks like `existSize` should be mapped into `from` and `newSize` into `to`

f41gh7: Good point! Updated

plkokanov: Applied the change, I also modified the changelog a bit to indicate that `from` now points to `existSize` and `to` to `newSize`

AndrewChubatiuk: Merged. Thanks for contribution

PR #1644, opened 2025-11-28
docs/vmagent: update examples for relabeling

zekker6:
Clarify difference between global remote write and service discovery relabelings, add relevant links to relabeling docs.

Added an example of using global service discovery relabeling for multitenancy.

### DIFF
--- a/docs/resources/vmagent.md
+++ b/docs/resources/vmagent.md
@@ -378,7 +378,7 @@ spec:
 
 ## Relabeling
 
-`VMAgent` supports global [service discovery relabeling](https://docs.victoriametrics.com/victoriametrics/relabeling/#service-discovery-relabeling), global remote write relabeling for all metrics and per remoteWrite target relabel config.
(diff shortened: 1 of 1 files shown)


docs/resources/vmagent.md
```diff
@@ -424,7 +425,7 @@ data:
       source_labels: [foo, bar]
 ```
 
-Second, add `relabelConfig` to `VMagent` spec for global relabeling with name of `Configmap` - `vmagent-relabel` and key `global-relabel.yaml`.
+Second, add `relabelConfig` to `VMagent` spec for global remote write relabeling with name of `Configmap` - `vmagent-relabel` and key `remote-write-relabel.yaml`.
```
AndrewChubatiuk: reading it for the first time had an impression about some `global remote write`, which is not the same as `global for all remote writes`

zekker6: Updated wording a bit to use same phrase when referring to global relabeling config, ptal

Issue #1655, opened 2025-12-01
VMAuth fails with "no matches for kind "HTTPRoute" on latest master


vrutkovs:
#1628 caused VMAuth to fail on clusters where Gateway API CRDs are not installed to fail with:
```
spec:
  image:
    tag: v1.130.0
  license: {}
  port: "8427"
  unauthorizedUserAccessSpec:
    discover_backend_ips: true
    url_map:
    - src_paths:
      - /select/.*
      url_prefix:
      - http://vmselect-vmks-test.vm.svc.cluster.local/.:8481/
    - src_paths:
      - /insert/.*
      url_prefix:
      - http://vminsert-vmks-test.vm.svc.cluster.local/.:8480/
status:
  observedGeneration: 1
  reason: 'cannot create or update vmauth deploy: cannot delete httproute from prev
    state: no matches for kind "HTTPRoute" in version "gateway.networking.k8s.io/v1"'
  updateStatus: failed
```

f41gh7:
As far as I understand, `gateway` API must be a part of default Kubernetes API for any up-to-date Kubernetes cluster. Am I wrong?

AndrewChubatiuk:
nope, unfortunately it's not true

AndrewChubatiuk:
just closing this, as it wasn't present in any release

PR #1656, opened 2025-12-01
check CRDs during operator's startup

AndrewChubatiuk:
check CRDs during operator startup
fixes https://github.com/VictoriaMetrics/operator/issues/1655


AndrewChubatiuk:
okay, waiting for your approach

vrutkovs:
>CRDs must be per-installed according to the requirements of operator and Kubernetes version compatibility.

We don't require this CRD to be present though, its support is optional. If we _do_ require it to be present this is a significant change

PR #1657, opened 2025-12-01
add `VM_GATEWAY_API_ENABLED` global parameter for enabling managing Gateway API resources

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1655


vrutkovs:
This would still make (faulty) requests to kubeapiserver, although no CRD is available?

AndrewChubatiuk:
it does, added toggle to explicitly enable gateway API support

vrutkovs:
That sounds much better to me, thanks!

I didn't check that it works here though

f41gh7:
Thanks for contribution!

Issue #1665, opened 2025-12-04
VMCluster: orphaned ServiceAccount when switching from default to custom serviceAccountName


func25:
When a `VMCluster` is created without `spec.serviceAccountName`, the operator auto-creates and owns a `ServiceAccount`. 

- If the same `VMCluster` is later updated to set a custom `spec.serviceAccountName`, the operator switches to using the new `ServiceAccount`, though the originally auto-created `ServiceAccount` is not cleaned up
- On `VMCluster` deletion, `deleteSA` is not called because `IsOwnsServiceAccount()` is now `false`, so that old `ServiceAccount` stays around with the operator finalizer still attached.

This orphaned `ServiceAccount` can block namespace deletion. It would be helpful if the operator removed or freed the previously auto-created `ServiceAccount` when switching to a custom `serviceAccountName`, or at least during `VMCluster` deletion.

AndrewChubatiuk:
fix was included into operator release v0.66.1

PR #1666, opened 2025-12-04
cleanup orphaned roles, bindings and service accounts

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1665


f41gh7:
Thanks for contribution!

PR #1671, opened 2025-12-05
controller: add object_namespace label to bad configs metric

f41gh7:
This commit adds object_namespace label to the metrics:
* operator_alertmanager_bad_objects_count
* operator_vmalert_bad_objects_count

 It helps to route alerts on object_namespace label.

### DIFF
--- a/internal/controller/operator/factory/vmalert/rules.go
+++ b/internal/controller/operator/factory/vmalert/rules.go
@@ -227,6 +227,7 @@ func selectRulesContent(ctx context.Context, rclient client.Client, cr *vmv1beta
 		vmRules = deduplicateRules(ctx, vmRules)
 	}
 	var brokenRulesByNamespace map[string]int
+	var brokenRulesTotal int
 	for _, pRule := range vmRules {
 		if !build.MustSkipRuntimeValidation {
 			if err := pRule.Validate(); err != nil {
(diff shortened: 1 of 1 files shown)


AndrewChubatiuk:
Do we have the same for VMagent child objects?

f41gh7:
> Do we have the same for VMagent child objects?

There is no such metrics for vmagent and vmauth objects. Probably, we could add it. Can you please create an issue for that?

AndrewChubatiuk:
And what do you think about moving parent component name to labels and change metric name to operator_bad_objects_count?

internal/controller/operator/factory/vmalert/rules.go
```diff
@@ -26,13 +26,16 @@ import (
...
-	Name: "operator_vmalert_bad_objects_count",
-	Help: "Number of incorrect objects by controller",
-	ConstLabels: prometheus.Labels{
-		"controller": "vmrules",
+var badConfigsTotal = prometheus.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "operator_vmalert_bad_objects_count",
+		Help: "Number of incorrect objects by controller",
+		ConstLabels: prometheus.Labels{
+			"controller": "vmrules",
```
AndrewChubatiuk, line 34: controller can be either vmrules or vmalerts, it's not static

AndrewChubatiuk: please check PR https://github.com/VictoriaMetrics/operator/pull/1676

f41gh7: yes, it's better to remove controller label.

vrutkovs:
Closed in favor of #1671

Issue #1501, opened 2025-08-01
Support DaemonSet mode for vlagent CRD


vadimalekseev:
This is required for log collection in a Kubernetes cluster. For more details, see: https://github.com/VictoriaMetrics/VictoriaLogs/issues/538

AndrewChubatiuk:
VLAgent supports [k8s logs collection](https://docs.victoriametrics.com/operator/resources/vlagent/#k8s-logs-collection) starting [0.67.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.67.0).

PR #1675, opened 2025-12-05
vlagent: support logs collection

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1501


api/operator/v1/vlagent_types.go
```diff
@@ -82,6 +85,29 @@ type VLAgentSpec struct {
 	vmv1beta1.CommonApplicationDeploymentParams `json:",inline,omitempty"`
 }
 
+type VLAgentK8sCollector struct {
+	// Enabled enables k8s logs collection
+	Enabled bool `json:"enabled,omitempty"`
```
f41gh7, line 92: It's better to add context to the `Enabled` param. That it changes installations type of vlagent from statefulset and daemonset.

 It also erases on-disk state of vlagent and loses already buffered data.

api/operator/v1/vlagent_types.go
```diff
@@ -82,6 +85,29 @@ type VLAgentSpec struct {
 	vmv1beta1.CommonApplicationDeploymentParams `json:",inline,omitempty"`
 }
 
+type VLAgentK8sCollector struct {
```
f41gh7, line 88: I think, it'd be useful to have `checkpointsPath` and `logsPath` options.

 By default, `checkpointsPath` should have value of `tmpDataDir` for persistence?

internal/controller/operator/factory/vlagent/rbac.go
```diff
@@ -0,0 +1,102 @@
...
+				"namespaces",
+				"nodes",
+			},
+		},
+	}
+)
+
+// createK8sAPIAccess - creates RBAC access rules for vlagent
+func createK8sAPIAccess(ctx context.Context, rclient client.Client, cr, prevCR *vmv1.VLAgent) error {
+	if err := ensureCRExist(ctx, rclient, cr, prevCR); err != nil {
```
f41gh7, line 40: there should be a check for single namespace mode. In this scenario, permissions must be provided manually.

api/operator/v1/vlagent_types.go
```diff
@@ -82,6 +85,29 @@ type VLAgentSpec struct {
...
+	DecolorizeFields []string `json:"decolorizeFields,omitempty"`
+
+	// MsgField defines fields that may contain the _msg field
+	MsgFields []string `json:"msgFields,omitempty"`
+
+	// TimeFields defines fields that may contain the _time field
+	TimeFields []string `json:"timeFields,omitempty"`
+
+	// ExtraFields defines extra fields to add to each collected log line
+	ExtraFields []string `json:"extraFields,omitempty"`
```
f41gh7: extraFields is not an array, it's json encoded string object.

internal/controller/operator/factory/vlagent/vlagent.go
```diff
@@ -210,14 +271,77 @@ func makeSpec(cr *vmv1.VLAgent) (*corev1.PodSpec, error) {
...
+			args = append(args, fmt.Sprintf("-kubernetesCollector.decolorizeFields=%q", strings.Join(cr.Spec.K8sCollector.DecolorizeFields, ",")))
+		}
+		if len(cr.Spec.K8sCollector.MsgFields) > 0 {
+			args = append(args, fmt.Sprintf("-kubernetesCollector.msgField=%q", strings.Join(cr.Spec.K8sCollector.MsgFields, ",")))
+		}
+		if len(cr.Spec.K8sCollector.TimeFields) > 0 {
+			args = append(args, fmt.Sprintf("-kubernetesCollector.timeField=%q", strings.Join(cr.Spec.K8sCollector.TimeFields, ",")))
+		}
+		if len(cr.Spec.K8sCollector.ExtraFields) > 0 {
+			args = append(args, fmt.Sprintf("-kubernetesCollector.extraField=%q", strings.Join(cr.Spec.K8sCollector.ExtraFields, ",")))
```
f41gh7: No need to join

internal/controller/operator/factory/vlagent/vlagent.go
```diff
@@ -210,14 +271,77 @@ func makeSpec(cr *vmv1.VLAgent) (*corev1.PodSpec, error) {
...
+			Name: "varlib",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/var/lib",
+				},
+			},
+		}, corev1.Volume{
+			Name: "vl-collector-data",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
```
f41gh7: Requires extra permissions and should be configurable.

internal/controller/operator/factory/vlagent/vlagent.go
```diff
@@ -210,14 +271,77 @@ func makeSpec(cr *vmv1.VLAgent) (*corev1.PodSpec, error) {
...
+					Path: "/var/lib/vl-collector",
+				},
+			},
+		})
+		if cr.Spec.RemoteWriteSettings == nil || cr.Spec.RemoteWriteSettings.TmpDataPath == nil {
+			volumes = append(volumes, corev1.Volume{
+				Name: vlAgentPersistentQueueMountName,
+				VolumeSource: corev1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{},
```
f41gh7, line 348: What is the purpose of this mount if operator unconditionally adds `vl-collector-data`?

internal/controller/operator/factory/vlagent/vlagent.go
```diff
@@ -569,6 +693,28 @@ func deleteOrphaned(ctx context.Context, rclient client.Client, cr *vmv1.VLAgent
...
+			return fmt.Errorf("cannot remove statefulset: %w", err)
+		}
+	} else {
+		if err := finalize.SafeDeleteWithFinalizer(ctx, rclient, &appsv1.DaemonSet{ObjectMeta: objMeta}, &owner); err != nil {
+			return fmt.Errorf("cannot remove daemonset: %w", err)
+		}
+	}
+	if !cr.IsOwnsServiceAccount() || !cr.Spec.K8sCollector.Enabled {
+		rbacMeta := metav1.ObjectMeta{Name: cr.GetClusterRoleName(), Namespace: cr.Namespace}
+		objects := []client.Object{
```
f41gh7, line 721: operator permissions check.

internal/controller/operator/factory/vlagent/vlagent.go
AndrewChubatiuk: replaced with emptydir in a default configuration

internal/controller/operator/factory/vlagent/vlagent.go
AndrewChubatiuk: added condition

internal/controller/operator/factory/vlagent/rbac.go
AndrewChubatiuk: added warning for limited namespaces mode

api/operator/v1/vlagent_types.go
AndrewChubatiuk: added `dataPath` (checkpointsPath is `${dataPath}/checkpoints.json`) and `logsPath`

api/operator/v1/vlagent_types.go
```diff
@@ -82,6 +85,39 @@ type VLAgentSpec struct {
...
+	// DecolorizeFields defines fields to remove ANSI color codes across logs ingested from Kubernetes
+	DecolorizeFields []string `json:"decolorizeFields,omitempty"`
+
+	// MsgField defines fields that may contain the _msg field
+	MsgFields []string `json:"msgFields,omitempty"`
+
+	// TimeFields defines fields that may contain the _time field
+	TimeFields []string `json:"timeFields,omitempty"`
+
+	// ExtraFields defines extra fields to add to each collected log line
```
f41gh7: Could you also please add an example from vlagent doc. And maybe we should add a validation for it at webhook. Since this field has some specific format.

internal/controller/operator/factory/vlagent/rbac.go
```diff
@@ -0,0 +1,114 @@
...
+			"namespaces",
+			"nodes",
+		},
+	}}
+)
+
+// createK8sAPIAccess - creates RBAC access rules for vlagent
+func createK8sAPIAccess(ctx context.Context, rclient client.Client, cr, prevCR *vmv1.VLAgent) error {
+	if !config.IsClusterWideAccessAllowed() {
+		logger.WithContext(ctx).Info(fmt.Sprintf("skipping cluster role and binding for vlagent=%s/%s since operator has WATCH_NAMESPACE set", cr.Namespace, cr.Name))
```
f41gh7, line 37: Could you please add check for `IsOwnsServiceAccount`? If serviceAccountName is set by user, there is no need at log message.

 If it's not set, we could provide a hint at log message.

api/operator/v1/vlagent_types.go
AndrewChubatiuk: not sure it's the best option, but also not sure having checkpointsPath flag in vlagent also makes sense

api/operator/v1/vlagent_types.go
AndrewChubatiuk: updated

api/operator/v1/vlagent_types.go
f41gh7: Well, in my opinion, it's better to make CRD api as much compatible with vlagent as possible. So having `checkpointsPath` as a config option should be fine. Because, `dataPath` could collide with buffer data path.


 Maybe, best option would provide `checkpointsPath` config option at CRD. If it's not set, attach `hostPath` mount at `/var/lib/vlagent_checkpoints`  and point flag `kubernetesCollector.checkpointsPath` for `/var/lib/vlagent_checkpoints/vlagent-kubernetes-checkpoints.json`.

 From practical stand-point, it makes sense to always persist checkpoints out of box. Rarely it'll be needed to specify some other path due to security restrictions?

f41gh7: The reason why it concerns me, that vlagent is `v1` CRD and we cannot simply remove or rename fields during refactoring.

internal/controller/operator/factory/vlagent/rbac.go
AndrewChubatiuk: this check is done before calling this function, rbac is not created for explicitly defined service account

api/operator/v1/vlagent_types.go
AndrewChubatiuk: replaced dataPath with checkpointsPath

internal/controller/operator/factory/vlagent/vlagent.go
```diff
@@ -569,6 +707,33 @@ func deleteOrphaned(ctx context.Context, rclient client.Client, cr *vmv1.VLAgent
...
+		if config.IsClusterWideAccessAllowed() {
+			rbacMeta := metav1.ObjectMeta{Name: cr.GetClusterRoleName(), Namespace: cr.Namespace}
+			objects = append(objects,
+				&rbacv1.ClusterRoleBinding{ObjectMeta: rbacMeta},
+				&rbacv1.ClusterRole{ObjectMeta: rbacMeta},
+			)
+		}
+		owner := cr.AsCRDOwner()
+		for _, o := range objects {
+			if err := finalize.SafeDeleteWithFinalizer(ctx, rclient, o, owner); err != nil {
```
f41gh7, line 727: WDYT of changing delete for RBAC with `SafeDeleteForSelectorsWithFinalizer`?

 Also, it looks like ServiceAccount already deleted above:

```
	if !cr.IsOwnsServiceAccount() {
		if err := finalize.SafeDeleteWithFinalizer(ctx, rclient, &corev1.ServiceAccount{ObjectMeta: objMeta}, &owner); err != nil {
			return fmt.Errorf("cannot remove serviceaccount: %w", err)
		}
	}
```

config/examples/vlagent-collector.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: VLAgent
+metadata:
+  name: example
+spec:
+  resources:
+    requests:
+      cpu: "50m"
+      memory: "350Mi"
+    limits:
+      cpu: "500m"
```
f41gh7, line 11: Could you please add volumes and volumeMounts example for persistent queue?

internal/controller/operator/factory/vlagent/vlagent.go
AndrewChubatiuk: fixed it

config/examples/vlagent-collector.yaml
AndrewChubatiuk: done

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #1683, opened 2025-12-11
config-reloader: removed 3rd-party config reloaders

AndrewChubatiuk:
removed 3rd-party config reloaders
prepared PR to be merged once we're ready for it


api/operator/v1beta1/vmextra_types.go
```diff
@@ -1167,6 +1167,7 @@ type CommonConfigReloaderParams struct {
 	// UseVMConfigReloader replaces prometheus-like config-reloader
 	// with vm one. It uses secrets watch instead of file watch
 	// which greatly increases speed of config updates
+	// Deprecated: this property is ignored and no longer needed
```
vrutkovs:
```suggestion
	// Removed since 0.67.0: this property is ignored and no longer needed
```

vrutkovs: It would be useful to mention when this was deprecated (not sure if its a common practice in the repo already)

PR #1687, opened 2025-12-12
hpa: added behavior and deprecated behaviour

AndrewChubatiuk:
replaced `behaviour` with `behavior` in docs and code comments
added HPA behavior property and deprecated behaviour


AndrewChubatiuk:
closing, as both are correct

PR #1698, opened 2025-12-16
vlagent: disable strict security for vlagent in collector mode

AndrewChubatiuk:
do not set strict security for vlagent in collector mode, since root access is required

### DIFF
--- a/internal/controller/operator/factory/vlagent/vlagent.go
+++ b/internal/controller/operator/factory/vlagent/vlagent.go
@@ -198,7 +198,6 @@ func newK8sApp(cr *vmv1.VLAgent) (client.Object, error) {
 				},
 			},
 		}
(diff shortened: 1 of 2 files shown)


internal/controller/operator/factory/vlagent/vlagent.go
```diff
@@ -198,7 +198,6 @@ func newK8sApp(cr *vmv1.VLAgent) (client.Object, error) {
 				},
 			},
 		}
-		build.DaemonSetAddCommonParams(dsSpec, useStrictSecurity, &cr.Spec.CommonApplicationDeploymentParams)
```
vrutkovs, line 201: We're not using `useStrictSecurity` anywhere else

vrutkovs:
imo setting `useStrictSecurity: true` should enable it running as non-root (and tbf it should be default) - there is no valid reason for vlagent to crash when it fails to read a file (which can be intermittent). I don't think its an operator issue at all

AndrewChubatiuk:
okay, closing this

Issue #1701, opened 2025-12-17
vmcluster: support `extraStorageNodes` for vmselect and vminsert


zekker6:
Currently, using multi-level deployments require to define additional storage nodes via `{vmselect/vminsert}.extraArgs.storageNodes`.
That is error prone as changes in number of storage nodes needs to be reflected for these flags as well.

VLCluster and VTCluster resources support [`extraStorageNodes`](https://docs.victoriametrics.com/operator/api/#vlselect-extrastoragenodes) which allows to add an additional list of storage nodes automatically, so that it is easier change number of storage nodes without changing other parameters. It would be great for VMCluster to support the same parameter.

AndrewChubatiuk:
VTCluster and VLCluster have this property on select level, it was done to simplify migration from single to cluster. there's no such option for data ingestion. wdyt about having targetRefs or clusterSelector instead as cluster nodes scaling require also to change listed storage nodes at `extraStorageNodes`

zekker6:
Using `targetRef` would be even better in terms of simplicity of use. But this might be limited to having VMClusters located within a single k8s cluster which is not always the case.

AndrewChubatiuk:
also this option wasn't added to vmcluster since vmsingle cannot at the moment be joined as a storage

f41gh7:
As far as I understand, multi-level config must be:

```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
metadata:
  name: top-level
spec:
  retentionPeriod: "4"
  vmselect:
    replicaCount: 2
    extraArgs:
      storageNode: vmselect-zone-0:8421,vmselect-zone1:8421
  vminsert:
    replicaCount: 2
    extraArgs:
      storageNode: vminsert-zone-0:8420,vminsert-zone-1:8420
---
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
metadata:
  name: zone-0
spec:
  retentionPeriod: "4"
  replicationFactor: 2
  vmstorage:
    replicaCount: 2
  vmselect:
    clusterNativeListenPort: "8421"
    replicaCount: 2
  vminsert:
    clusterNativeListenPort: "8420"
    replicaCount: 2
---
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
metadata:
  name: zone-1
spec:
  retentionPeriod: "4"
  replicationFactor: 2
  vmstorage:
    replicaCount: 2
  vmselect:
[...]

PR #1703, opened 2025-12-18
support both arrays and strings in extraArgs values

AndrewChubatiuk:
this allows to set extraArgs values not only as string, but as slices as well
this should allow to make infinite extraArgs values more readable

related issue https://github.com/VictoriaMetrics/operator/issues/1701


docs/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ aliases:
 * FEATURE: [vlagent](https://docs.victoriametrics.com/operator/resources/vlagent): support logs collection. See [#1501](https://github.com/VictoriaMetrics/operator/issues/1501).
 * FEATURE: [vmoperator](https://docs.victoriametrics.com/operator/): use `operator_bad_objects_total` metric with `object_namespace` and `crd` labels to track invalid objects managed by VMAgent, VMAuth, VMAlert and VMAlertmanager. Old `operator_alertmanager_bad_objects_count` and `operator_vmalert_bad_objects_count` are deprecated and will be removed in next releases.
 * FEATURE: [vmoperator](https://docs.victoriametrics.com/operator/): added HPA support for all cluster CR storage. See [#1678](https://github.com/VictoriaMetrics/operator/issues/1678).
+* FEATURE: [vmoperator](https://docs.victoriametrics.com/operator/): support both array and string values in extraArgs.
```
vrutkovs, line 30:
```suggestion
* FEATURE: [vmoperator](https://docs.victoriametrics.com/operator/): support both array, comma-separated and string values in extraArgs.
```

vrutkovs (changes requested):
Not sure if "comma-separated string" is alertmanager-specific. If not lets add more unit tests for this case?

AndrewChubatiuk:
adding this feature still will not help to properly organize flags, that should be configured together like ones, which are related to downsampling, retention filters, remote write. better introduce dedicated sections instead

PR #1704, opened 2025-12-20
consistently use 127.0.0.1 in reloader, snapshot and alertmanager URLs instead of localhost for setups without IPV6 enabled

AndrewChubatiuk:


internal/config/config.go
```diff
@@ -664,6 +664,15 @@ func MustGetBaseConfig() *BaseOperatorConf {
 	return opConf
 }
 
+// GetLocalhost returns localhost value depending on global configuration
+func GetLocalhost() string {
+	cfg := MustGetBaseConfig()
+	if cfg.EnableTCP6 {
+		return "localhost"
```
vrutkovs, line 671: Why not `[::1]`? In `EnableTCP6` do we prefer IPv6 or make it optional (and use IPv4 generally)?

AndrewChubatiuk: EnableTCP6 enables dual stack, not tcp6 only, decided to keep localhost to minimize diff for existing setups

Issue #1696, opened 2025-12-15
VLAgent: k8sCollector errors out if it cannot read logs


syphernl:
### Describe the bug

I am running VL in Minikube. 
I enabled `k8sCollector` in VLAgent so it can read the cluster logs and forward them to VL. 
This however results in a fatal error preventing VLAgent from running.

The entries in `/var/logs/containers` are symlinks to entries in `/var/lib/docker/containers`, but the `VLAgent` pod mounts both `/var/log` and `/var/lib` as hostPath so  `/var/lib/docker/containers` should be available.

Either way: VLAgent should not crash if it cannot read particular logfiles, it should log this error and continue.

### To Reproduce

- Run VLAgent in Minikube, deployed via the Operator
- See VLAgent crashes because it cannot read the logs

### Version

victoria-logs-v1.41.0

### Logs

```
2025-12-15T15:28:03.283851594Z 2025-12-15T15:28:03.283Z info    VictoriaLogs/app/vlagent/kubernetescollector/kubernetes.go:54   started Kubernetes log collector for node "victoria"
2025-12-15T15:28:03.283855344Z 2025-12-15T15:28:03.283Z info    VictoriaLogs/app/vlagent/main.go:58     started vlagent in 0.006 seconds
2025-12-15T15:28:03.284461524Z 2025-12-15T15:28:03.284Z info    VictoriaMetrics@v1.131.0/lib/httpserver/httpserver.go:145       started server at http://0.0.0.0:9429/
2025-12-15T15:28:03.284475452Z 2025-12-15T15:28:03.284Z info    VictoriaMetrics@v1.131.0/lib/httpserver/httpserver.go:147       pprof handlers are exposed at http://0.0.0.0:9429/debug/pprof/
2025-12-15T15:28:03.294485902Z 2025-12-15T15:28:03.294Z panic   VictoriaLogs/app/vlagent/kubernetescollector/file_collector.go:293      FATAL: cannot open file "/var/log/containers/vmsingle-victoriametrics-75f6848954-zkg9h_vm_vmsingle-8d668e79ac7bcc5161834257c1955162d2b5241993c5df6be990f054a387114f.log": open /var/log/containers/vmsingle-victoriametrics-75f6848954-zkg9h_vm_vmsingle-8d668e79ac7bcc5161834257c1955162d2b5241993c5df6be990f054a387114f.log: permission denied
[...]

PR #1705, opened 2025-12-20
vlagent: use single volume for all vlagent data by default

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1696

- added collector extraFilter, includePodLabels, includePodAnnotations, includeNodeLabels and includeNodeAnnotations support
- use single volume for all vlagent data by default, before there was a separate path for checkpoints and remotewrite data by default
- update strict security mode for vlagent, before it failed


api/operator/v1/vlagent_types.go
```diff
@@ -198,8 +219,8 @@ type VLAgentRemoteWriteSettings struct {
 	// Whether to show -remoteWrite.url in the exported metrics. It is hidden by default, since it can contain sensitive auth info
 	// +optional
 	ShowURL *bool `json:"showURL,omitempty"`
-	// Path to directory where temporary data for remote write component is stored (default /vlagent_pq/vlagent-remotewrite-data)
-	// If defined, operator ignores spec.storage field and skips adding volume and volumeMount for pq
+	// Path to directory where temporary data for remote write component is stored.
+	// Defaults to /vlagent-data/vlagent-remotewrite-data or /var/lib/vlagent-data/vlagent-remotewrite-data in collector mode
```
vrutkovs, line 228: If undefined defaults to `/var/lib/vlagent-data`

internal/controller/operator/factory/vlagent/vlagent.go
```diff
@@ -288,6 +287,21 @@ func newPodSpec(cr *vmv1.VLAgent) (*corev1.PodSpec, error) {
 		if len(cr.Spec.K8sCollector.ExtraFields) > 0 {
 			args = append(args, fmt.Sprintf("-kubernetesCollector.extraFields=%s", cr.Spec.K8sCollector.ExtraFields))
 		}
+		if len(cr.Spec.K8sCollector.ExcludeFilter) > 0 {
+			args = append(args, fmt.Sprintf("-kubernetesCollector.excludeFilter=%s", cr.Spec.K8sCollector.ExcludeFilter))
+		}
+		if ptr.Deref(cr.Spec.K8sCollector.IncludePodLabels, true) {
+			args = append(args, "-kubernetesCollector.includePodLabels")
+		}
+		if ptr.Deref(cr.Spec.K8sCollector.IncludePodAnnotations, false) {
```
vrutkovs, line 296:
```suggestion
		if ptr.Deref(cr.Spec.K8sCollector.IncludePodAnnotations, true) {
```
We should not be setting this flag if its unset, right?

AndrewChubatiuk: kept it's value the same as vlagent has

api/operator/v1/vlagent_types.go
AndrewChubatiuk: value in PR is correct, this is spec.remoteWriteSettings.tmpData, which is a data path for remote write data

Issue #1707, opened 2025-12-22
Operator stuck object reconcile on `errRecreate`


f41gh7:
Currently, if operator encounters `errRecreate` added at c280a4e5b216cb2c06e34962d7507cfce8170eaf. It silently waits for object to be removed. It leads to infinity loop, if object has some finalizer.

 It either should log proper error message or inform user in some other way.

AndrewChubatiuk:
added log message that shows recreating resource and it's finalizers

AndrewChubatiuk:
logging was added to 0.67.0, additionally going to review finalizers for all CRs

AndrewChubatiuk:
fixed this bug starting 0.68.0

PR #1708, opened 2025-12-22
reconcile: log errRecreate errors, added namespace to error message for namespaced resources

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1707

### DIFF
--- a/internal/controller/operator/factory/reconcile/reconcile.go
+++ b/internal/controller/operator/factory/reconcile/reconcile.go
@@ -1,13 +1,18 @@
 package reconcile
 
 import (
+	"context"
 	"errors"
+	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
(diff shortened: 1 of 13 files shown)


internal/controller/operator/factory/reconcile/reconcile.go
```diff
@@ -134,17 +139,21 @@ func isRecreate(err error) bool {
...
 
 type errRecreate struct {
-	origin error
+	msg string
 }
 
-// Error implements errors.Error interface
-func (e *errRecreate) Error() string {
-	return e.origin.Error()
+func newErrRecreate(ctx context.Context, kind string, r client.Object) *errRecreate {
```
vrutkovs: `kind` can be extracted from the object: `r.GetObjectKind().GroupVersionKind().Kind`

AndrewChubatiuk: updated

Issue #1078, opened 2024-08-20
VMProbe supporting pods, not just ingresses


paradisfm:
### Is your feature request related to a problem? Please describe

I would like to create a gRPC probe for pods with certain labels using Blackbox Exporter. 

### Describe the solution you'd like

 I would really appreciate an addition of pod support to targets the VMProbe can handle. 

### Describe alternatives you've considered

I'm trying to work around the ingress-only limitation with a VMPodScrape, but not having much luck thus far -- I can't seem to get pod discovery working in tandem with a blackbox probe.

paradisfm:
i have successfully constructed what i want with a vmpodscrape, however i stand by the idea that supporting pods for a vmprobe would be better in terms of ease of use.

AndrewChubatiuk:
these PRs introduce VMProbes support for HTTPRoute, GRPCRoute and K8s Service
https://github.com/VictoriaMetrics/operator/pull/1715
https://github.com/VictoriaMetrics/VictoriaMetrics/pull/10253

AndrewChubatiuk:
starting 0.68.0 VMProbe additionally to ingress also supports, service, pod and node roles
check example https://github.com/VictoriaMetrics/operator/blob/master/config/examples/vmprobe-k8s.yaml

PR #1720, opened 2026-01-08
vmprobe: support service, pod and node

AndrewChubatiuk:
added `spec.targets.kubernetes` and deprecated `spec.targets.ingress`. `spec.targets.kubernetes` supports besides `ingress` also `pod`, `service` and `node` roles. later gateway api roles support can be easily added
added `spec.targets.static` and deprecated `spec.targets.staticConfig` for consistency
fixes https://github.com/VictoriaMetrics/operator/issues/1078, https://github.com/VictoriaMetrics/operator/issues/1716


AndrewChubatiuk:
@cubic-dev-ai review this PR

config/crd/overlay/crd.yaml
```diff
@@ -26645,6 +26652,138 @@ spec:
...
+                      properties:
+                        namespaceSelector:
+                          properties:
+                            any:
+                              type: boolean
+                            matchNames:
+                              items:
+                                type: string
+                              type: array
+                          type: object
```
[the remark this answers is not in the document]
AndrewChubatiuk: it's a temporary workaround, since same `VMProbeTargetKubernetes` struct is being used by both `kubernetes` and `ingress` sections. Once `ingress` is finally removed this section becomes required

api/operator/v1beta1/vmprobe_types.go
```diff
@@ -82,17 +92,33 @@ type VMProbeTargetStaticConfig struct {
...
 }
 
-// ProbeTargetIngress defines the set of Ingress objects considered for probing.
+// VMProbeTargetKubernetes defines the set of k8s objects considered for probing.
 // +k8s:openapi-gen=true
-type ProbeTargetIngress struct {
-	// Select Ingress objects by labels.
+type VMProbeTargetKubernetes struct {
+	// Role defines k8s role name
+	// +kubebuilder:validation:Enum=service;ingress;pod;node
```
vrutkovs, line 99:
```suggestion
	// +kubebuilder:validation:Enum=service;ingress;pod;node;httproute
```
For HTTPRoute implementation, which is an alternative to Ingress

vrutkovs: Perhaps its best to implement it later?

api/operator/v1beta1/vmprobe_types.go
```diff
@@ -154,17 +180,28 @@ func (cr *VMProbe) Validate() error {
...
+			return fmt.Errorf("spec.targets.kubernetes[%d].role is required", i)
 		}
-		if err := checkRelabelConfigs(cr.Spec.Targets.Ingress.RelabelConfigs); err != nil {
-			return fmt.Errorf("invalid ingress.relabelingConfigs: %w", err)
+		if err := t.validate(fmt.Sprintf("spec.targets.kubernetes[%d]", i)); err != nil {
+			return err
 		}
-	case cr.Spec.Targets.StaticConfig != nil:
+	}
+	if err := cr.Spec.Targets.Ingress.validate("spec.targets.ingress"); err != nil {
```
vrutkovs: ingress is deprecated - do we want to validate it last to find errors in supported vars quicker?

api/operator/v1beta1/vmprobe_types.go
```diff
@@ -82,17 +92,33 @@ type VMProbeTargetStaticConfig struct {
 	RelabelConfigs []*RelabelConfig `json:"relabelingConfigs,omitempty"`
 }
 
-// ProbeTargetIngress defines the set of Ingress objects considered for probing.
+// VMProbeTargetKubernetes defines the set of k8s objects considered for probing.
 // +k8s:openapi-gen=true
-type ProbeTargetIngress struct {
-	// Select Ingress objects by labels.
+type VMProbeTargetKubernetes struct {
```
vrutkovs, line 97: Use GVK instead of a single `Role`? This would lock us onto the version and predefined set of groups, which may depend on k8s version

vrutkovs (changes requested):

api/operator/v1beta1/vmprobe_types.go
AndrewChubatiuk: httproute should be supported by vmagent first

api/operator/v1beta1/vmprobe_types.go
AndrewChubatiuk: Role is also used in other scrape types, it's better to keep it as is

api/operator/v1beta1/vmprobe_types.go
AndrewChubatiuk: updated

PR #1725, opened 2026-01-11
chore: common compression functions

AndrewChubatiuk:
use common gzip/gunzip functions

### DIFF
--- a/internal/controller/operator/factory/build/build.go
+++ b/internal/controller/operator/factory/build/build.go
@@ -1,7 +1,10 @@
 package build
 
 import (
+	"bytes"
+	"compress/gzip"
 	"fmt"
+	"io"
 	"path"
 	"strconv"
 	"strings"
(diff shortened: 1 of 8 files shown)


internal/controller/operator/factory/build/build.go
```diff
@@ -251,3 +254,27 @@ func RelabelVolumeTo(volumes []corev1.Volume, mounts []corev1.VolumeMount, cr re
 	}
 	return volumes, mounts
 }
+
+// GzipConfig compresses data config
+func GzipConfig(data []byte) ([]byte, error) {
+	var buf bytes.Buffer
+	w := gzip.NewWriter(&buf)
+	if _, err := w.Write(data); err != nil {
+		return nil, err
```
vrutkovs:
```suggestion
		return nil, fmt.Errorf("failed to write compressed data to buffer: %w", err)
```

internal/controller/operator/factory/build/build.go
```diff
@@ -251,3 +254,27 @@ func RelabelVolumeTo(volumes []corev1.Volume, mounts []corev1.VolumeMount, cr re
...
+	}
+	if err := w.Close(); err != nil {
+		return nil, fmt.Errorf("failed to flush compressed data to buffer, %w", err)
```
vrutkovs:
```suggestion
		return nil, fmt.Errorf("failed to flush compressed data to buffer: %w", err)
```

PR #1702, opened 2025-12-17
vmsingle: introduce scraping support

AndrewChubatiuk:
moved scraping logic to vmscrapes package and introduced scraping support for vmsingle (disabled by default). ingestOnlyMode should be set to false to enable scraping
fixes https://github.com/VictoriaMetrics/operator/issues/1694

PR #1726, opened 2026-01-11
chore: decouple scraping logic from vmagent code

AndrewChubatiuk:
decoupling scrape configuration functions from vmagent to be able to reuse it later by vmsingle
preparation for https://github.com/VictoriaMetrics/operator/pull/1702

### DIFF
--- a/internal/controller/operator/vmagent_controller.go
+++ b/internal/controller/operator/vmagent_controller.go
@@ -36,7 +36,7 @@ import (
 )
 
 var (
-	agentSync           sync.Mutex
+	agentSync           sync.RWMutex
 	agentReconcileLimit = limiter.NewRateLimiter("vmagent", 5)
 )
 
@@ -88,9 +88,10 @@ func (r *VMAgentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 	if err := r.Get(ctx, req.NamespacedName, instance); err != nil {
 		return result, &getError{origin: err, controller: "vmagent", requestObject: req}
 	}
+
 	if !instance.IsUnmanaged(nil) {
-		agentSync.Lock()
-		defer agentSync.Unlock()
+		agentSync.RLock()
+		defer agentSync.RUnlock()
 	}
 
 	RegisterObjectStat(instance, "vmagent")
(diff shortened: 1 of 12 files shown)


api/operator/v1beta1/common_scrapeparams.go
```diff
@@ -548,8 +555,62 @@ type CommonScrapeParams struct {
...
+	} else if cr.VMAgentExternalLabelName != nil {
+		labelName = cr.VMAgentExternalLabelName
+	}
+	if labelName != nil {
+		if *labelName != "" {
+			prometheusExternalLabelName = *labelName
+		} else {
+			prometheusExternalLabelName = ""
+		}
+	}
```
vrutkovs:
```suggestion
func (cr *CommonScrapeParams) externalLabelName() string {
	// Use "prometheus" external label name by default if field is missing.
	// in case of migration from prometheus to vmagent, it helps to have same labels
	// Do not add external label if field is set to empty string.

	if cr.ExternalLabelName != nil {
		return *cr.ExternalLabelName
	}
	if cr.VMAgentExternalLabelName != nil {
		return *cr.VMAgentExternalLabelName
	}
	return "prometheus"
}

func (cr *CommonScrapeParams) externalLabels(defaultLabelValue string) map[string]string {
	m := map[string]string{}

	prometheusExternalLabelName := cr.externalLabelName()
```

internal/controller/operator/vmagent_controller.go
```diff
@@ -89,7 +89,7 @@ func (r *VMAgentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 	if err := r.Get(ctx, req.NamespacedName, instance); err != nil {
 		return result, &getError{origin: err, controller: "vmagent", requestObject: req}
 	}
-	if !instance.IsUnmanaged() {
+	if !instance.IsScrapeUnmanaged(nil) {
```
vrutkovs: Do we have separate "full" vmagent unmanaged and only scrape unmanaged?

AndrewChubatiuk: no, we don't, but currently IsUnmanaged returns true if it's not scrapable, but it impacts stream aggregation and relabelling configs as well, that's why renamed this function

AndrewChubatiuk: But already renamed it back

christensenjairus:
@AndrewChubatiuk @vrutkovs  Are we sure we haven't broken functionality here? This is the only commit between the two most recent versions that looks to be related to an issue I'm seeing.

W/ the new version (`0.67.0`) if I have multiple scrapes defined for the same endpointslices, vmagent gets the definitions of all of them jumbled.

Here's an example: 
### v0.66.1 (working as intended)
Screenshot 2026-01-30 at 5 29 09 PM


### v0.67.0
Screenshot 2026-01-30 at 5 32 19 PM

It gets the ports, protocols, everything confused, and even duplicated, as if it's trying to scrape every port it knows exists

AndrewChubatiuk:
this PR is not included into 0.67.0, I'll check changes that happened before
could you please share these service scrape objects and vmagent CR?

vrutkovs:
Please file a new bug for this, as Andrew has pointed out this PR didn't make it in 0.67

AndrewChubatiuk:
looked through diff 0.66.1...0.67.0 and found one change, that could lead to this issue, [created PR](https://github.com/VictoriaMetrics/operator/pull/1771) with a fix

PR #1729, opened 2026-01-14
vmanomaly: add support of server configuration

zekker6:
Previously, it was only possible to change port of the server as a part of configuration. Add support of other properties supported by vmanomaly.


AndrewChubatiuk (approved):
LGTM!

Issue #1740, opened 2026-01-20
do not list Config reloader as the default container.


tiny-pangolin:
Currently, some of the pods created by vmoperator list config-reloader as the default container which leads to users running kubectl get logs <pod> and only get logs from config-reloader which is usually not the container that users want to see logs from, this also happens when tools like headlamp show logs in their ui. It would be helpful if pods that contain multiple containers list the Victoriametrics component as the default container instead of config reloader.

AndrewChubatiuk:
order of vmagent containers was changed and it's expected to be available in upcoming release

AndrewChubatiuk:
vmagent is a default container in pod, that is managed by VMAgent CR starting operator [v0.67.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.67.0)

PR #1741, opened 2026-01-21
vmagent: make vmagent container default

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1740


vrutkovs:
Tests still assume the first container will be config-reloader.

Nitpick: commit message describes what it's doing ("change the container order") but not why ("so that `kubectl logs` would default to the container we're interested the most")

PR #1750, opened 2026-01-26
fix(vmalertmanager): cluster.peer argument hostnames

drsherluck:
`--cluster-peer` argument values are using `cr.PrefixName()` for the namespace part in the service dns (prefix-name.prefix-name.svc.cluster.local. instead of prefix-name.namespace.svc.cluster.local.)

---
## Summary by cubic
Fixes VMAlertmanager --cluster-peer hostnames so peers resolve correctly. The service DNS suffix is now built from the namespace, not the prefixed resource name.

- **Bug Fixes**
  - Build clusterPeerDomain as "namespace.svc.<cluster-domain>" or "namespace" (when no cluster domain), instead of using PrefixedName.
  - Prevents malformed addresses like "prefix-name.namespace.svc.cluster.local." that broke peer discovery.

<sup>Written for commit 5b4a0e9e948df869de6f6887f01d9d12e892eaa2. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/vmalertmanager/statefulset.go
+++ b/internal/controller/operator/factory/vmalertmanager/statefulset.go
@@ -213,10 +213,10 @@ func makeStatefulSetSpec(cr *vmv1beta1.VMAlertmanager) (*appsv1.StatefulSetSpec,
 
 	var clusterPeerDomain string
 	if cr.Spec.ClusterDomainName != "" {
-		clusterPeerDomain = fmt.Sprintf("%s.%s.svc.%s.", cr.PrefixedName(), cr.Namespace, cr.Spec.ClusterDomainName)
+		clusterPeerDomain = fmt.Sprintf("%s.svc.%s.", cr.Namespace, cr.Spec.ClusterDomainName)
 	} else {
 		// The default DNS search path is .svc.<cluster domain>
-		clusterPeerDomain = cr.PrefixedName()
+		clusterPeerDomain = cr.Namespace
 	}
 
 	for i := int32(0); i < ptr.Deref(cr.Spec.ReplicaCount, 0); i++ {


AndrewChubatiuk:
Thanks for contribution!

internal/controller/operator/factory/vmalertmanager/statefulset.go
```diff
@@ -213,10 +213,10 @@ func makeStatefulSetSpec(cr *vmv1beta1.VMAlertmanager) (*appsv1.StatefulSetSpec,
 
 	var clusterPeerDomain string
 	if cr.Spec.ClusterDomainName != "" {
-		clusterPeerDomain = fmt.Sprintf("%s.%s.svc.%s.", cr.PrefixedName(), cr.Namespace, cr.Spec.ClusterDomainName)
+		clusterPeerDomain = fmt.Sprintf("%s.svc.%s.", cr.Namespace, cr.Spec.ClusterDomainName)
 	} else {
 		// The default DNS search path is .svc.<cluster domain>
-		clusterPeerDomain = cr.PrefixedName()
+		clusterPeerDomain = cr.Namespace
```
f41gh7, line 219: It'll generate dns name:

pod-idx.service. Which is wrong format of Kubernetes Statefulset stable id. It requires either pod name with idx or pod name idx with statefulset service name.

Did you even test this changes?

vrutkovs:
This is a tough one - the code could certainly use some love, but unfortunately, the suggested change breaks the resolution as well.

The problem here is in https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id: the pod hostname needs to include both the StatefulSet name and the Service name. Unfortunately, both of these are `cr.PrefixedName()`, which gives an illusion of "duplicated" strings. Moreover, k8s allows us to omit namespace - and this code gladly does so, creating an illusion of "omitted" namespace. Altogether, this makes the fix seem "easy", but it creates a DNS name that cannot be mapped to the pod.

@drsherluck  Would you like to submit a new pull request improving this? I think the requirements would be:
* explicitly set namespace and use the full pod DNS
* add unit tests
* explain the change in the changelog

Thanks in advance!

drsherluck:
> This is a tough one - the code could certainly use some love, but unfortunately, the suggested change breaks the resolution as well.
> 
> The problem here is in [kubernetes.io/docs/concepts/workloads/controllers/statefulset#stable-network-id](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-network-id): the pod hostname needs to include both the StatefulSet name and the Service name. Unfortunately, both of these are `cr.PrefixedName()`, which gives an illusion of "duplicated" strings. Moreover, k8s allows us to omit namespace - and this code gladly does so, creating an illusion of "omitted" namespace. Altogether, this makes the fix seem "easy", but it creates a DNS name that cannot be mapped to the pod.

I made a mistake with this PR and I honestly didn't test (I only ran `make test` and `make test-e2e` with no problems) it because I thought it was an easy and obvious fix, but it was a misunderstanding by my part. Sorry :(

> @drsherluck Would you like to submit a new pull request improving this? I think the requirements would be:
> 
> * explicitly set namespace and use the full pod DNS
> * add unit tests
> * explain the change in the changelog
> 
> Thanks in advance!

Yes, I will create a new PR today that fixes my mistake.

Issue #1530, opened 2025-09-16
Auto-sync resource's statefulset volumeClaimTemplates annotation changes to existing PVC


Haleygo:
After the PVC creation, the StatefulSet no longer controls the PVC, so changes to the StatefulSet’s volumeClaimTemplates are not synchronized with the PVC, see [this issue](https://github.com/kubernetes/kubernetes/issues/121178).
So for CR like vmagent, vmcluster, changes to `xxxStorage.volumeClaimTemplate.metadata.annotations` aren’t synced to existing PVCs and require manual PVC edits. It would be nice if vm-operator could auto-sync such changes, just as it does for storage size.

AndrewChubatiuk:
this cannot be easily implemented, since you don't know in advance which annotations are automatically added. it's possible to add them only, but not remove.

AndrewChubatiuk:
starting 0.68.0 PVC annotations and labels are synced automatically

PR #1758, opened 2026-01-27
reconcile/pvc: do not print warning, when target PVC size is smaller,…

### LINKED ISSUE #1747: pvc-allow-volume-expansion annotation does not work on PVC's consumed by VMSingle & VLSingle
We use **pvc-autoresizer** to automatically increase PVC sizes. This causes `.spec.storage.resources.requests.storage` in the VictoriaMetrics CR to become out-of-sync with the actual deployed PVC size. 
The operator then attempts to resize the PVC back to the configured size but fails.

Example log:

```txt
2026-01-26T10:03:56.622481829Z {"level":"info","ts":"2026-01-26T10:03:56Z","logger":"controller.VLSingle","msg":"cannot decrease PVC=vlsingle-victorialogs size from=38Gi to=10Gi, please check VolumeClaimTemplate configuration","vlsingle":"victorialogs","namespace":"victorialogs"}
```

### Current Behavior

According to https://github.com/VictoriaMetrics/operator/issues/867 and https://github.com/VictoriaMetrics/operator/blob/master/config/examples/vmcluster_storage_class_unexpandable.yaml the annotation:

```yaml
operator.victoriametrics.com/pvc-allow-volume-expansion: "false"
```

can be set to prevent the operator from decreasing PVC size. The given example is for `VMCluster`. 

I have applied this annotation to the PVC consumed by  `VMSingle` and `VLSingle` but it appears this does not prevent the operator from attempting to resize.

### Expected Behavior

* Annotation `operator.victoriametrics.com/pvc-allow-volume-expansion: "false"` should also work for PVC's consumed by `VMSingle` and `VLSingle` objects.

### Notes

* Current behavior leads to unnecessary error logs and potential conflicts with external PVC management.
* Likely this feature is **not yet implemented** for PVC's consumed by `VMSingle` and `VLSingle`.

AndrewChubatiuk:
… than actual when `operator.victoriametrics.com/pvc-allow-volume-expansion` annotation set to `false`

use single function for updating StatefulSet ClaimTemplates and for PVC reconcile, which now besides expanding storage for StatefulSet PVCs also updates metadata. additionally fixed tests data and added check of PVC update results

fixes https://github.com/VictoriaMetrics/operator/issues/1530
fixes https://github.com/VictoriaMetrics/operator/issues/1747

---
## Summary by cubic
Suppress PVC downscale warnings when operator.victoriametrics.com/pvc-allow-volume-expansion=false is set on the claim/template, and ensure PVC metadata (labels/annotations/finalizers) is synced during reconciliation. This reduces noisy logs and fixes #1747 and #1530.

- **Bug Fixes**
  - Skip the downscale warning when operator.victoriametrics.com/pvc-allow-volume-expansion=false.

- **Refactors**
  - Consolidated PVC reconciliation and StatefulSet PVC updates into updatePVC/updateSTSPVC, which also updates PVC metadata.

<sup>Written for commit 8be63f570a0f946c745f800e1d7c8aabdc3ed374. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/reconcile/pvc.go
+++ b/internal/controller/operator/factory/reconcile/pvc.go
@@ -5,13 +5,10 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	vmv1beta1 "github.com/VictoriaMetrics/operator/api/operator/v1beta1"
 	"github.com/VictoriaMetrics/operator/internal/controller/operator/factory/logger"
 )
 
@@ -40,40 +37,6 @@ func PersistentVolumeClaim(ctx context.Context, rclient client.Client, newPVC, p
 			" To fix this, make backup for this pvc, delete pvc finalizers and restore from backup.")
 		return nil
 	}
-	newSize := newPVC.Spec.Resources.Requests.Storage()
-	oldSize := currentPVC.Spec.Resources.Requests.Storage()
 
-	isResizeNeeded := mayGrow(ctx, newPVC, newSize, oldSize)
-	if !isResizeNeeded &&
-		equality.Semantic.DeepEqual(newPVC.Labels, currentPVC.Labels) &&
-		isObjectMetaEqual(currentPVC, newPVC, prevPVC) {
-		return nil
-	}
-	if isResizeNeeded {
-		// check if storage class is expandable
-		isExpandable, err := isStorageClassExpandable(ctx, rclient, newPVC)
-		if err != nil {
-			return fmt.Errorf("failed to check storageClass expandability for PVC %s: %v", newPVC.Name, err)
-		}
-		if !isExpandable {
-			// don't return error to caller, since there is no point to requeue and reconcile this when sc is unexpandable
-			logger.WithContext(ctx).Info(fmt.Sprintf("storageClass %s for PVC %s doesn't support live resizing", ptr.Deref(newPVC.Spec.StorageClassName, "default"), newPVC.Name))
-			return nil
-		}
-	}
-
-	newResources := newPVC.Spec.Resources.DeepCopy()
-	// keep old spec with new resource requests
-	newPVC.Spec = currentPVC.Spec
-	newPVC.Spec.Resources = *newResources
-
-	vmv1beta1.AddFinalizer(newPVC, currentPVC)
-	mergeObjectMetadataIntoNew(currentPVC, newPVC, prevPVC)
-
-	logger.WithContext(ctx).Info(fmt.Sprintf("updating PVC %s configuration", newPVC.Name))
-
-	if err := rclient.Update(ctx, newPVC); err != nil {
-		return err
-	}
-	return nil
+	return updatePVC(ctx, rclient, currentPVC, newPVC, prevPVC)
 }
--- a/internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
+++ b/internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
@@ -3,6 +3,7 @@ package reconcile
 import (
 	"context"
 	"fmt"
+	"maps"
 	"strings"
 	"time"
 
@@ -84,7 +85,7 @@ func getPVCFromSTS(pvcName string, sts *appsv1.StatefulSet) *corev1.PersistentVo
 	return nil
 }
 
-func growSTSPVC(ctx context.Context, rclient client.Client, sts *appsv1.StatefulSet) error {
+func updateSTSPVC(ctx context.Context, rclient client.Client, sts *appsv1.StatefulSet) error {
 	// fast path
 	if sts.Spec.Replicas != nil && *sts.Spec.Replicas == 0 {
 		return nil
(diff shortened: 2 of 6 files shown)


internal/controller/operator/factory/reconcile/pvc.go
```diff
@@ -43,21 +43,29 @@ func PersistentVolumeClaim(ctx context.Context, rclient client.Client, newPVC, p
 	newSize := newPVC.Spec.Resources.Requests.Storage()
 	oldSize := currentPVC.Spec.Resources.Requests.Storage()
 
-	isResizeNeeded := mayGrow(ctx, newPVC, newSize, oldSize)
-	if !isResizeNeeded &&
+	direction := resizeDirection(currentPVC, newPVC)
+	if direction == 0 &&
```
vrutkovs:
```suggestion
	switch resizeDirection(currentPVC, newPVC) {
	case 0:
		// Size remains the same
		if equality.Semantic.DeepEqual(newPVC.Labels, currentPVC.Labels) &&
			isObjectMetaEqual(currentPVC, newPVC, prevPVC) {
			return nil
		}
	case -1:
		// PVC needs to be shrunk
		// do no return error
		// probably, user updated pvc manually
		// without applying this changes to the configuration.
		logger.WithContext(ctx).Info(fmt.Sprintf("cannot decrease PVC=%s size from=%s to=%s, please check VolumeClaimTemplate configuration", newPVC.Name, oldSize.String(), newSize.String()))
		return nil
	case 1:
		// PVC needs to be expanded

		// check if storage class is expandable
		expandable, err := isExpandable(ctx, rclient, newPVC)
		if err != nil {
			return fmt.Errorf("failed to check storageClass expandability for PVC %s: %v", newPVC.Name, err)
		}
		if !expandable {
			// don't return error to caller, since there is no point to requeue and reconcile this when sc is unexpandable
			sc := ptr.Deref(newPVC.Spec.StorageClassName, "default")
			logger.WithContext(ctx).Info(fmt.Sprintf("PVC=%s (StorageClass=%s) doesn't support live resizing", sc, newPVC.Name))
			return nil
		}
	default:
		return fmt.Errorf("unknown resize direction for PVC %s", newPVC.Name)
	}
```

vrutkovs (changes requested):
Also needs tests

internal/controller/operator/factory/reconcile/pvc.go
AndrewChubatiuk: this suggestion is invalid, as it doesn't fix an issue, would like to avoid having
```
logger.WithContext(ctx).Info(fmt.Sprintf("cannot decrease PVC=%s size from=%s to=%s, please check VolumeClaimTemplate configuration", newPVC.Name, oldSize.String(), newSize.String()))
```
if annotation is set to `false` and annotations validation only happens in isExpandable function

vrutkovs: Lets start small - rewrite with `switch` instead of nested `if`s.

>validation only happens in isExpandable function

It should be renamed to reflect its purpose - we call `isExpandable` when we're checking if the volume can be shrunk

AndrewChubatiuk: switch is not applicable in this case as it'll have case 0 and default

AndrewChubatiuk: updated logic, just added a check for pvc-autoresizer annotations to skip throwing a warning

internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
```diff
@@ -219,10 +219,14 @@ func mayGrow(ctx context.Context, pvc *corev1.PersistentVolumeClaim, newSize, ex
 	case 0:
 		return false
 	case -1:
-		// do no return error
-		// probably, user updated pvc manually
-		// without applying this changes to the configuration.
-		logger.WithContext(ctx).Info(fmt.Sprintf("cannot decrease PVC=%s size from=%s to=%s, please check VolumeClaimTemplate configuration", pvc.Name, existSize.String(), newSize.String()))
+		switch {
+		// do nothing if pvc-autoresize annotation is set
+		case pvc.Annotations["resize.topolvm.io/enabled"] == "true":
```
vrutkovs: Is it the right annotation? The bug report mentions `operator.victoriametrics.com/pvc-allow-volume-expansion` - its odd that we hardcode TopoLVM here

AndrewChubatiuk: `operator.victoriametrics.com/pvc-allow-volume-expansion` is annotations of our operator, which defines if operator can perform storage expansion (target size is bigger than actual pvc size). in issue user describes different case, when volume was already expanded by pvc-autoresize and operator now wants to decrease it's size. in implementation decided not to rely on vm operator annotation, as it has different meaning and use value of pvc-autoresize annotation

vrutkovs: Oh, I see. Lets update the CHANGELOG to mention the exact annotation we skip. I am worried that there would be requests to add more - so I'd prefer to require a custom, victoriametrics-scoped annotation, so that might be something we want to implement now

AndrewChubatiuk: checked two issues, where `operator.victoriametrics.com/pvc-allow-volume-expansion` was introduced and seems like with `operator.victoriametrics.com/pvc-allow-volume-expansion: true` it should force operator to expand storage and with `operator.victoriametrics.com/pvc-allow-volume-expansion: false` it should ignore any storage size changes and basically it was implemented for PVC resizer. Updated code once again

AndrewChubatiuk:
@cubic-dev-ai review this PR

PR #1761, opened 2026-01-27
vmalert: fail on absent notifiers

### LINKED ISSUE #1757: vmalert: using `notifiers` selector which do not match vmalertmanager
Currently, using vmalert with misconfigured `notifiers` which does not match anything leads to creating a vmalert pod with `notifiers.url` being empty. vmalert does not fail with this configuration (I'll create a separate issue for that), so combined with operator behaviour this leads to successfully running vmalert which will never send any alerts silently.

It would be great to have any sigh of error for such case, e.g. do not create a pod until matching pods of alertmanager do exist or something else to help the user to understand where the issue is.

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1757

---
## Summary by cubic
vmalert now returns an error if no notifier targets are found, instead of starting with an empty -notifier.url. This prevents silent misconfigurations; ExtraArgs notifier.url/notifier.config are respected. Fixes #1757.

- **Bug Fixes**
  - Return error when no notifiers are discovered via selectors or static config.
  - Do not append empty -notifier.url; use ExtraArgs notifier.url or notifier.config when provided.
  - Updated tests to reflect new behavior.

- **Migration**
  - Configure notifiers via spec.notifiers/spec.notifier, or set ExtraArgs: notifier.url or notifier.config.
  - Deployments that previously ran without notifiers will now fail until configured.

<sup>Written for commit fa4f4471ad5e0479c67767d08ebfbf5c2cdc3f36. Summary will update on new commits.</sup>


AndrewChubatiuk:
fixed tests

PR #1768, opened 2026-01-29
fix flag name in the docs

dmitryk-dk:
In the documentation, we defined a flag as `resyncInterval`, but in the code, it is named as `resync-interval`

https://github.com/VictoriaMetrics/operator/blob/master/cmd/config-reloader/main.go#L57

Screenshot 2026-01-29 at 13 01 14

Fixed docs to be consistent with the code.

---
## Summary by cubic
Corrected the flag name to resync-interval in docs and API comments to match the config-reloader binary. Replaced resyncInterval in all configReloaderExtraArgs examples so users pass the right flag.

<sup>Written for commit bc6e5670b3ef0fd81e929aa7b531e26af25397dd. Summary will update on new commits.</sup>


AndrewChubatiuk (approved):
Thanks for a fix!

PR #1769, opened 2026-01-29
upgraded tools in makefile

AndrewChubatiuk:
updated all tools in makefile
replaced `Deprecated:` prefixed comments with custom `deprecated` markers that are rendered as shown below
image

@makasim @vrutkovs what do you think about adding the same `+released` marker to inform about release since which parameter is available?

---
## Summary by cubic
Upgraded build tools and improved API/CRD docs by switching to custom deprecated markers and a pinned crd-ref-docs fork. Also increased doc render depth and regenerated client/informer code to v0.35 with context-aware watch support.

- **Refactors**
  - Replaced “Deprecated:” comments with +deprecated markers (with values) and updated the doc template to render deprecation notes.
  - Updated docs generation: added --max-depth=60, set Kubernetes version to 1.35, regenerated CRDs (controller-gen v0.20.0) and docs.
  - Regenerated clients/informers/listers to v0.35: moved to cache.ToListWatcherWithWatchListSemantics, added context-aware List/Watch, updated fake client to pass ListOptions and declare unsupported WatchList semantics, and simplified setConfigDefaults.

- **Dependencies**
  - Bumped tool versions: kustomize, controller-tools, envtest, golangci-lint, codegenerator, kind, OLM, operator-sdk, OPM, yq, ginkgo.
  - Switched crd-ref-docs to github.com/AndrewChubatiuk/crd-ref-docs and pinned.
  - CI: install libgpgme-dev in GitHub Actions before tests.

<sup>Written for commit f67f1c251058400887b903a715043eae2df594db. Summary will update on new commits.</sup>


makasim:
> @makasim @vrutkovs what do you think about adding the same +released marker to inform about release since which parameter is available?

It's reasanoble to have it in general. Though I might not know all the specifics of the operator repository development process

api/operator/v1beta1/vmextra_types.go
```diff
@@ -240,10 +240,6 @@ type EmbeddedObjectMetadata struct {
 // If neither `emptyDir` nor `volumeClaimTemplate` is specified, then by default an [EmptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) will be used.
 // +k8s:openapi-gen=true
 type StorageSpec struct {
-	// Deprecated: subPath usage will be disabled by default in a future release, this option will become unnecessary.
```
vrutkovs, line 243: I'd rather set `+deprecated={deprecated_in: "v0.61.0", removed_in: "v0.69.0"}` here instead

vrutkovs:
>what do you think about adding the same +released marker to inform about release since which parameter is available?

That seems to duplicate CHANGELOG functionality. Also, our docs are for the latest released version only, so this tag is useless for docs too. What other purposes for this tag do you see?

Makefile
```diff
@@ -414,7 +415,7 @@ install-tools: crd-ref-docs client-gen lister-gen informer-gen controller-gen ku
 .PHONY: crd-ref-docs
 crd-ref-docs: $(CRD_REF_DOCS)
 $(CRD_REF_DOCS): $(LOCALBIN)
-	$(call go-install-tool,$(CRD_REF_DOCS),github.com/elastic/crd-ref-docs,$(CRD_REF_DOCS_VERSION))
+	$(call go-install-tool,$(CRD_REF_DOCS),github.com/AndrewChubatiuk/crd-ref-docs,$(CRD_REF_DOCS_VERSION))
```
vrutkovs, line 419:
```suggestion#1769 
	# Required for custom markers - https://github.com/AndrewChubatiuk/crd-ref-docs/pull/1
	$(call go-install-tool,$(CRD_REF_DOCS),github.com/AndrewChubatiuk/crd-ref-docs,$(CRD_REF_DOCS_VERSION))
```

api/operator/v1beta1/vmextra_types.go
AndrewChubatiuk: this property is noop for a very long time and it's deprecated since 2020

Makefile
AndrewChubatiuk: https://github.com/elastic/crd-ref-docs/pull/194

AndrewChubatiuk:
> That seems to duplicate CHANGELOG functionality. Also, our docs are for the latest released version only, so this tag is useless for docs too. What other purposes for this tag do you see?

- people may use not up to date docs for opertor/k8s-stack installation, where operator may lack properties described in API.
- those who already uses k8s-stack/operator of old version and would like to enable some functionality on CR using property

it's hard to check changelog each time you need to enable a new property

api/operator/v1beta1/vmextra_types.go
vrutkovs: Ah, okay, lets remove it now then

Makefile
AndrewChubatiuk: added comment

vrutkovs (approved):

PR #1801, opened 2026-02-12
test: fixed allure attachments

AndrewChubatiuk:
fixed attachments in allure reports

---
## Summary by cubic
Replaced the external Allure integration with a built-in writer and binary-safe attachments to fix broken/missing e2e report artifacts. Results now write to ./allure-results, and failed tests attach a per-test crust-gather .tar.gz.

- **Bug Fixes**
  - Built-in Allure writer: outputs to ./allure-results, links tests/steps (incl. Before/After Suite), supports parallel runs and nested steps.
  - New AddAttachment API: binary-safe; crust-gather attached as application/gzip; per-test archives use a deterministic xxhash; works across parallel processes.
  - Switched ReportAfterSuite to allure.FromGinkgoReport; Makefile sets REPORTS_DIR and builds ./allure-report; CI uploads ./allure-report; .gitignore ignores allure-results/ and allure-report/.

- **Dependencies**
  - Removed ginkgo2allure and allure-go; added cespare/xxhash/v2.
  - Bumped sigs.k8s.io/structured-merge-diff/v6 to v6.3.2 and removed the temporary replace.

<sup>Written for commit 308e5c7ab63c6da7014c979b363e2f8bd674ca51. Summary will update on new commits.</sup>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,12 +4,12 @@ go 1.25.5
 
 require (
 	github.com/Masterminds/semver/v3 v3.4.0
-	github.com/Moon1706/ginkgo2allure v0.3.0
 	github.com/VictoriaMetrics/VictoriaMetrics v1.132.0
 	github.com/VictoriaMetrics/metrics v1.40.2
 	github.com/VictoriaMetrics/metricsql v0.84.8
 	github.com/VictoriaMetrics/operator/api v0.51.3
 	github.com/caarlos0/env/v11 v11.3.1
+	github.com/cespare/xxhash/v2 v2.3.0
 	github.com/fsnotify/fsnotify v1.9.0
 	github.com/go-logr/logr v1.4.3
 	github.com/go-test/deep v1.1.1
@@ -53,7 +53,6 @@ require (
 	github.com/aws/smithy-go v1.24.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/bmatcuk/doublestar/v4 v4.9.1 // indirect
-	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/emicklei/go-restful/v3 v3.12.2 // indirect
 	github.com/evanphx/json-patch v5.9.0+incompatible // indirect
@@ -75,7 +74,6 @@ require (
 	github.com/go-openapi/swag/typeutils v0.25.4 // indirect
 	github.com/go-openapi/swag/yamlutils v0.25.4 // indirect
 	github.com/go-task/slim-sprig/v3 v3.0.0 // indirect
-	github.com/goccy/go-json v0.10.5 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang-jwt/jwt/v5 v5.3.0 // indirect
 	github.com/google/btree v1.1.3 // indirect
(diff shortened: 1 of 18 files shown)


test/e2e/suite/allure/attachment.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+type MimeType string
+
+const (
+	MimeTypeJSON MimeType = "application/json"
+	MimeTypeCSV  MimeType = "text/csv"
+	MimeTypeText MimeType = "text/plain"
+	MimeTypePNG  MimeType = "image/png"
+	MimeTypeHTML MimeType = "text/html"
```
vrutkovs:
```suggestion
```
Since its a in-repo implementation we don't need to carry those - we'll be attaching archives only, others can be added later

test/e2e/suite/allure/attachment.go
```diff
@@ -0,0 +1,94 @@
...
+	case MimeTypeJSON:
+		return "json"
+	case MimeTypeCSV:
+		return "csv"
+	case MimeTypeHTML:
+		return "html"
+	case MimeTypePNG:
+		return "png"
+	case MimeTypeText:
+		return "txt"
```
vrutkovs:
```suggestion
```

test/e2e/suite/allure/label.go
```diff
@@ -0,0 +1,46 @@
...
+
+	labelSubSuite = "subSuite"
+
+	labelParentSuite = "parentSuite"
+
+	labelParametrized = "parametrized"
+
+	labelLink = "link"
+
+	labelLead = "lead"
```
vrutkovs: I don't think we need those anyway

test/e2e/suite/allure/api.go
```diff
@@ -0,0 +1,83 @@
+package allure
```
vrutkovs, line 6: We should preserve the copyright of the original implementation - in every file we reused

go.mod
vrutkovs:
```suggestion
```

vrutkovs (changes requested):

test/e2e/suite/allure/api.go
AndrewChubatiuk: there's no license in a repo

test/e2e/suite/allure/label.go
AndrewChubatiuk: removed some of them

test/e2e/suite/allure/api.go
vrutkovs: Just the copyright and the source then. Not having a license is problematic though, but we can handle it later

AndrewChubatiuk: do you have an example?

vrutkovs: See https://github.com/kubernetes/kubernetes/blob/master/pkg/api/v1/pod/util.go#L1-L15

There is a license though: https://github.com/ozontech/allure-go/blob/280c1bb29451100cb7ac93b2a43c9e2cf8fa2d95/LICENSE.md and https://github.com/Moon1706/ginkgo2allure/blob/main/LICENSE

AndrewChubatiuk: I was using this repo https://github.com/ramich2077/allure-ginkgo/

vrutkovs: oh, okay, lets remove the license part and use github nickname as a author attribution

AndrewChubatiuk: added header to each file

PR #1806, opened 2026-02-16
reconcile: fix skipped PVC update, when it should not be resized, but metadata changed

AndrewChubatiuk:
previously PVC update was skipped in case if storage is not resizable, while led to stale metadata

---
## Summary by cubic
Fixes PVC reconciliation so metadata updates are applied even when there’s no size change, a shrink is requested, or the storage class can’t expand. Prevents stale labels, annotations, and finalizers on existing PVCs.

- **Bug Fixes**
  - Apply metadata changes even when resize is skipped (no size delta, non-expandable storage class, or shrink attempt).
  - Read PVCExpandableLabel from the existing PVC; don’t shrink; log why resize is skipped.

- **Refactors**
  - Split logic into modifyPVC (computes changes, returns modified) and updatePVC (updates only when needed). Tests adjusted for finalizer retention and resourceVersion.

<sup>Written for commit 68466a7248f8d45b53f31b0f029da3b78e4fd6f1. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
+++ b/internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
@@ -152,7 +152,7 @@ func modifyPVC(ctx context.Context, rclient client.Client, existingObj, newObj, 
 			case "true":
 				expandable = true
 			default:
-				return false, fmt.Errorf("not expected value format for annotation=%q: %q, want true or false", vmv1beta1.PVCExpandableLabel, v)
+				return false, fmt.Errorf("unexpected value format for annotation=%q: %q, want true or false", vmv1beta1.PVCExpandableLabel, v)
 			}
 		}
 


internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
```diff
@@ -123,67 +123,77 @@ func updateSTSPVC(ctx context.Context, rclient client.Client, sts *appsv1.Statef
...
 		if ok {
 			switch strings.ToLower(v) {
 			case "false":
-				return nil
+				return metaChanged, nil
 			case "true":
 				expandable = true
 			default:
-				return fmt.Errorf("not expected value format for annotation=%q: %q, want true or false", vmv1beta1.PVCExpandableLabel, v)
+				return false, fmt.Errorf("not expected value format for annotation=%q: %q, want true or false", vmv1beta1.PVCExpandableLabel, v)
```
vrutkovs:
```suggestion
				return false, fmt.Errorf("unexpected value format for annotation=%q: %q, want true or false", vmv1beta1.PVCExpandableLabel, v)
```

Issue #1640, opened 2025-11-26
vmcluster maxUnavailable clashes with PDB


thejuan:
As part of the introduction of https://github.com/VictoriaMetrics/operator/pull/1458, evict was chosen over delete explicitly to honour PDBs.

This causes issues and confusion when distinct configuration is desired for distinct use cases i.e. a low maxUnavailable PDB control normal operation voluntary disruption like Karpenter and node draining. A high `maxUnavailable` is used to control the desired deployment rate (usually in conjuction with a multi AZ setup). In this case PDB will win and prevents a fast roll-out.

Kubernetes 1.35 is bringing [native maxUnavailable for statefulsets](https://github.com/kubernetes/enhancements/issues/961) which if integrated here will solve the issue. It could be a long time coming for users to get to that version.

I'd like to float the idea of making the use of evict/delete configurable when maxUnavailable is set

vrutkovs:
The goal for this potential change is to speed up vmstorage rollout - by default it uses graceful shutdown, which means an update rollout takes nodes one by one and thus takes quite a lot of time.

The alternative is described in https://docs.victoriametrics.com/victoriametrics/cluster-victoriametrics/#updating--reconfiguring-cluster-nodes - if we accept that read (not write!) disruption is acceptable for a short period of time, then we should update the operator to evict pod simultaneously. Instead of waiting for instances to update gracefully new instances are created in the shortest amount of time possible, so that we wouldn't have to wait for graceful shutdown and let the vmagents accumulate the queue. This avoids rows rerouting but has a significant drawback - read path would be disrupted until vmstorages are back (thus the focus on minimizing their downtime).

If this solution is acceptable, we don't need PDBs or maxUnavailable changes at all - the operator can handle this change on any configuration and doesn't need kes k8s features.

thejuan:
We have rerouting on, what we have seen in the past is that the first pods to come up will get overwhelmed. 
a value of 50% maybe better unless there is a way to control the impact of rerouting without disabling it.

We will test with this and report back

AndrewChubatiuk:
in version 0.68.2 with maxUnavailable 100% deletion will be applied to pods instead of eviction, which allows to bypass PDB restrictions for minimum downtime deployment scenario

PR #1815, opened 2026-02-18
reconcile: delete statefulset pods when maxUnavailable set to 100%

### LINKED ISSUE #1706: Ignore PodDisruptionBudget during minimal downtime upgrade strategy.
Currently, operator uses pod eviction during statefulSet rolling update. It makes hard to use podDisruptionBudget with it. 

Disruption budget is useful for vmstorage operation and it instructs other Kubernetes components how many replicas should be kept.
But for minimal downtime it's crucial to remove all pods at once, which pod disruption will prevent.

 Possible solution could be to switch for pod.Delete, if statefulset unavailable pods is equal to 100%.

AndrewChubatiuk:
fixes #1706, #1640

---
## Summary by cubic
Delete StatefulSet pods when MaxUnavailable is 100% for fast, predictable full restarts that bypass PDB limits. Rolling updates choose delete at 100% or evict otherwise; e2e now watches Deleted events via client.WithWatch to confirm; tests use controller-runtime interceptors and Go synctest; CHANGELOG updated.

- **New Features**
  - Delete pods instead of evicting when MaxUnavailable=100%; otherwise evict in batches sized by MaxUnavailable.

- **Bug Fixes**
  - Skip readiness wait when replicas are unset.
  - Treat not-found on delete as success.
  - Ignore terminating and non-StatefulSet-owned pods during update checks.

<sup>Written for commit e97896f4c634b9425d15208e1e693f42ace7a861. Summary will update on new commits.</sup>


AndrewChubatiuk:
@cubic-dev-ai review this PR

AndrewChubatiuk:
@cubic-dev-ai review this PR

internal/controller/operator/factory/reconcile/statefulset.go
```diff
@@ -286,27 +304,35 @@ func performRollingUpdateOnSts(ctx context.Context, podMustRecreate bool, rclien
...
+					} else {
+						// evict pod to trigger re-creation
+						podEviction := policyv1.Eviction{ObjectMeta: pod.ObjectMeta}
+						if err := rclient.SubResource("eviction").Create(ctx, &pod, &podEviction); err != nil {
+							// retry distruption interrupt error:
+							// https://github.com/kubernetes/kubernetes/blob/9a50e306361ea936e57fb6eb8c635f971e7bb707/pkg/registry/core/pod/storage/eviction.go#L418
+							if strings.Contains(err.Error(), "Cannot evict pod as it would violate the pod's disruption budget") {
+								return false, nil
+							}
+							return false, fmt.Errorf("cannot evict pod %s: %w", pod.Name, err)
```
vrutkovs:
```suggestion
						return true, nil
					}
					// evict pod to trigger re-creation
					podEviction := policyv1.Eviction{ObjectMeta: pod.ObjectMeta}
					if err := rclient.SubResource("eviction").Create(ctx, &pod, &podEviction); err != nil {
						// retry distruption interrupt error:
						// https://github.com/kubernetes/kubernetes/blob/9a50e306361ea936e57fb6eb8c635f971e7bb707/pkg/registry/core/pod/storage/eviction.go#L418
						if strings.Contains(err.Error(), "Cannot evict pod as it would violate the pod's disruption budget") {
							return false, nil
						}
						return false, fmt.Errorf("cannot evict pod %s: %w", pod.Name, err)
					}
```
Exit early on `o.delete`

internal/controller/operator/factory/reconcile/statefulset_test.go
```diff
@@ -262,38 +276,247 @@ func Test_podIsReady(t *testing.T) {
...
+		ctx := context.Background()
+		var mu sync.Mutex
+		var actions []string
+		fclient := k8stools.GetTestClientWithObjectsAndInterceptors(o.predefinedObjects, interceptor.Funcs{
+			SubResourceCreate: func(ctx context.Context, cl client.Client, _ string, obj client.Object, subResource client.Object, _ ...client.SubResourceCreateOption) error {
+				pod, podOk := obj.(*corev1.Pod)
+				_, evictionOk := subResource.(*policyv1.Eviction)
+				if evictionOk && podOk {
+					mu.Lock()
+					actions = append(actions, "Evict")
```
vrutkovs: Oh, I was hoping `fclient` would intercept `Evict` actions just like we do for `Get`, `Update` and `Create`? Is it because its targeted on a subresource?

internal/controller/operator/factory/reconcile/statefulset_test.go
```diff
@@ -262,38 +276,247 @@ func Test_podIsReady(t *testing.T) {
...
+				return cl.Get(ctx, key, obj)
+			},
+		})
+		err := performRollingUpdateOnSts(ctx, fclient, o.sts, o.opts)
 		if o.wantErr {
 			assert.Error(t, err)
 		} else {
 			assert.NoError(t, err)
 		}
+		assert.ElementsMatch(t, actions, o.actions)
```
vrutkovs: I don't think its a good idea to ignore the order of actions - why not `Equal`?

test/e2e/vmcluster_test.go
```diff
@@ -1638,7 +1638,7 @@ up{baz="bar"} 123
 					},
 				},
 			),
-			Entry("configures vmstorage with MaxUnavailable 100% but limited by a PDB", "maxunavailable-100-percent-pdb",
+			Entry("configures vmstorage with MaxUnavailable 100% and ignored PDB limitation", "maxunavailable-100-percent-pdb",
```
vrutkovs, line 1643: Can we check that the pods were removed - maybe catch events?

internal/controller/operator/factory/reconcile/statefulset_test.go
AndrewChubatiuk: yes

internal/controller/operator/factory/reconcile/statefulset_test.go
AndrewChubatiuk: because of concurrency order is not predictable. or there should be a list of actions per pod

AndrewChubatiuk: added actions map

PR #1819, opened 2026-02-18
distributed: do no exclude zone when specs are immutable

AndrewChubatiuk:
## Summary by cubic
Keep zones in the VMDistr load balancer when VMCluster/VMAgent specs are unchanged, and skip the pause between those zones. Harden pod readiness polling to be more resilient and clearer.

- **Bug Fixes**
  - VMDistr zones: detect per-zone VMCluster/VMAgent spec changes using k8s semantic equality (after defaulting); only exclude/update zones and pause when a diff exists. Unchanged zones stay in rotation.
  - Reconcile: waitForPodReady now tolerates NotFound during polling, handles deletion, improves error messages, and avoids nil pod handling.

<sup>Written for commit 98609014d25019f2d4d5c641adf27b65f6603cb5. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/vmdistributed/zone.go
+++ b/internal/controller/operator/factory/vmdistributed/zone.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/cespare/xxhash/v2"
 	discoveryv1 "k8s.io/api/discovery/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -32,6 +33,7 @@ type zones struct {
 	httpClient *http.Client
 	vmagents   []*vmv1beta1.VMAgent
 	vmclusters []*vmv1beta1.VMCluster
+	hasChanges []bool
 }
 
 func (zs *zones) Len() int {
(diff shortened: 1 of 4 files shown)


internal/controller/operator/factory/vmdistributed/zone.go
```diff
@@ -179,7 +185,7 @@ func (zs *zones) upgrade(ctx context.Context, rclient client.Client, cr *vmv1alp
 	nsnCluster := types.NamespacedName{Name: vmCluster.Name, Namespace: vmCluster.Namespace}
 	logger.WithContext(ctx).Info("reconciling VMCluster", "item", item, "name", nsnCluster)
 
-	if !vmCluster.CreationTimestamp.IsZero() && !vmAgent.CreationTimestamp.IsZero() {
+	if !vmCluster.CreationTimestamp.IsZero() && !vmAgent.CreationTimestamp.IsZero() && zs.hasChanges[i] {
```
vrutkovs:
```suggestion
	// vmCluster or vmAgent have been created
	needsLBUpdate := !vmCluster.CreationTimestamp.IsZero() && !vmAgent.CreationTimestamp.IsZero()
	// No vmcluster or vmagent spec changes required
	if zs.hasChanges[i] {
		needsLBUpdate = true
	}

	if needsLBUpdate {
```
lets add some more details here

vrutkovs: Perhaps we should split `hasChanges` into `hasVMClusterChanges` and `hasVMAgentChanges` so that later we could omit VMCluster/VMAgent reconciles (however they can apply metadata only?). What do you think?

AndrewChubatiuk: reconcile step also checks status, specs can be up to date, but resources can be still in a process of creation/updating

vrutkovs: Right, agreed, lets keep it as is

PR #1829, opened 2026-02-19
fix: update vmselect/vminsert service selector when RequestsLoadBalancer is set

### LINKED ISSUE #1814: VMDistributed: point to vmauth when requestsLoadBalancer is enabled
VMCluster may have vmauth requestsLoadBalancer enabled, so created vmagents should be pointing to vmauth instead of vminserts directly

vrutkovs:
Make sure that RequestsLoadBalancer setting changes labels and selector of vminsert/vmselect internal services and points to lb instead of vmselect/vminsert services directly

Fixes #1814

---
## Summary by cubic
Update vmselect and vminsert internal services to route through the vmauth RequestsLoadBalancer when enabled. This ensures traffic goes via the LB instead of directly to vmselect/vminsert pods.

- **Bug Fixes**
  - Set Service.Spec.Selector to the balancer (ClusterComponentBalancer) for vmselect and vminsert when RequestsLoadBalancer is enabled.
  - Added tests covering both insert and select services to verify selector and label updates.

<sup>Written for commit 3e4808f4962197e28f93d13355bc46b5ff15eba0. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/vmcluster/vmcluster.go
+++ b/internal/controller/operator/factory/vmcluster/vmcluster.go
@@ -200,6 +200,7 @@ func buildVMSelectService(cr *vmv1beta1.VMCluster) *corev1.Service {
 		svc.Spec.ClusterIP = corev1.ClusterIPNone
 		svc.Spec.Type = corev1.ServiceTypeClusterIP
 		svc.Labels[vmv1beta1.VMAuthLBServiceProxyJobNameLabel] = b.PrefixedName()
+		svc.Spec.Selector = cr.SelectorLabels(vmv1beta1.ClusterComponentBalancer)
 	}
 
 	return svc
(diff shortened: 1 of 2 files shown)


internal/controller/operator/factory/vmcluster/vmcluster.go
AndrewChubatiuk, line 338: when requests load balancer enabled this service has a different name and vmauth uses it to query insert instances, same is for select, with this change vmauth will be querying itself

vrutkovs: Oh, this changes the _internal_ VMInsert / VMSelect service, which is already pointing to VMInsert / VMInsert. But aren't we creating a new service for LB (in `createOrUpdateLBProxyService`?) and we should be pointing "external" service to it as well - I can't find the place where this happens

AndrewChubatiuk: when requestsloadbalancer is enabled `createOrUpdateLBProxyService` creates "external" vminsert/vmselect services that target vmauth with the same name as `buildVMSelectService`/`buildVMInsertService` create, when requestsLoadBalancer is disabled

vrutkovs: Oh, I see. In that case, #1814 is probably an invalid issue - I'll doublecheck vmauth metrics to see if the traffic is flowing throuigh them correctly

vrutkovs:
Verified the metrics - the requests are passing through vmauth as expected

Issue #1832, opened 2026-02-19
Dry-run mode


vrutkovs:
Changes in the operator may be destructive, the operator should have a mode where diffs are printed, but not applied

AndrewChubatiuk:
looks like a feature that requires UI

vrutkovs:
There are several options here:
* print diffs to the log
    Can we ensure the diff would be pretty-printed? See also #1821 
* Store it in the object annotation

Perhaps we should research whether other operators are doing something similar

AndrewChubatiuk:
Suppose its better to make a cli tool that returns a pretty diff between already deployed resources  and desired instead of making logging more noisy (want treat this and diff improve as completely different tasks). Additionally it allows to check diff before operator upgrade. It can be an operator subcommand 

operator diff vmagent test -n default

vrutkovs:
My idea was:
* before upgrade, users turn on dry-run mode
* operator upgrade
* operator prints diffs into the log (#1821), possibly printing whether pod rollouts will occur explicitly
* administrator waits until the maintenance period starts, switches the dry run mode off
* rollouts are performed, cluster state is reconciled, maintenance ends

AndrewChubatiuk:
With a separate cli tool there's no need to pause operator and it possible to generate a terraform like fancy diff with list of resources that are expected to me created, removed, updated or recreated

AndrewChubatiuk:
released in 0.68.4

PR #1882, opened 2026-02-24
vmoperator: set default -http.shutdownDelay for HTTP-serving components

### LINKED ISSUE #1834: vmoperator: set default -http.shutdownDelay for HTTP-serving components
Many VictoriaMetrics components serve `/health` (readiness) and user traffic on the same HTTP server. During pod termination this can cause brief disconnects if traffic is still routed to the terminating pod.

VictoriaMetrics supports `-http.shutdownDelay`, which returns non-OK responses from `/health` for a grace period before shutting down the HTTP server, giving kube/LBs/meshes time to stop routing.

Request:

- Have `vmoperator` automatically add `-http.shutdownDelay` by default for applicable components (e.g. `VMAgent`, `VMSingle`, `VMCluster` components like `vmselect`/`vminsert`/`vmstorage`, `VMAlert`, `VMAuth`, etc.). Do not override a user-specified `spec.extraArgs.http.shutdownDelay` (or component-specific `extraArgs`).
- Choose a reasonable default (optionally derived from the `readinessProbe` settings).

endesapt:
Fixes: #1834

Created AddHTTPShutdownDelayArg func that adds -http.shutdownDelay based on the factors:
- users http.shutdownDelay extra argument
- readinessProbe.PeriodSeconds(or 5 seconds as default)*readinessProbe.FailiureTreshold(or 10 by default)
- 50 seconds by default

Applied it to every component that has http.shutdownDelay flag

---
## Summary by cubic
Switches HTTP-serving components to a Kubernetes preStop SleepAction (15s) so load balancers can drain before pods stop, replacing the `-http.shutdownDelay` flag. Adds per-CR control, a global toggle, and sets `terminationGracePeriodSeconds` to 30s by default.

- **New Features**
  - Injects a preStop SleepAction (Kubernetes ≥1.29) for VMAgent, VMAlert, VMAuth (incl. LB), VMSingle, VLAgent, VLSingle, and VM/VL/VT cluster Select/Insert; skips storage components and respects existing preStop hooks.
  - Adds `preStopSleepSeconds` to CRs (default 15s when `VM_ENABLE_DEFAULT_PRESTOP_HOOK=true`; `0` disables). Skips on unsupported Kubernetes versions and validates it is less than `terminationGracePeriodSeconds`.
  - Adds `VM_ENABLE_DEFAULT_PRESTOP_HOOK` (default: true). Sets default `terminationGracePeriodSeconds` to 30s.

- **Migration**
  - Set `VM_ENABLE_DEFAULT_PRESTOP_HOOK=false` before upgrading to avoid rolling updates; re-enable after.
  - Tune or disable via `preStopSleepSeconds` in CRs; note the new 30s default for `terminationGracePeriodSeconds`.

<sup>Written for commit e949fbf67dfd11ce02f9372b83ebfbbb92108b28. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VictoriaMetrics/operator/pull/1882?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


internal/controller/operator/factory/build/container_test.go
```diff
@@ -205,6 +205,30 @@ func Test_addExtraArgsOverrideDefaults(t *testing.T) {
 	})
 }
 
+func TestAddHTTPShutdownDelayArg(t *testing.T) {
+	t.Run("adds default derived from built-in readiness defaults", func(t *testing.T) {
```
AndrewChubatiuk: please rewrite this using f-test pattern, like other tests even in this file

endesapt: Thanks, done

AndrewChubatiuk:
@endesapt @vrutkovs 
should it also depend on terminationGracePeriodSeconds?

internal/controller/operator/factory/build/container.go
```diff
@@ -216,6 +216,35 @@ func AddExtraArgsOverrideDefaults(args []string, extraArgs map[string]string, da
...
+
+// AddHTTPShutdownDelayArg adds default -http.shutdownDelay flag if user didn't override it in extraArgs.
+// If readiness probe is provided, delay is derived from readiness probe timings.
+func AddHTTPShutdownDelayArg(args []string, extraArgs map[string]string, probes *vmv1beta1.EmbeddedProbes) []string {
+	if _, ok := extraArgs["http.shutdownDelay"]; ok {
+		return args
+	}
+
+	delaySeconds := defaultReadinessPeriodSeconds * defaultReadinessFailureThreshold
+	if probes != nil && probes.ReadinessProbe != nil {
```
vrutkovs: I don't think it's a good idea - this makes the results less predictable. `terminationGracePeriodSeconds` would be a factor which would influence this result.

Instead, I propose to set a valid default value (possibly separate for each component?) and set readinessPeriod / terminationGracePeriodSeconds unless they are already defined.

Also, we should apply this to new instances only - we should avoid unnecessary rollouts after operator upgrades

internal/controller/operator/factory/build/container_test.go
```diff
@@ -205,6 +205,42 @@ func Test_addExtraArgsOverrideDefaults(t *testing.T) {
...
+		probes:    nil,
+		wantArgs:  []string{"-http.shutdownDelay=50s"},
+	})
+	f(opts{
+		extraArgs: map[string]string{"http.shutdownDelay": "5s"},
+		probes:    nil,
+		wantArgs:  nil,
+	})
+	f(opts{
+		extraArgs: nil,
```
vrutkovs: Test to add: shutdownDelay and probe both set

vrutkovs (changes requested):

internal/controller/operator/factory/build/container.go
endesapt: I investigated a little bit and i am now concerned, if we even need readiness probe for that? As i read [Pod Termination Flow](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination-flow) it is said that when pod is terminated, it is instantly marked in EndpointSlices as `ready:false` which prevents all the traffic to it? So we need to rely only on `terminationGracePeriodSeconds` to understand how much time we have to process traffic that we already recieved.
So i guess the final scheme will be without readinessProbe and will be:
1. if `spec.terminationGracePeriodSeconds` is not set, set it to default(for each component or united)
2. `http.shutdownDelay` is inherited from it if not defined by user.

endesapt: > (possibly separate for each component?) 

If i didn't miss anything, i can change the code accordingly, but i dont know what the values can be for each component (or even what the unified value should be), because i dont have much experience in maintaining VM components. Would be glad if you help me with that.

I can just start with setting unified default to 30s and then we will change it to appropriate value

vrutkovs: >i am now concerned, if we even need readiness probe for that?

No, not necessarily, but my understanding is that readinessProbe needs to align with shutdownDelay - when the container goes down, we'll run x readiness checks + terminationGracePeriodSeconds. We want to minimize the downtime, so readinessProbe needs to run quite often (its cheap), wait for 2 failure and only afterwards mark the container as failed.
I don't think we need to adjust probe timings here, but if we do - we should do that here and preferably for new resources only.

>I can just start with setting unified default to 30s and then we will change it to appropriate value

That sounds reasonable, make sure the implementation makes it easy to tweak it for, say, vmstorage later easily

endesapt: > That sounds reasonable, make sure the implementation makes it easy to tweak it for, say, vmstorage later easily

Understand

>we'll run x readiness checks + terminationGracePeriodSeconds. We want to minimize the downtime, so readinessProbe needs to run quite often (its cheap), wait for 2 failure and only afterwards mark the container as failed.

So still for `shutdownDelay` we will have the scheme i described earlier right? Readiness probe will fail during the termination cause of the flag, and it will be marked as `ready: false; serving: "false"` in EndpointSlice stopping the traffic.

I am talking about this:

1. if `spec.terminationGracePeriodSeconds` is not set, set it to default(for each component or united)
2. `http.shutdownDelay` is inherited from it if not defined by user.

> for new resources only.

I will try to do that, but it would be great if you can give some directions on how to do that? How to distinct new resources from older ones? By their annotations maybe?

vrutkovs: >we will have the scheme i described earlier right?

Yes, I think we're on the same page. I don't think readiness probes need to be changed for now.

>How to distinct new resources from older ones?

Most reconcile procedured have prevCR var - if its nil its an empty CR

endesapt:
Changed `AddHTTPShutdownDelayArg` to work as we discussed and work only for new resources. 
But for now i didnt change the `terminationGracePeriodSeconds`, cause default value is still 30 seconds.

I wanted to set it explicitly in daemonset,deployment,statefulset like that:

https://github.com/VictoriaMetrics/operator/pull/1882/changes/f1e4300d5bba09abba0b241408fe03095aafc65a#diff-e7d6e0f45db6d4efd51be09d7ff37703298db2a37b4b0ee87054f1c549d77176R26-R28

But then i understood that it will make old deployments restart. I can try pass the check to the functions, but i think it will be messy. Do we need it?

As for changing defaults for distinct components, it will be pretty easy. As `httpShutdownDelay` relies on `terminationGracePeriod`, we can change default value for each component like this:

https://github.com/VictoriaMetrics/operator/blob/cf1840949906dde52fadb59e694893a310a705de/internal/controller/operator/factory/build/defaults.go#L339-L341

internal/controller/operator/factory/build/container.go
```diff
@@ -216,6 +216,34 @@ func AddExtraArgsOverrideDefaults(args []string, extraArgs map[string]string, da
...
+	if !isNewResource {
+		return args
+	}
+
+	var delaySeconds int64
+	if terminationGracePeriodSeconds != nil {
+		delaySeconds = *terminationGracePeriodSeconds
+	} else {
+		delaySeconds = DefaultTerminationGracePeriodSeconds
+	}
```
vrutkovs:
```suggestion
	delaySeconds := DefaultTerminationGracePeriodSeconds
	if terminationGracePeriodSeconds != nil {
		delaySeconds = *terminationGracePeriodSeconds
	}
```

internal/controller/operator/factory/build/defaults.go
vrutkovs:
```suggestion
			spec.VMStorage.TerminationGracePeriodSeconds = &DefaultTerminationGracePeriodSeconds
```

internal/controller/operator/factory/build/defaults.go
```diff
@@ -343,7 +343,7 @@ const (
 	vmStorageDefaultDBPath = "vmstorage-data"
 )
 
-var defaultTerminationGracePeriod = int64(30)
+var defaultTerminationGracePeriod = DefaultTerminationGracePeriodSeconds
```
vrutkovs:
```suggestion
```

internal/controller/operator/factory/build/daemonset.go
```diff
@@ -26,7 +26,6 @@ func DaemonSetAddCommonParams(dst *appsv1.DaemonSet, useStrictSecurity bool, par
 	dst.Spec.Template.Spec.TerminationGracePeriodSeconds = params.TerminationGracePeriodSeconds
 	dst.Spec.Template.Spec.TopologySpreadConstraints = params.TopologySpreadConstraints
 	dst.Spec.Template.Spec.ImagePullSecrets = params.ImagePullSecrets
-	dst.Spec.Template.Spec.TerminationGracePeriodSeconds = params.TerminationGracePeriodSeconds
```
vrutkovs: Perhaps we should leave defaulting (its useful for other purposes too) and then we can rely on `terminationGracePeriodSeconds` always being set?

vrutkovs (changes requested):
LGTM on the approach, few comments on the code

internal/controller/operator/factory/build/daemonset.go
endesapt: I deleted the 29th line because it is just the copy of 26 line and this repeats in each file (daemonset, deployment, statefulset)
About setting the defaults, I thought that setting `terminationGracePeriodSeconds` will force the restart of the pods, but changing it from nothing to 30 will not restart them, because it is the default value in k8s.
So i can actually return the defaulting logic without any risks i guess

endesapt: I will change the code according to your comments

internal/controller/operator/factory/build/container.go
endesapt: I think we can change it to just 
```golang
delaySeconds := *terminationGracePeriodSeconds
```
cause we are planning to always set the default value if not already set by user

internal/controller/operator/factory/build/defaults.go
endesapt: In case we make default values, we can remove this at all i guess

internal/controller/operator/factory/build/container.go
endesapt: No, we can't because `DeploymentAddCommonParams` comes after setting args for the pod itself, didnt see that

endesapt:
Hope i got you right

vrutkovs (commented):
LGTM, @AndrewChubatiuk could you have a look? I'm not sure if defaulting style is ciorrect

internal/controller/operator/factory/vtsingle/vtsingle.go
```diff
@@ -227,6 +227,8 @@ func makePodSpec(r *vmv1.VTSingle) (*corev1.PodTemplateSpec, error) {
 			MountPath: path.Join(vmv1beta1.ConfigMapsDir, c),
 		})
 	}
+
+	args = build.AddHTTPShutdownDelayArg(args, r.Spec.ExtraArgs, r.Spec.TerminationGracePeriodSeconds, r.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, r.Spec.ExtraArgs, r.Spec.TerminationGracePeriodSeconds, r.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vtcluster/storage.go
```diff
@@ -304,6 +304,7 @@ func buildVTStoragePodSpec(cr *vmv1.VTCluster) (*corev1.PodTemplateSpec, error)
 		})
 	}
 
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.Storage.ExtraArgs, cr.Spec.Storage.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.Storage.ExtraArgs, cr.Spec.Storage.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vtcluster/select.go
```diff
@@ -304,6 +304,7 @@ func buildVTSelectPodSpec(cr *vmv1.VTCluster) (*corev1.PodTemplateSpec, error) {
 		})
 	}
 
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.Select.ExtraArgs, cr.Spec.Select.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.Select.ExtraArgs, cr.Spec.Select.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vtcluster/insert.go
```diff
@@ -197,6 +197,7 @@ func buildVTInsertPodSpec(cr *vmv1.VTCluster) (*corev1.PodTemplateSpec, error) {
 		})
 	}
 
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.Insert.ExtraArgs, cr.Spec.Insert.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.Insert.ExtraArgs, cr.Spec.Insert.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vmsingle/vmsingle.go
```diff
@@ -247,7 +247,7 @@ func makeSpec(ctx context.Context, cr *vmv1beta1.VMSingle) (*corev1.PodTemplateS
 
 	volumes, vmMounts = build.LicenseVolumeTo(volumes, vmMounts, cr.Spec.License, vmv1beta1.SecretsDir)
 	args = build.LicenseArgsTo(args, cr.Spec.License, vmv1beta1.SecretsDir)
-
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.ExtraArgs, cr.Spec.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.ExtraArgs, cr.Spec.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vmcluster/vmcluster.go
```diff
@@ -1511,6 +1514,7 @@ func buildVMAuthLBDeployment(cr *vmv1beta1.VMCluster) (*appsv1.Deployment, error
 	volumes, vmounts = build.LicenseVolumeTo(volumes, vmounts, cr.Spec.License, vmv1beta1.SecretsDir)
 	args = build.LicenseArgsTo(args, cr.Spec.License, vmv1beta1.SecretsDir)
 
+	args = build.AddHTTPShutdownDelayArg(args, spec.ExtraArgs, spec.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, spec.ExtraArgs, spec.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vmcluster/vmcluster.go
```diff
@@ -1084,6 +1086,7 @@ func makePodSpecForVMStorage(ctx context.Context, cr *vmv1beta1.VMCluster) (*cor
 	volumes, vmMounts = build.LicenseVolumeTo(volumes, vmMounts, cr.Spec.License, vmv1beta1.SecretsDir)
 	args = build.LicenseArgsTo(args, cr.Spec.License, vmv1beta1.SecretsDir)
 
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VMStorage.ExtraArgs, cr.Spec.VMStorage.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VMStorage.ExtraArgs, cr.Spec.VMStorage.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vmcluster/vmcluster.go
```diff
@@ -863,6 +864,7 @@ func makePodSpecForVMInsert(cr *vmv1beta1.VMCluster) (*corev1.PodTemplateSpec, e
 	volumes, vmMounts = build.LicenseVolumeTo(volumes, vmMounts, cr.Spec.License, vmv1beta1.SecretsDir)
 	args = build.LicenseArgsTo(args, cr.Spec.License, vmv1beta1.SecretsDir)
 
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VMInsert.ExtraArgs, cr.Spec.VMInsert.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VMInsert.ExtraArgs, cr.Spec.VMInsert.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vmcluster/vmcluster.go
```diff
@@ -658,6 +658,7 @@ func makePodSpecForVMSelect(cr *vmv1beta1.VMCluster) (*corev1.PodTemplateSpec, e
 	volumes, vmMounts = build.LicenseVolumeTo(volumes, vmMounts, cr.Spec.License, vmv1beta1.SecretsDir)
 	args = build.LicenseArgsTo(args, cr.Spec.License, vmv1beta1.SecretsDir)
 
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VMSelect.ExtraArgs, cr.Spec.VMSelect.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VMSelect.ExtraArgs, cr.Spec.VMSelect.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vmauth/vmauth.go
```diff
@@ -325,6 +325,7 @@ func makeSpecForVMAuth(cr *vmv1beta1.VMAuth) (*corev1.PodTemplateSpec, error) {
 		return nil, fmt.Errorf("cannot apply patch for initContainers: %w", err)
 	}
 
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.ExtraArgs, cr.Spec.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.ExtraArgs, cr.Spec.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vmalert/vmalert.go
```diff
@@ -523,8 +523,9 @@ func buildArgs(cr *vmv1beta1.VMAlert, ruleConfigMapNames []string, ac *build.Ass
 	}
 
 	args = build.LicenseArgsTo(args, cr.Spec.License, vmv1beta1.SecretsDir)
-
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.ExtraArgs, cr.Spec.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.ExtraArgs, cr.Spec.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -664,6 +664,7 @@ func newPodSpec(cr *vmv1beta1.VMAgent, ac *build.AssetsCache) (*corev1.PodSpec,
 	args = build.StreamAggrArgsTo(args, "streamAggr", streamAggrKeys, streamAggrConfigs...)
 
 	args = build.AppendArgsForInsertPorts(args, cr.Spec.InsertPorts)
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.ExtraArgs, cr.Spec.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.ExtraArgs, cr.Spec.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vlsingle/vlsingle.go
```diff
@@ -237,6 +237,7 @@ func makePodSpec(r *vmv1.VLSingle) (*corev1.PodTemplateSpec, error) {
 	volumes, vmMounts = build.LicenseVolumeTo(volumes, vmMounts, r.Spec.License, vmv1beta1.SecretsDir)
 	args = build.LicenseArgsTo(args, r.Spec.License, vmv1beta1.SecretsDir)
 
+	args = build.AddHTTPShutdownDelayArg(args, r.Spec.ExtraArgs, r.Spec.TerminationGracePeriodSeconds, r.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, r.Spec.ExtraArgs, r.Spec.TerminationGracePeriodSeconds, r.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vlcluster/vmauth_lb.go
```diff
@@ -161,6 +161,7 @@ func buildVMauthLBDeployment(cr *vmv1.VLCluster) (*appsv1.Deployment, error) {
 	volumes, vmMounts = build.LicenseVolumeTo(volumes, vmMounts, cr.Spec.License, vmv1beta1.SecretsDir)
 	args = build.LicenseArgsTo(args, cr.Spec.License, vmv1beta1.SecretsDir)
 
+	args = build.AddHTTPShutdownDelayArg(args, spec.ExtraArgs, spec.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, spec.ExtraArgs, spec.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vlcluster/vlstorage.go
```diff
@@ -310,6 +310,7 @@ func buildVLStoragePodSpec(cr *vmv1.VLCluster) (*corev1.PodTemplateSpec, error)
 		})
 	}
 
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VLStorage.ExtraArgs, cr.Spec.VLStorage.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VLStorage.ExtraArgs, cr.Spec.VLStorage.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vlcluster/vlselect.go
```diff
@@ -308,6 +308,7 @@ func buildVLSelectPodSpec(cr *vmv1.VLCluster) (*corev1.PodTemplateSpec, error) {
 		})
 	}
 
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VLSelect.ExtraArgs, cr.Spec.VLSelect.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VLSelect.ExtraArgs, cr.Spec.VLSelect.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vlcluster/vlinsert.go
```diff
@@ -207,6 +207,7 @@ func buildVLInsertPodSpec(cr *vmv1.VLCluster) (*corev1.PodTemplateSpec, error) {
 		})
 	}
 
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VLInsert.ExtraArgs, cr.Spec.VLInsert.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.VLInsert.ExtraArgs, cr.Spec.VLInsert.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/vlagent/vlagent.go
```diff
@@ -427,6 +427,7 @@ func newPodSpec(cr *vmv1.VLAgent) (*corev1.PodSpec, error) {
 		agentVolumeMounts = append(agentVolumeMounts, cvm)
 	}
 	volumes, agentVolumeMounts = addRemoteWriteAssetsToVolumes(volumes, agentVolumeMounts, cr)
+	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.ExtraArgs, cr.Spec.TerminationGracePeriodSeconds, cr.ParsedLastAppliedSpec == nil)
```
AndrewChubatiuk:
```suggestion
	args = build.AddHTTPShutdownDelayArg(args, cr.Spec.ExtraArgs, cr.Spec.TerminationGracePeriodSeconds, cr.Status.LastAppliedSpec == nil)
```

internal/controller/operator/factory/build/container.go
AndrewChubatiuk: > Also, we should apply this to new instances only - we should avoid unnecessary rollouts after operator upgrades

@vrutkovs i don't get your idea. why do we need it?

endesapt:
brought branch up to date to master, edited LastAppliedSpec according to the review

endesapt:
@vrutkovs @AndrewChubatiuk Can you approve requested changes if everything is alright?

internal/controller/operator/factory/build/container.go
endesapt: If we apply new args as `-http.shutdownDelay` to Deployment spec, it will force Pods to rolling update. And as we are adding this to almost every vm operator component, it will make a lot of rolling updates.
I guess it is a lot of pressure on a cluster with lots of vm deployments that's why we are trying to avoid that.

vrutkovs:
This change looks good, so no worries about that but I'm afraid I misinformed you: we would like to ensure that operator update will not cause unnecessary pod rollouts. This is a longstanding task and not limited to this PR (this is why I asked to update the PR and apply this change to new resources only).

However, it appears this decision will later bite us - we'll have to handle removing this argument too, so its getting more and more complicated. The radical solution is to implement "dry run mode" - #1832, where users would be able to verify that operator update will cause rollouts and can plan their maintenance accordingly. 

As a result, we'll have to change the directions _again_ - no need to take the "freshness" of the resource into account, we'll tackle the unexpected rollouts via different means. Sorry about this taking too long - we accidentally tripped a long-standing problem, nothing wrong with your code. 

I understand if you want to leave the PR as is - we'll update it accordingly in that case, no problem with that.

endesapt:
> I understand if you want to leave the PR as is

I like to contribute to VictoriaMetrics - it is my way to learn and to say thanks to your project(i hope it helps). Changes is a part of it, so it is okay. I will revert the "freshness" when i will have time.

AndrewChubatiuk (dismissed):
LGTM! Will merge this PR now without any changes to http.shutdownDelay or preStop hook as it already contains many changes/fixes, will update issue later with possible solution for it

endesapt:
>will update issue later with possible solution for it

Nice! Will be waiting for updates and would be glad to discuss and to implement our final decision later

AndrewChubatiuk:
hey @vrutkovs @endesapt 
pushed commit to this PR, which replaces http.shutdownDelay flag with preStop hooks. which seems to be a more native to k8s solution than adding http.shutdownDelay

internal/controller/operator/factory/build/container.go
```diff
@@ -192,6 +192,25 @@ func AddExtraArgsOverrideDefaults(args []string, extraArgs map[string]string, da
 	return args
 }
 
+// Lifecycle applies default lifecycle hooks to the container from params.
+// It adds a preStop sleep hook to prevent traffic loss during pod termination,
+// unless the user has already set a preStop hook or preStopSleepSeconds is zero.
+func Lifecycle(container *corev1.Container, params *vmv1beta1.CommonAppsParams) {
+	if params == nil || params.PreStopSleepSeconds == nil || *params.PreStopSleepSeconds <= 0 {
```
vrutkovs, line 201: We need to make sure we're not exceeding `terminationGracePeriodSeconds` otherwise the pod would get SIGKILL in preStop

internal/controller/operator/factory/build/container.go
```diff
@@ -192,6 +192,25 @@ func AddExtraArgsOverrideDefaults(args []string, extraArgs map[string]string, da
...
+		return
+	}
+	if container.Lifecycle == nil {
+		container.Lifecycle = &corev1.Lifecycle{}
+	}
+	if container.Lifecycle.PreStop == nil {
+		container.Lifecycle.PreStop = &corev1.LifecycleHandler{
+			Exec: &corev1.ExecAction{
```
vrutkovs: This would not work on scratch images:
```
$ podman run --rm --entrypoint /bin/sleep -ti quay.io/victoriametrics/vmselect:v1.144.0-enterprise-cluster-scratch       
Trying to pull quay.io/victoriametrics/vmselect:v1.144.0-enterprise-cluster-scratch...
Getting image source signatures
Copying blob e2ff00bd30fc done   | 
Copying blob cd3a2503f8ac done   | 
Copying config d4d4216138 done   | 
Writing manifest to image destination
Error: crun: executable file `/bin/sleep` not found: No such file or directory: OCI runtime attempted to invoke a command that was not found
```
We should use `Sleep` instead of `Exec` here

vrutkovs (changes requested):
I think we're not setting this for V*Storages?

internal/controller/operator/factory/build/container.go
AndrewChubatiuk: good point, preStop native sleep is available starting k8s 1.29. makes sense to use native sleep and add default preStop hooks only if both global option is true and k8s version >= 1.29

AndrewChubatiuk: updated implementation

internal/controller/operator/factory/build/container.go
AndrewChubatiuk: added validation for this case

AndrewChubatiuk:
> I think we're not setting this for V*Storages?

by default there's not preStop hook for storages, it makes no sense, but it's possible to configure it manually

vrutkovs (commented):
LGTM, lets add a CHANGELOG entry about this

AndrewChubatiuk:
@vrutkovs added an entry

test/e2e/suite/suite.go
```diff
@@ -56,6 +56,7 @@ func GetClient(data []byte) client.WithWatch {
 	Expect(dec.Decode(&cfg)).To(Succeed())
 	// operator settings
 	envs := map[string]string{
+		"VM_ENABLE_DEFAULT_PRESTOP_HOOK":                 "false",
```
vrutkovs, line 59: This needs to be included in every test for now - this flag won't be required once we release the new version

docs/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ aliases:
...
+* FEATURE: [vmoperator](https://docs.victoriametrics.com/operator/): add default `preStop` lifecycle hook (15s sleep) to applicable component pods to prevent traffic loss during pod termination on rolling updates. Requires Kubernetes >= 1.29. Controlled by `VM_ENABLE_DEFAULT_PRESTOP_HOOK` env var (default: `true`). Set to `false` before upgrading to avoid triggering rolling updates on existing resources. Storage components (`VMStorage`, `VLStorage`, `VTStorage`) are excluded as their clients handle retries. The sleep duration is configurable via `spec.preStopSleepSeconds` on any CR; setting it to `0` disables the hook. Validation rejects `preStopSleepSeconds >= terminationGracePeriodSeconds`. See [#1834](https://github.com/VictoriaMetrics/operator/issues/1834).
```
vrutkovs: I think we need `*Update note*` here too - this would cause pod restarts for the majority of the users

AndrewChubatiuk: done

vrutkovs (approved):

AndrewChubatiuk:
thanks @endesapt 
for initial work on this issue

PR #1883, opened 2026-02-24
tests: replace most t.Fatalf calls

vrutkovs:
Testify asserts produce a significantly better output

---
## Summary by cubic
Replaced most t.Fatalf/t.Errorf in tests with testify assertions for clearer failures and cleaner test code. No functional changes; tests only.

- **Refactors**
  - Switched manual checks to assert.NoError/assert.Error/assert.Equal and related helpers.
  - Replaced reflect.DeepEqual with assert.Equal and removed reflect imports.
  - Standardized validations with assert.Contains/assert.Empty/assert.NotNil/assert.Len/assert.True.
  - Used assertions to simplify control flow instead of manual if/return paths.
  - Applied across v1beta1 API and operator factory tests (build, converter, k8stools, reconcile, vmagent, vmalert, vmalertmanager, vlagent, vlsingle, vtsingle, vmauth, vmanomaly).

<sup>Written for commit e9909ed2ad066068ed83118e6738dd220a687cc3. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/vlsingle/vlsingle_test.go
+++ b/internal/controller/operator/factory/vlsingle/vlsingle_test.go
@@ -32,9 +32,10 @@ func TestCreateOrUpdateVLSingle(t *testing.T) {
 		t.Helper()
 		fclient := k8stools.GetTestClientWithObjects(o.predefinedObjects)
 		err := CreateOrUpdate(context.TODO(), fclient, o.cr)
-		if (err != nil) != o.wantErr {
-			t.Errorf("CreateOrUpdate() error = %v, wantErr %v", err, o.wantErr)
-			return
+		if o.wantErr {
+			assert.Error(t, err)
+		} else {
+			assert.NoError(t, err)
 		}
 	}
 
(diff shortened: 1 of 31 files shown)


internal/controller/operator/factory/vlsingle/vlsingle_test.go
```diff
@@ -195,19 +196,20 @@ func TestCreateOrUpdateVLSingleService(t *testing.T) {
 		fclient := k8stools.GetTestClientWithObjects(o.predefinedObjects)
 		ctx := context.TODO()
 		err := createOrUpdateService(ctx, fclient, o.cr, nil)
-		if (err != nil) != o.wantErr {
-			t.Errorf("CreateOrUpdateService() error = %v, wantErr %v", err, o.wantErr)
-			return
+		if o.wantErr {
+			assert.Error(t, err)
```
AndrewChubatiuk, line 200: why not just return after error check instead of placing a whole successful path into `else`?

vrutkovs: Yeah, this is even worse.

I was attempting to find the middle ground between
```
		if o.wantErr {
			assert.Error(t, err)
		} else {
			assert.NoError(t, err)
		}
```
and 
```
		if o.wantErr {
			assert.Error(t, err)
			return
		}
		assert.NoError(t, err)
		svc := build.Service(o.cr, o.cr.Spec.Port, nil)
```

PR #1889, opened 2026-02-26
replace converter with a controller for each type, that should be converted 

AndrewChubatiuk:
after controller-runtime upgrade without KUBE_FEATURE_WatchListClient env variable set to false prometheus conversion is not working (at least with kind 0.31.0).

replaced prometheus converter with controllers for each type
before resource watchers were added once respective prom CRD appears now controller checks presence of CRD on startup only and skips controller start if CRD is absent

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,7 +14,7 @@ aliases:
 ## tip
 
 **Update note 1**: deprecated env variables for Prometheus CRs conversion `VM_ENABLEDPROMETHEUSCONVERTER_PODMONITOR`, `VM_ENABLEDPROMETHEUSCONVERTER_SERVICESCRAPE`, `VM_ENABLEDPROMETHEUSCONVERTER_PROMETHEUSRULE`, `VM_ENABLEDPROMETHEUSCONVERTER_PROBE`, `VM_ENABLEDPROMETHEUSCONVERTER_SCRAPECONFIG`. Use `-controller.disableReconcileFor` command-line flag with comma-separated list of controller names, that should be disabled.
-**Update note 2**: removed `operator_prometheus_converter_watch_events_total` metric since migration of Prometheus object watchers to controllers made this counter useless
+**Update note 2**: removed `operator_prometheus_converter_watch_events_total` metric since migration of Prometheus object watchers to controllers made this counter obsolete.
 **Update note 3**: made `controller.prometheusCRD.resyncPeriod` command line flag noop, which was relevant to Prometheus object watchers.
 
 * Dependency: [vmoperator](https://docs.victoriametrics.com/operator/): Updated default versions for VM apps to [v1.138.0](https://github.com/VictoriaMetrics/VictoriaMetrics/releases/tag/v1.138.0) version


AndrewChubatiuk:
@cubic-dev-ai

AndrewChubatiuk:
@cubic-dev-ai

internal/config/config.go
```diff
@@ -524,7 +524,7 @@ type BaseOperatorConf struct {
 
 	EnabledPrometheusConverter struct {
 		PodMonitor         bool `default:"true" env:"PODMONITOR"`
-		ServiceScrape      bool `default:"true" env:"SERVICESCRAPE"`
+		ServiceMonitor     bool `default:"true" env:"SERVICESCRAPE"`
```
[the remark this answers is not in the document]
AndrewChubatiuk: cannot rename env variable without breaking change, for now made internal variable in sync with prom CRD name

AndrewChubatiuk:
@cubic-dev-ai

internal/controller/operator/factory/k8stools/client_utils.go
```diff
@@ -133,108 +129,3 @@ func ListObjectsByNamespace[T any, PT listing[T]](ctx context.Context, rclient c
 	}
 	return nil
 }
-
-var (
-	activeWatchers         = prometheus.NewGaugeVec(prometheus.GaugeOpts{Name: "operator_prometheus_converter_active_watchers"}, []string{"namespace"})
```
vrutkovs, line 138: Do we provide an alternative to these metrics?

internal/controller/operator/factory/reconcile/reconcile.go
```diff
@@ -37,10 +37,21 @@ func InitDeadlines(intervalCheck, appWaitDeadline, podReadyDeadline, statusInter
...
-// mergeMaps performs 3-way merge for labels and annotations
-// it deletes only labels and annotations managed by operator CRDs
-// 3-rd party kubernetes annotations and labels must be preserved
-func mergeMaps(existingMap, newMap, prevMap map[string]string) map[string]string {
+// mergeMaps performs maps merge depending on a given strategy:
+// * prefer-vm: returns existing map
+// * prefer-prom: returns new map
+// * merge-vm-priority: performs 3-way merge by merging keys that absent in prev map from new map into existing map
+// * merge-prom-priority: performs 3-way merge by merging keys that absent in prev map from existing map into new map
+func mergeMaps(existingMap, newMap, prevMap map[string]string, strategy vmv1beta1.MetadataStrategy) map[string]string {
```
vrutkovs: We can simply have a new function for that, no need to change existing signature, so that we would not carry `strategy := vmv1beta1.MetadataStrategyMergePromPriority` every time

internal/controller/operator/factory/reconcile/reconcile.go
```diff
@@ -71,24 +82,36 @@ func mergeMaps(existingMap, newMap, prevMap map[string]string) map[string]string
 	return dst
 }
 
-func mergeMeta(existingObj, newObj client.Object, prevMeta *metav1.ObjectMeta, owner *metav1.OwnerReference, shouldRemoveFinalizer bool) (bool, error) {
+func mergeMeta(existingObj, newObj client.Object, prevMeta *metav1.ObjectMeta, owner *metav1.OwnerReference, shouldRemoveFinalizer, isConversion bool) (bool, error) {
```
vrutkovs: Also sounds like we'd rather have `mergeMetaWithConvertion` reusing parts of `mergeMeta`. We're gonna use convertion is one codepart only - no need to carry this flag every time

internal/manager/manager.go
```diff
@@ -110,7 +111,7 @@ var (
 	defaultKubernetesMajorVersion = managerFlags.Uint64("default.kubernetesVersion.major", 1, "Major version of kubernetes server, if operator cannot parse actual kubernetes response")
 	printDefaults                 = managerFlags.Bool("printDefaults", false, "print all variables with their default values and exit")
 	printFormat                   = managerFlags.String("printFormat", "table", "output format for --printDefaults. Can be table, json, yaml or list")
-	promCRDResyncPeriod           = managerFlags.Duration("controller.prometheusCRD.resyncPeriod", 0, "Configures resync period for prometheus CRD converter. Disabled by default")
+	_                             = managerFlags.Duration("controller.prometheusCRD.resyncPeriod", 0, "Configures resync period for prometheus CRD converter. Disabled by default")
```
vrutkovs, line 114: That looks like a breaking change?

AndrewChubatiuk: not breaking, as we're keeping this flag, but it becomes noop

internal/controller/operator/factory/k8stools/client_utils.go
AndrewChubatiuk: yes, moved it here internal/controller/operator/controllers.go

internal/controller/operator/factory/reconcile/reconcile.go
AndrewChubatiuk: updated

internal/controller/operator/factory/reconcile/reconcile.go
AndrewChubatiuk: done

ringerc:
I hit the same issue but prepared a much simpler fix for the failure to convert the prometheus objects, as shown in https://github.com/VictoriaMetrics/operator/pull/1999

I'm testing it out locally at the moment.

What other benefits does this larger change offer? I have no real preference so long as the issue gets fixed, I'm just curious if there are other benefits to per-resource controllers.

ringerc:
Will test to see if this also addresses https://github.com/VictoriaMetrics/operator/issues/1998

internal/controller/operator/controllers.go
```diff
@@ -63,11 +63,16 @@ var (
 		Name: "operator_controller_reconcile_errors_total",
 		Help: "Counts number context.Canceled errors",
 	})
+	activeConverterWatchers = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "operator_prometheus_converter_active_watchers",
```
vrutkovs, line 67: Is this metrics new? We should cover it in the docs

internal/controller/operator/promalertmanagerconfig_controller.go
```diff
@@ -0,0 +1,101 @@
...
+	activeConverterWatchers.WithLabelValues("alertmanagerconfig").Add(1)
+}
+
+// Scheme implements interface.
+func (r *PromAlertmanagerConfigReconciler) Scheme() *runtime.Scheme {
+	return r.OriginScheme
+}
+
+// Reconcile general reconcile method for controller
+// +kubebuilder:rbac:groups=monitoring.coreos.com,resources=alertmanagerconfigs,verbs=get;list;watch;create;update;patch;delete
```
vrutkovs:
```suggestion
// +kubebuilder:rbac:groups=monitoring.coreos.com,resources=alertmanagerconfigs,verbs=get;list;watch
```
I don't think we need permissions for mutation?

internal/controller/operator/promalertmanagerconfig_controller.go
```diff
@@ -0,0 +1,101 @@
...
+func (r *PromAlertmanagerConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&promv1alpha1.AlertmanagerConfig{}).
+		WithOptions(getDefaultOptions()).
+		Complete(r)
+}
+
+// IsDisabled returns true if controller should be disabled
+func (*PromAlertmanagerConfigReconciler) IsDisabled(cfg *config.BaseOperatorConf, disabledControllers sets.Set[string]) bool {
+	return disabledControllers.HasAny("VMAlertmanager", "VMAlertmanagerconfig") || !cfg.EnabledPrometheusConverter.AlertmanagerConfig
```
vrutkovs:
```suggestion
	return disabledControllers.HasAny("VMAlertmanager", "VMAlertmanagerConfig") || !cfg.EnabledPrometheusConverter.AlertmanagerConfig
```

internal/controller/operator/prompodmonitor_controller.go
```diff
@@ -0,0 +1,96 @@
...
+func (r *PromPodMonitorReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&promv1.PodMonitor{}).
+		WithOptions(getDefaultOptions()).
+		Complete(r)
+}
+
+// IsDisabled returns true if controller should be disabled
+func (*PromPodMonitorReconciler) IsDisabled(cfg *config.BaseOperatorConf, disabledControllers sets.Set[string]) bool {
+	return disabledControllers.HasAll("VMAgent", "VMSingle") || disabledControllers.Has("VMPodScrape") || !cfg.EnabledPrometheusConverter.PodMonitor
```
vrutkovs, line 95: Do we want to require both VMAgent and VMSingle contollers to be enabled here? I think we want at least one

internal/manager/manager.go
vrutkovs: We should add this to "update note 1" too

internal/controller/operator/controllers.go
AndrewChubatiuk: nope, it was moved from internal/controller/operator/factory/k8stools/client_utils.go

internal/controller/operator/prompodmonitor_controller.go
AndrewChubatiuk: this logic means that at least one VMAgent or VMSingle controller should be enabled: if both controllers are disable -> podMonitor should be disabled

AndrewChubatiuk:
thanks for review, just addressed your comments

internal/controller/operator/prompodmonitor_controller.go
vrutkovs: ooh, gotcha

PR #1897, opened 2026-02-27
feat: allow creating vmdistributed without vmauth set

### LINKED ISSUE #1826: VMDistributed: allow configurations without vmauth
Some users may want to use VMDistributed as a place to declaratively declare multiple VMCluster / VMAgents without VMAuth. We should allow VMAuth to be optional (at the expense of possible partial results on read path)

vrutkovs:
Users should be able to provide their own LB instead of VMAuth (taking care of proper switching themselves), so VMAuth should be optional for VMDistributed

Fixes #1826

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1211,6 +1211,7 @@ Appears in: [VMDistributedSpec](#vmdistributedspec)
 
 | Field | Description |
 | --- | --- |
+| enabled<a href="#vmdistributedauth-enabled" id="vmdistributedauth-enabled">#</a><br/>_boolean_ | _(Optional)_<br/>Enabled defines if vmauth should be created. |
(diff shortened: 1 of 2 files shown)


api/operator/v1alpha1/vmdistributed_types.go
```diff
@@ -42,7 +42,8 @@ type VMDistributedSpec struct {
 	ParsingError string `json:"-" yaml:"-"`
 	// VMAuth is a VMAuth definition (name + optional spec) that acts as a proxy for the VMUsers created by the operator.
 	// Use an inline spec to define a VMAuth object in-place or provide a name to reference an existing VMAuth.
-	VMAuth VMDistributedAuth `json:"vmauth,omitempty"`
+	// +optional
+	VMAuth *VMDistributedAuth `json:"vmauth,omitempty"`
```
AndrewChubatiuk: let's add `Disabled` toggle into VMDistributedAuth struct instead. in a version in master omitted VMAuth section means it should get it's name from CR Name

PR #1911, opened 2026-03-02
vlcluster,vtcluster: do not ignore extraStorageNode, when default storage is not enabled

### LINKED ISSUE #1910: VLCluster: `extraStorageNodes` silently ignored without in-cluster `vlstorage`
In [`vlselect.go#L206`](https://github.com/VictoriaMetrics/operator/blob/master/internal/controller/operator/factory/vlcluster/vlselect.go#L206), the `extraStorageNodes` block is nested inside the `VLStorage != nil` guard, so it is unreachable when no in-cluster `vlstorage` is defined. No `-storageNode` args are emitted.

The same issue exists in [`vtcluster/select.go#L206`](https://github.com/VictoriaMetrics/operator/blob/master/internal/controller/operator/factory/vtcluster/select.go#L206).

Existing tests always pair `extraStorageNodes` with a `vlstorage` spec, so this path has no coverage.

The use case is a standalone `VLSelect` for use in a multi-tier (global query) architecture where no `VLInsert` or `VLStorage` are necessary.

### Example

```yaml
apiVersion: operator.victoriametrics.com/v1
kind: VLCluster
metadata:
  name: vlselect-global
  namespace: victoria-logs
spec:
  vlselect:
    extraStorageNodes:
      - addr: vlselect-regional.other-cluster.example.com:9471
  # no vlstorage section
```

**Expected:** `-storageNode=vlselect-regional...` in the `vlselect` Deployment args.
**Actual:** no `-storageNode` arg is present, pod starts as a standard storage node, failing due to no PV for storage.

Edit: I realize the workaround is to use `extraArgs.storageNode`, but my _expectation_ is that `extraStorageNodes` would deliberately set this. At least warn on startup or in the operator logs rather than fail silently.

AndrewChubatiuk:
fixes #1910

---
## Summary by cubic
Fixes VLSelect and VTSelect ignoring ExtraStorageNodes when default storage is disabled, so select pods connect to the provided storage nodes. Also validates and rejects duplicate storageNode addresses across defaults and extras (fixes #1910).

- **Bug Fixes**
  - Build `-storageNode` flag and collect node IDs outside the default-storage check.
  - Append `-storageNode` only when nodes exist; always include `ExtraStorageNodes` even without default storage.
  - Validate uniqueness of storageNode addresses from default storage (when enabled), `ExtraArgs["storageNode"]`, and `ExtraStorageNodes`.

<sup>Written for commit de3e677f96c93c1be4493d75072252392a6d4568. Summary will update on new commits.</sup>

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -36,6 +36,7 @@ aliases:
 * BUGFIX: [vmoperator](https://docs.victoriametrics.com/operator/): wait for STS deletion in case of recreation without throwing an error.
 * BUGFIX: [vmdistributed](https://docs.victoriametrics.com/operator/resources/vmdistributed/): ignore VMAuth update/delete operations if controller is disabled.
 * BUGFIX: [vmalertmanager](https://docs.victoriametrics.com/operator/resources/vmalertmanager/): fixed ignored tracing config, when no alertmanagerconfig CRs collected. See [#1983](https://github.com/VictoriaMetrics/operator/issues/1983).
+* BUGFIX: [vmagent](https://docs.victoriametrics.com/operator/resources/vmagent/): apply scrape class relabellings before job ones. See [#1997](https://github.com/VictoriaMetrics/operator/issues/1997).
 * BUGFIX: [vlcluster](https://docs.victoriametrics.com/operator/resources/vlcluster/) and [vtcluster](https://docs.victoriametrics.com/operator/resources/vtcluster/): do not ignore ExtraStorageNodes for select, when default storage is disabled. See [#1910](https://github.com/VictoriaMetrics/operator/issues/1910).
 
 ## [v0.68.1](https://github.com/VictoriaMetrics/operator/releases/tag/v0.68.1)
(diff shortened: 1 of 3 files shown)


internal/controller/operator/factory/vlcluster/vlselect.go
```diff
@@ -233,20 +233,22 @@ func buildVLSelectPodSpec(cr *vmv1.VLCluster) (*corev1.PodTemplateSpec, error) {
...
 		}
-		if len(cr.Spec.VLSelect.ExtraStorageNodes) > 0 {
-			for i, node := range cr.Spec.VLSelect.ExtraStorageNodes {
-				idx := i + len(storageNodeIds)
-				storageNodeFlag.Add(node.Addr, idx)
-			}
+	}
+	if len(cr.Spec.VLSelect.ExtraStorageNodes) > 0 {
+		for i, node := range cr.Spec.VLSelect.ExtraStorageNodes {
+			idx := i + len(storageNodeIds)
```
vrutkovs, line 253: Perhaps we should deduplicate the list? Users may accidentally set the same node twice.

Also, needs tests

AndrewChubatiuk: throwing error in case of duplicates instead since also per storage auth can be added and deduplication can break an order

PR #1916, opened 2026-03-03
chore: refactored probe building and cluster defaults management

AndrewChubatiuk:
- temporary replaced caarlos0/env with fork till [PR](https://github.com/caarlos0/env/pull/412) is merged
- consolidate all functions for cluster defaults population
- merged EmbeddedProbes, CommonDefaultableParams and CommonApplicationDeploymentParams into CommonAppsParams as all these structs are used together in all CRs
- moved setting terminationGracePeriodSeconds to defaults management

---
## Summary by cubic
Refactored probe handling and app/cluster defaults to simplify CRDs and centralize config. Probes are now built from CommonAppsParams, support proxy protocol via extraArgs, and termination grace is controlled by operator defaults per app.

- **Refactors**
  - Replaced EmbeddedProbes, CommonDefaultableParams, and CommonApplicationDeploymentParams with CommonAppsParams across CRDs; removed Probe() from CRs. Probe builder now mutates containers and consumes CommonAppsParams.
  - Added UseProxyProtocol(extraArgs) helper and UseProxyProtocol() on CRs to drive probe behavior.
  - Moved terminationGracePeriodSeconds out of CRDs; defaults come from VM_<APP>_TERMINATION_GRACE_PERIOD_SECONDS (default 30). Updated docs, CRDs, and tests.
  - Security helpers and Add*CommonParams now read UseStrictSecurity/SecurityContext from CommonAppsParams; AddServiceAccountTokenVolume updated accordingly.
  - Consolidated defaults application and license handling; tests initialize defaults via AddDefaults where needed.
  - Temporarily replaced caarlos0/env with a fork to unblock builds.

- **Migration**
  - Update manifests to use CommonAppsParams and drop EmbeddedProbes.
  - Move liveness/readiness/startup probe settings under CommonAppsParams.
  - If terminationGracePeriodSeconds was set in specs, switch to VM_<APP>_TERMINATION_GRACE_PERIOD_SECONDS.

<sup>Written for commit 3de2008c6df6393d5b6a1b0f5162114da322a75b. Summary will update on new commits.</sup>


api/operator/v1beta1/vmextra_types.go
```diff
@@ -130,6 +131,17 @@ func ClusterPrefixedName(kind ClusterComponent, name, prefix string, internal bo
 	return fmt.Sprintf("%s%s%s-%s", prefix, string(kind), suffix, name)
 }
 
+// UseProxyProtocol implements build.probeCRD interface
```
vrutkovs:
```suggestion
// UseProxyProtocol is a helper for build.probeCRD interface implementations
```
I don't think this is correct - `UseProxyProtocol() bool` is in the interface

api/operator/v1beta1/vmextra_types.go
```diff
@@ -130,6 +131,17 @@ func ClusterPrefixedName(kind ClusterComponent, name, prefix string, internal bo
...
+func UseProxyProtocol(extraArgs map[string]string) bool {
+	if v, ok := extraArgs[httpUseProxyProtocol]; ok {
```
vrutkovs: Why are we doing this? This is supposedly a simple `bool`, isn't it?

Lets add some tests for this if its justified

internal/config/config.go
```diff
@@ -767,8 +799,14 @@ func (labels *Labels) Set(value string) error {
 // ConfigAsMetrics exposes major configuration params as prometheus metrics
 func ConfigAsMetrics(r metrics.RegistererGatherer, cfg *BaseOperatorConf) {
 	opts := getEnvOpts()
-	mapper := func(v string) string {
-		return opts.Environment[v]
+	rawEnvVars := make(map[string]string)
```
vrutkovs, line 802: This is a dangerous function - we'll be expanding all env vars, not limited to the env var we want.

Lets move this expansion to `UseProxyProtocol`?

internal/controller/operator/factory/build/container.go
```diff
@@ -19,95 +19,71 @@ const probeTimeoutSeconds int32 = 5
...
+		if probe.PeriodSeconds == 0 {
+			probe.PeriodSeconds = 5
+		}
+		if probe.FailureThreshold == 0 {
+			probe.FailureThreshold = 10
+		}
+		if probe.TimeoutSeconds == 0 {
+			probe.TimeoutSeconds = probeTimeoutSeconds
+		}
+		if probe.SuccessThreshold == 0 {
```
vrutkovs, line 73: Would the user still be able to disable this - by unsetting or setting to 0?

vrutkovs (changes requested):

internal/config/config.go
AndrewChubatiuk: it's a subset of functionality from carlos0/env library, which we're using already and this function is only used once to init metrics with env variables names and values

this expansion has nothing to do with useProxyProtocol

internal/controller/operator/factory/build/container.go
AndrewChubatiuk: it's not possible, logic is the same in master

api/operator/v1beta1/vmextra_types.go
AndrewChubatiuk: use proxy protocol can contain multiple comma-separated bool values, this function extracts first one and checks if it's true

internal/config/config.go
vrutkovs: Oh, I see - its used in declaration like `VM_OPERATOR_VERSION,expand`. Why would we want to change anything here - is it for recursive env var expansion? In that case its best to limit the depth to some sensible value - 3 or maybe even 5

api/operator/v1beta1/vmextra_types.go
vrutkovs: Oh, this is a setting for each `httpListenAddr`, which may be specified multiple times, gotcha

internal/config/config.go
AndrewChubatiuk: > is it for recursive env var expansion?

yes

> its best to limit the depth to some sensible value - 3 or maybe even 5

agree, will update soon

vrutkovs (approved):

rrhubenov:
Hi, greatly appreciate the work you're putting in!
Just FYI it seems this PR included some breaking changes that weren't documented in the release notes (or at least I could not find them). An example one that we hit in [Gardener](https://github.com/gardener/gardener) was the removal/renaming of the `CommonDefaultableParams` struct to `CommonAppsParams`.

AndrewChubatiuk:
hey @rrhubenov 
we do not track API changes in changelog at the moment. @vrutkovs should we should start doing it?

vrutkovs:
Sure, I think this would be important to track. I'll update the release guide - https://github.com/VictoriaMetrics/operator/pull/2400

PR #1931, opened 2026-03-04
test: rework PVC expand test cases with STS

vrutkovs:
* reuse objects where possible
* add test cases around PVC expansion


AndrewChubatiuk:
> reuse objects where possible

what value are we getting from this?
it just makes tests less readable

vrutkovs:
My intention to make it more readable - explicitly state which part of the object has changed, retaining the same ObjectMeta. If you insist I can revert that part

AndrewChubatiuk:
imo it's harder to read content between comment and start of function
i liked your approach with preRun function in other tests, could something similar be applied here as well?

vrutkovs:
yeah, okay, lets do "base PVC and preRun function to adjust it"

Issue #1840, opened 2026-02-23
vmdistributed: support authenticated vmauths


vrutkovs:
Allow switching to authenticated VMAuth by moving the configuration from UnauthorizedAccessConfig

AndrewChubatiuk:
looks like it depends on https://github.com/VictoriaMetrics/VictoriaMetrics/issues/9730
VMDistributed VMAuth will have predefined `read` and `write` backends, which then can be referenced by VMUser's `spec.backend` (or `spec.backends`) field

AndrewChubatiuk:
feature will be available in upcoming 0.69.0 release

AndrewChubatiuk:
VMDistributed support configuring authentication starting operator release 0.69.0. Please check [docs](https://docs.victoriametrics.com/operator/resources/vmdistributed/index.html#authorization) for more information

PR #1903, opened 2026-03-02
vmdistributed: merge all vmclusters into a single read and all vmagents into a single write backend in vmauth config

AndrewChubatiuk:
## Summary by cubic
Consolidates VMDistributed VMAuth by merging all VMCluster/vmselect targets into one read backend and all VMAgents into one write backend. Adds VMUser targetRefs[*].crds to route to multiple CRDs; deprecates targetRefs[*].crd.

- **New Features**
  - VMUser: add targetRefs[*].crds for multiple CRD targets; crd is deprecated. API schema/docs/deepcopy, URL resolution, and validation updated (exactly one of crd/crds/static is required).

- **Bug Fixes**
  - VMDistributed: VMAuth now emits two targetRefs (write=VMAgents, read=VMCluster/vmselect) with aggregated CRDs per backend, reducing config size and stabilizing routing; tests updated.

<sup>Written for commit 73d85c65cf8c9f6c24621f1b704663c820357a83. Summary will update on new commits.</sup>

PR #1936, opened 2026-03-05
vmuser: support multiple objects of same kind instead of multiple kinds for a single backend

AndrewChubatiuk:
[this PR](https://github.com/VictoriaMetrics/operator/pull/1903) introduces `targetRefs[*].crds`, which was basically needed for vmdistributed, now proposing to add `targetsRefs[*].crd.objects` instead. multiple objects of a single kind most likely have same path prefixes and endpoints, which is important to get predictable response while querying given backend.

additionally given below changes fix issue https://github.com/VictoriaMetrics/operator/issues/1840:
 - added vmauth spec.defaultTargetRefs that can be referenced in VMUser spec
 - disabled unauthorized access in vmdistributed by default

---
## Summary by cubic
VMUser now supports multiple same‑kind targets per backend via `targetRefs[*].crd.objects`. VMAuth adds `spec.defaultTargetRefs` so VMUser and unauthorized access can reference shared `read`/`write` backends by name; VMDistributed predefines these and disables unauthorized access by default.

- **Migration**
  - Replace `targetRefs[*].crds` with `targetRefs[*].crd.objects`; keep `crd.kind` and list each object as `{name, namespace}`.
  - For a single target, `crd.name` and `crd.namespace` still work.
  - Use `targetRefs[*].name` to reference `VMAuth.spec.defaultTargetRefs`; inline fields override defaults. `UnauthorizedUserAccessSpec.targetRefs` also supports names.
  - Validation still requires exactly one of `crd` or `static` per TargetRef; CRD schema, examples, docs, and controller/tests are updated.

<sup>Written for commit 047667bb19542afaead4d4fc620075d824ed5f5d. Summary will update on new commits.</sup>


api/operator/v1beta1/vmuser_types.go
```diff
@@ -55,17 +54,16 @@ type VMUserSpec struct {
...
+	// +optional
+	Name string `json:"name,omitempty"`
 	// CRD describes exist operator's CRD object,
 	// operator generates access url based on CRD params.
 	// +optional
 	CRD *CRDRef `json:"crd,omitempty"`
-	// CRD describes existing operator's CRD objects,
-	// operator generates access url based on CRD params.
-	// +optional
-	CRDs []CRDRef `json:"crds,omitempty"`
```
vrutkovs, line 68: I don't think this ever made it to a released version, right?

config/examples/vmdistributed-migrate-from-chart.yaml
```diff
@@ -6,6 +6,10 @@ spec:
   vmauth:
     spec:
       replicaCount: 1
+      unauthorizedUserAccessSpec:
+        targetRefs:
+          - name: read
```
vrutkovs, line 11: Are these hardcoded names in VMDistributed? We should add those to the docs

internal/controller/operator/factory/build/build.go
```diff
@@ -322,7 +322,7 @@ func mergeMapsRecursive(baseMap, overrideMap map[string]any) {
 
 // MergeDeep merges an override object into a base one.
 // Fields present in the override will overwrite corresponding fields in the base.
-func MergeDeep[T comparable](base, override T) error {
+func MergeDeep[T comparable](base, override T, reverse bool) error {
```
vrutkovs, line 325: Why not swap it on call time?

internal/controller/operator/factory/vmauth/vmusers_config.go
```diff
@@ -36,20 +36,97 @@ type parsedObjects struct {
...
 func (pos *parsedObjects) buildConfig(ctx context.Context, rclient client.Client, cr *vmv1beta1.VMAuth, ac *build.AssetsCache) ([]byte, error) {
 	// loads info about exist operator object kind for crdRef.
-	crdURLCache := make(map[string]string)
+	objURLs := make(map[string]string)
+	backends := make(map[string]*vmv1beta1.TargetRef)
+	updateObjURLs := func(ref *vmv1beta1.TargetRef, isDefault bool) error {
+		if ref == nil {
+			return nil
+		}
+		if isDefault {
```
vrutkovs: Check that `len(ref.Name) > 0` first

internal/controller/operator/factory/vmauth/vmusers_config.go
```diff
@@ -36,20 +36,97 @@ type parsedObjects struct {
...
+		} else if len(ref.Name) > 0 {
+			if r, ok := backends[ref.Name]; ok {
+				if err := build.MergeDeep(ref, r, true); err != nil {
+					return fmt.Errorf("failed to merge target refs: %w", err)
+				}
+			} else {
+				return fmt.Errorf("failed to find defaultTargetRef=%s", ref.Name)
+			}
+		}
+		if ref.CRD == nil {
```
vrutkovs: Lets break CRD check in a separate function - it would be easier to unit test

api/operator/v1beta1/vmuser_types.go
AndrewChubatiuk: it was in 0.69 rc

config/examples/vmdistributed-migrate-from-chart.yaml
AndrewChubatiuk: right, added to VMAuth `spec.defaultTargetRefs` property, which allows to add predefined backends, that can be referenced in unanauthorized user section and VMUser as well. for VMDistributed this list has two items with `read` and `write` names

let's mentioned it in docs.

this change is kind a breaking, as it requires existing VMDistributed users to update CR manuallly, but suppose it's fine since CRD is in alpha

internal/controller/operator/factory/build/build.go
AndrewChubatiuk: before in merge result was stored to base and in case of repeated parameters override had higher priority, with this change still want to store merge result to base, but change priorities

config/examples/vmdistributed-migrate-from-chart.yaml
AndrewChubatiuk: added authorization section to VMDistributed docs and section about predefined backends to VMAuth

internal/controller/operator/factory/vmauth/vmusers_config.go
AndrewChubatiuk: updated

internal/controller/operator/factory/vmauth/vmusers_config.go
AndrewChubatiuk: done

internal/controller/operator/factory/build/build.go
vrutkovs: ah, yes, we mutate the vars here.

PR #1945, opened 2026-03-09
test: add upgrade e2e tests

### LINKED ISSUE #1952: Introduce upgrade tests
Add a new test suite to catch unexpected pod rollouts:
* Install an old version of the operator (0.65.0)
* Add a resource (VMAgent)
* Take down the operator, run the latest version
* Ensure that generated resources (deployment) didn't get spec.template.spec updated, which would result into a pod rollout.

If we're introducing a change which would cause rollouts (e.g. config-reloader version bump) we should provide an object setting to avoid it (e.g. pin VM and config-reloader version)

vrutkovs:
Ensure that operator upgrade doesn't cause unnecessary rollouts. If there is a rollout expected (i.e. reloader change) the test case should apply changes to the initial resource to avoid a rollout.

Fixes #1952

TODO:
* [x] Make sure we cover all resources
* [x] Sort containers by name
* [x] Set custom config-reloader version to avoid false positives
* [x] 0.65 -> 0.66 config-reloader changes. Any way to overcome or just ignore?
* [x] VM_LOOPBACK
* [x] Remove DONOTMERGE commit

Deferred:
* [ ] Use Cloud manifests
* [ ] Find other breaks in the changelog

### DIFF
--- a/test/e2e/upgrade/upgrade_test.go
+++ b/test/e2e/upgrade/upgrade_test.go
@@ -1002,9 +1002,6 @@ var _ = Describe("operator upgrade", Label("upgrade"), func() {
 				RequestsLoadBalancer: vmv1beta1.VMAuthLoadBalancer{
 					Enabled: true,
 					Spec: vmv1beta1.VMAuthLoadBalancerSpec{
-						CommonConfigReloaderParams: vmv1beta1.CommonConfigReloaderParams{
-							ConfigReloaderImage: "quay.io/victoriametrics/operator:config-reloader-v0.65.0",
-						},
 						CommonAppsParams: vmv1beta1.CommonAppsParams{
 							ReplicaCount: ptr.To[int32](1),
 							Image: vmv1beta1.Image{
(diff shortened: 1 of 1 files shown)


AndrewChubatiuk:
while fixing [this issue](https://github.com/VictoriaMetrics/operator/issues/1740), that was done for 0.67.0 release, order of VMAgent containers was changed, this impacts your work on this PR

vrutkovs:
Added a todo to sort containers by name (I don't think the order change would cause a rollout)

api/operator/v1beta1/vmcluster_types.go
```diff
@@ -1005,8 +1005,9 @@ type VMAuthLoadBalancerSpec struct {
 	RollingUpdate *appsv1.RollingUpdateDeployment `json:"rollingUpdate,omitempty"`
 	// License configures enterprise features license key
 	// +optional
-	License          *License `json:"license,omitempty"`
-	CommonAppsParams `json:",inline"`
+	License                    *License `json:"license,omitempty"`
+	CommonConfigReloaderParams `json:",inline,omitempty"`
```
AndrewChubatiuk: why was it added? config reloader is not needed for vmcluster

vrutkovs: We need that to avoid container spec changes from previous installs. Its noop for new VMClusters

AndrewChubatiuk: How it impacts container specs if it's noop?

vrutkovs: Its noop in current version, but if I understood correctly, users could have specified it previously (probably was noop back then too?) and now if we remove this option, it would result into a pod rollout.

Not sure how widespread this case was however - perhaps its easier to just drop the test cases for these?

AndrewChubatiuk: config-reloader was noop even before, that's why it was removed. only reason to keep it if for some reason people have it set

vrutkovs: Yeah, okay, removing the unnecessary config-reloader pin

PR #1946, opened 2026-03-10
ci: do not trigger release on update

AndrewChubatiuk:
currently CI is triggered each time release is updated

---
## Summary by cubic
Update GitHub Actions release workflow to trigger only on the 'released' event. This prevents duplicate or noisy runs on release creation or prerelease updates.

<sup>Written for commit 5f6db23185f6d34bf279f7aafc87e4abe81fef12. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,8 +2,7 @@ name: Release
 on:
   release:
     types:
-      - created
-      - prereleased
+      - released
 jobs:
   release:
     name: Release on GitHub


.github/workflows/release.yaml
```diff
@@ -2,8 +2,7 @@ name: Release
 on:
   release:
     types:
-      - created
-      - prereleased
```
vrutkovs, line 6: We do want this to be running RCs - this is the only way for customers to consume those, as it builds packages in official registries and prepares CRD.yaml

AndrewChubatiuk: afaik released event includes prereleases as well

vrutkovs: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release

>The prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.

Issue #1786, opened 2026-02-05
`log-format` config reloader extra arg is no longer supporter


vrutkovs:
Existing VMAgent CRs were migrated to use config-reloader in 0.67.0, but configuration options were not:
```yaml
configReloaderExtraArgs:
    log-format: json
```
would make the vmagent crashloop, although previously the configuration was valid.

We should add this particular option as an alias to `loggerFormat`. Additionally, we should inspect previous options and attempt to avoid configuration migration if possible, otherwise we should explicitly list which options are no longer supported.

AndrewChubatiuk:
added config reloader arguments to [docs page](https://docs.victoriametrics.com/operator/configuration/#config-reloader-flags) and mentioned need to check config reloader command line arguments in [0.67.0 update notes](https://docs.victoriametrics.com/operator/changelog/#v0670)

vrutkovs:
Reopening - the alias to support existing configs was not added

AndrewChubatiuk:
which aliases besides this one are expected to be added?

vrutkovs:
I'm afraid there's too many to add every single one - and, most importantly, we don't want to encourage the usage of the old ones. Lets handle it on case-by-case basis, like in the original description

AndrewChubatiuk:
fixed in 0.68.3

PR #1954, opened 2026-03-12
feat: add log-format as an alias to loggerFormat

vrutkovs:
Some installations with outdated config-reloader may be upgraded to latest versions of the operator. In order to keep the configuration compatible with previous configurations this introduces a compat flag for logger format.

Fixes https://github.com/VictoriaMetrics/operator/issues/1786


cmd/config-reloader/main.go
```diff
@@ -68,6 +68,8 @@ var (
 		"Optional TLS server name to use for connections to -reload-url.")
 	tlsInsecureSkipVerify = flag.Bool("reload.tlsInsecureSkipVerify", true,
 		"Whether to skip tls verification when connecting to -reload-url")
+	logFormat = flag.String("log-format", "",
```
AndrewChubatiuk, line 71:
```suggestion
	flag.StringVar(logFormat, "log-format", *logFormat,
```
what do you think about such solution?

AndrewChubatiuk:
is `log-format` the only flag we need to provide alias for?

cmd/config-reloader/main.go
vrutkovs: I don't understand what is it doing - `logFormat` is not declared anywhere

AndrewChubatiuk: my fault, forget it's derived from victoriametrics

PR #1963, opened 2026-03-13
vmanomaly/config: add support of retention configuration

zekker6:
See docs at: https://docs.victoriametrics.com/anomaly-detection/components/settings/#configuration


AndrewChubatiuk (approved):

internal/controller/operator/factory/vmanomaly/config/config_test.go
```diff
@@ -311,6 +311,84 @@ settings:
...
+				Namespace: "monitoring",
+			},
+			Spec: vmv1.VMAnomalySpec{
+				License: &vmv1beta1.License{
+					Key: ptr.To("test"),
+				},
+				ConfigRawYaml: `
+models:
+  model_zscore:
+    class: 'zscore'
```
fred-navruzov, line 328: note: zscore (offline) was deprecated, and now converts to `zscore_online` with warning in logs - https://docs.victoriametrics.com/anomaly-detection/components/models/#deprecated-models

PR #1964, opened 2026-03-13
controllers: use single error handling for all CRs

AndrewChubatiuk:
## Summary by cubic
Unifies error handling across all controllers and standardizes how status metadata is read from CRs. Adds graceful shutdown handling and better error propagation while keeping NotFound silent and requeuing on cancellations and conflicts.

- **Refactors**
  - Removed `handleReconcileErrWithoutStatus`; all controllers now use a single `handleReconcileErr`.
  - Moved `GetStatusMetadata` from Status structs to parent CRs; updated helpers (`waitForStatus`, status updates) to read metadata from the CR.
  - Added `GetStatus`/`DefaultStatusFields` for `VMAlertmanagerConfig`, `VMRule`, `VMUser`, and scrape CRs (`VMNodeScrape`, `VMPodScrape`, `VMProbe`, `VMScrapeConfig`, `VMServiceScrape`, `VMStaticScrape`); standardized controllers to use value instances and pass pointers.

- **Bug Fixes**
  - Requeue on `context.Canceled` unless caused by graceful shutdown (`operator.ErrShutdown`); keep NotFound silent and back off on conflicts.
  - Emit events only on real errors; consistent error metrics across all CRs.
  - Stop reconciliation on invalid specs via `spec.ParsingError`; `VMRule` and `VMUser` capture JSON parse errors during unmarshal.
  - Propagate errors from related-resource updates while continuing other items (VMAgent/VMSingle scrapes, alert/rule config updates).

<sup>Written for commit 551ba04be719c35b05e79ed6b5577e54826ac516. Summary will update on new commits.</sup>

PR #1965, opened 2026-03-13
Requeue zone update when context is cancelled

### LINKED ISSUE #1962: handleReconcileErr swallows context.Canceled without requeueing, permanently dropping CRs from work queue
### Describe the bug

When a controller's `Reconcile` function returns a `context.Canceled` error, `handleReconcileErr` in `internal/controller/operator/controllers.go` silently discards it and returns `(result, nil)` — which tells controller-runtime **not to requeue** the item. If the cancellation interrupted reconciliation before completion (e.g., finalizer not yet added, child resources not yet created), the CR is permanently dropped from the work queue and never reconciled again.

This affects all controllers that use `handleReconcileErr`, but we hit it specifically on the **VMCluster controller** in a VMDistributed deployment where az2's VMCluster CR was created but never reconciled — no finalizer, no `last-applied-spec` annotation, no child StatefulSets/Deployments, no status. The CR existed as a zombie indefinitely.

### Root cause

https://github.com/VictoriaMetrics/operator/blob/8db45b7b260d1db48036f18fb3f7af86efbb80b3/internal/controller/operator/controllers.go#L127-L129

```go
case errors.Is(err, context.Canceled):
    contextCancelErrorsTotal.Inc()
    return originResult, nil  // ← returns nil error = no requeue
```

Returning `nil` tells controller-runtime the reconciliation succeeded. The item is removed from the work queue. If the CR was only partially reconciled (e.g., created but finalizer not added), no future event will re-trigger reconciliation unless an external change (annotation, spec update) generates a new watch event.

### How to reproduce

1. Deploy a `VMDistributed` CR with 2+ zones on a busy operator instance (multiple CI namespaces sharing the same operator)
2. The VMDistributed controller creates zone VMCluster CRs sequentially
3. If the VMCluster controller's reconciliation of a newly created CR is interrupted by context cancellation (e.g., manager restart, leader election change, or high controller-runtime queue pressure), the CR is permanently orphaned
4. The VMDistributed controller enters an infinite loop: it tries to `waitForStatus` on the orphaned VMCluster, times out after `ReadyTimeout` (default 5m), requeues, and repeats forever

### Observed behavior

- VMCluster CR `az2` existed with `generation: 1` but:
  - No `apps.victoriametrics.com/finalizer`
  - No `operator.victoriametrics/last-applied-spec` annotation
  - No `.status` (empty)
  - No child workloads (no StatefulSets, Deployments, or Pods)
- VMCluster controller processed `az2` in other namespaces (staging, other CI branches) but never in the affected namespace
- VMDistributed controller cycled every ~6 minutes: reconcile az1 (operational) → wait for az2 status → 5min timeout → requeue
- **Workaround:** Adding an annotation to the VMCluster CR (`kubectl annotate vmcluster az2 operator.victoriametrics.com/trigger-reconcile=$(date -u +%s)`) generated a new watch event, re-enqueued the CR, and the VMCluster controller reconciled it successfully within seconds — all pods came up and status reached `operational`

### Expected behavior

A `context.Canceled` error during reconciliation should **requeue with backoff**, not silently drop the item. The CR should eventually be reconciled to completion.

### Suggested fix

```go
case errors.Is(err, context.Canceled):
    contextCancelErrorsTotal.Inc()
    return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```

Or alternatively, return the error to let the rate limiter handle requeueing:

```go
case errors.Is(err, context.Canceled):
    contextCancelErrorsTotal.Inc()
    return originResult, err  // let controller-runtime requeue with backoff
```

The same pattern applies to `handleReconcileErrWithoutStatus` at line ~198 which has the identical issue.

### Environment

- **Kubernetes:** EKS (AWS)
- **Trigger:** VMDistributed with 2 zones, operator shared across multiple CI namespaces
- **CRD affected:** VMCluster (but bug exists in shared `handleReconcileErr` used by all controllers)

### Additional context

The `handleReconcileErrWithoutStatus` function at L185-200 has the same bug — `context.Canceled` returns `nil` without requeueing. Both should be fixed.

This is particularly impactful for VMDistributed deployments because the VMDistributed controller creates child CRs (VMCluster, VMAgent) and then blocks in `waitForStatus` — if the child CR is dropped from the VMCluster controller's queue, the VMDistributed controller enters a permanent timeout loop with no self-healing path.

vrutkovs:
Attach a more detailed error every time we cancel the context. Requeue the request if the cancellation occurred during zone processing. This would prevent some zones from being left untouched, as otherwise the controller would restart from scratch.

Fixes #1962

### DIFF
--- a/internal/controller/operator/factory/vmdistributed/zone.go
+++ b/internal/controller/operator/factory/vmdistributed/zone.go
@@ -2,7 +2,6 @@ package vmdistributed
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -195,8 +194,9 @@ func (zs *zones) upgrade(ctx context.Context, rclient client.Client, cr *vmv1alp
 	}
 	if needsLBUpdate {
 		// wait for empty persistent queue
-		if err := zs.waitForEmptyPQ(ctx, rclient, defaultMetricsCheckInterval, i); err != nil {
-			return fmt.Errorf("zone=%s: failed to wait till VMCluster=%s queue is empty: %w", item, nsnCluster.String(), err)
+		zs.waitForEmptyPQ(ctx, rclient, defaultMetricsCheckInterval, i)
+		if ctx.Err() != nil {
+			return fmt.Errorf("zone=%s: failed to wait till VMCluster=%s queue is empty", item, nsnCluster.String())
 		}
 
 		// excluding zone from VMAuth LB
@@ -217,8 +217,9 @@ func (zs *zones) upgrade(ctx context.Context, rclient client.Client, cr *vmv1alp
 	}
 
 	// wait for empty persistent queue
-	if err := zs.waitForEmptyPQ(ctx, rclient, defaultMetricsCheckInterval, i); err != nil {
-		return fmt.Errorf("zone=%s: failed to wait till VMAgent queue for VMCluster=%s is drained: %w", item, nsnCluster.String(), err)
+	zs.waitForEmptyPQ(ctx, rclient, defaultMetricsCheckInterval, i)
+	if ctx.Err() != nil {
+		return fmt.Errorf("zone=%s: failed to wait till VMAgent queue for VMCluster=%s is drained", item, nsnCluster.String())
 	}
 
 	// restore zone in VMAuth LB
@@ -290,7 +291,7 @@ func getMetricsAddrs(ctx context.Context, rclient client.Client, vmAgent *vmv1be
 	return addrs
 }
 
-func (zs *zones) waitForEmptyPQ(ctx context.Context, rclient client.Client, interval time.Duration, clusterIdx int) error {
+func (zs *zones) waitForEmptyPQ(ctx context.Context, rclient client.Client, interval time.Duration, clusterIdx int) {
 	vmCluster := zs.vmclusters[clusterIdx]
 	clusterURLHash := fmt.Sprintf("%016X", xxhash.Sum64([]byte(vmCluster.GetRemoteWriteURL())))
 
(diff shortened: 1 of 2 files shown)


AndrewChubatiuk:
this solution most likely doesn't cover a case described in an issue, when VMCluster reconciliation for some reason returned context.Canceled and VMDistributed waits for it's readiness forever

internal/controller/operator/controllers.go
```diff
@@ -126,6 +127,10 @@ func handleReconcileErr[T client.Object, ST reconcile.StatusWithMetadata[STC], S
 	switch {
 	case errors.Is(err, context.Canceled):
 		contextCancelErrorsTotal.Inc()
+		var errZone *vmdistributed.ErrZone
+		if errors.As(err, &errZone) {
+			return ctrl.Result{Requeue: true}, nil
```
AndrewChubatiuk: why not just ignore cause from cmd/main.go and requeue for all others cases?
in this case other causes are not needed

AndrewChubatiuk: added it here https://github.com/VictoriaMetrics/operator/pull/1964/changes/00bc895aef09dbd5935d494e19d82c2db7e91ef8

vrutkovs: Good idea, thanks

AndrewChubatiuk: also let's keep only this cancelWithCause and drop the rest

vrutkovs: This would be removed in #1964 anyway? I want to keep this PR minimal so that it can be backported to 0.68 (not so sure about #1964).

Perhaps its easier to merge these changes in #1964?

AndrewChubatiuk: I mean all WithCancelCause added in this PR. Let's drop the rest besides one, which actually impacts reconcile behaviour, #1964 initially for a different purpose, it keeps only one function for reconcile errors handling and processes all reconcile errors in this function

internal/controller/operator/factory/vmdistributed/zone.go
```diff
@@ -324,17 +325,6 @@ func (zs *zones) waitForEmptyPQ(ctx context.Context, rclient client.Client, inte
 	}
 
 	var wg sync.WaitGroup
-	var resultErr error
```
vrutkovs, line 327: I like that we remove this ugly block, but not sure its worth removing return error and read it from the context instead

AndrewChubatiuk: this function returns nothing besides context.Canceled, other errors are treated as transient

vrutkovs: Ah, okay, lets roll with it then. Could you LGTM the PR?

PR #1966, opened 2026-03-15
operator/deployment: replace httpGet probes with tcpSocket

### LINKED ISSUE #1824: Add TCP startup checks
When TLS / mTLS is enabled existing healthchecks would fail as they expect the response to be plain text.

After a discussion with the team we've figures out that the healthchecks need to be reworked:
* These are not checking if the application is running, they are checking whether web server is running. As a result, these can be checked via TCP probe
* The checks are not updated throughout the lifecycle of the app, so instead of liveness / readiness check these can be converted into startupProbe

Plan:
* Document current workaround for TLS users - custom startup probe, unset liveness/readiness checks 
* Add TCP probe startup checks
* Announce deprecation in liveness/readiness checks
* After 3 releases liveness / readiness checks will be removed - this will cause pod rollout
* Document healthcheck strategy

----
Original issue

### Is your feature request related to a problem? Please describe

With mTLS enabled in k8s environment kubelet has to have a valid certificate for liveness / readiness checks. This makes pod setup cumbersome

Provide an unauthenticated `/healthz` endpoint, which doesn't require mTLS or TLS at all

Running a heathcheck via a command in the pod
Operator instrumenting a pod with correct certificates

### Additional information

_No response_

sudharsans:
## What

Replace `livenessProbe` and `readinessProbe` `httpGet` actions with `tcpSocket` in the operator deployment manifest.

## Why
The current `httpGet` probes fail when the health-probe server is served over TLS, because the kubelet sends a plain HTTP request.

`tcpSocket` probes only check that the port is accepting connections — agnostic to HTTP vs HTTPS — so they work correctly for both configurations.

Fixes #1824

---
## Summary by cubic
Replace the operator deployment’s liveness/readiness `httpGet` probes with `tcpSocket` so health checks work over both HTTP and HTTPS. Fixes #1824 by avoiding plain-HTTP requests against TLS health endpoints.

<sup>Written for commit a207047b47198fe7461e1760dbb97448dd894f26. Summary will update on new commits.</sup>

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -74,14 +74,12 @@ spec:
             drop:
             - "ALL"
         livenessProbe:
-          httpGet:
-            path: /health
+          tcpSocket:
             port: 8081
           initialDelaySeconds: 15
           periodSeconds: 20
         readinessProbe:
-          httpGet:
-            path: /ready
+          tcpSocket:
             port: 8081
           initialDelaySeconds: 5
           periodSeconds: 10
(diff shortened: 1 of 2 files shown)


AndrewChubatiuk:
#1824 was related to CR probes, not operator ones

AndrewChubatiuk:
closing this PR as not relevant

Issue #1970, opened 2026-03-16
Cluster is operational prematurely on PVC expand


vrutkovs:
<img width="3584" height="1184" alt="Image" src="https://github.com/user-attachments/assets/68207984-c22d-4a35-9ab7-7ec35b074645" />
<img width="3576" height="1572" alt="Image" src="https://github.com/user-attachments/assets/de653558-e30a-481c-90ea-ed3eb49a7963" />

AndrewChubatiuk:
fix will be available in upcoming 0.69.0 and 0.68.4

AndrewChubatiuk:
fix is available in 0.68.4

PR #1971, opened 2026-03-16
reconcile: wait till PVC is ready

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1970

---
## Summary by cubic
Wait for PVCs to be fully ready (Bound with requested capacity and not Resizing) before continuing reconcile and during StatefulSet PVC expansion. This pauses pod rollouts until storage is actually usable.

- **Bug Fixes**
  - Added `waitForPVCReady` with polling via `VM_PVC_WAIT_READY_INTERVAL` and `VM_PVC_WAIT_READY_TIMEOUT`; waits for requested capacity, PVC to be Bound, and not Resizing; skips if PVC is deleting.
  - Wrapped PVC get/create/update in a retry-on-conflict flow and skipped updates for PVCs with a non-zero DeletionTimestamp.
  - Invoked the wait after PVC reconcile and for each PVC in StatefulSet expansion; unified readiness polling via config (`reconcile.Init` now takes `BaseOperatorConf`; added `VM_WAIT_READY_INTERVAL` for VM CRs; renamed internals to `AppWaitReadyTimeout` and `PodWaitReadyInterval`, env names unchanged).

<sup>Written for commit 1dbeca63fa1837f510f65d0949c793b8b2d7ccc6. Summary will update on new commits.</sup>


internal/controller/operator/factory/reconcile/pvc.go
```diff
@@ -20,24 +21,58 @@ import (
...
+	case !newObj.CreationTimestamp.IsZero():
+		generation = newObj.Generation
+	case !existingObj.CreationTimestamp.IsZero():
+		generation = existingObj.Generation
 	}
+	return waitForPVCBound(ctx, rclient, nsn, generation)
+}
 
-	return updatePVC(ctx, rclient, &existingObj, newObj, prevObj, owner)
+func waitForPVCBound(ctx context.Context, rclient client.Client, nsn types.NamespacedName, generation int64) error {
```
vrutkovs: Lets add unit tests for this function

internal/controller/operator/factory/reconcile/pvc.go
```diff
@@ -20,24 +21,58 @@ import (
...
+		if err := rclient.Get(ctx, nsn, &pvc); err != nil {
+			if k8serrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, fmt.Errorf("cannot get PVC=%s: %w", nsn.String(), err)
+		}
+		if !pvc.DeletionTimestamp.IsZero() {
+			return true, fmt.Errorf("cannot wait for PVC=%s, which is being terminated", nsn.String())
+		}
+		if generation > pvc.Generation {
```
vrutkovs: Not sure why we need generation here at all - and why would PVC generation match the deployment?

AndrewChubatiuk: it's a generation from PVC, that was received during create/update, passing it to waitForPVCReady to ensure subsequent Get API call doesn't return stale object

AndrewChubatiuk:
@cubic-dev-ai

internal/controller/operator/factory/reconcile/pvc.go
AndrewChubatiuk: replaced generation with requested size

internal/controller/operator/factory/reconcile/pvc.go
```diff
@@ -20,24 +22,62 @@ import (
...
+		return updatePVC(ctx, rclient, &existingObj, newObj, prevObj, owner)
+	})
+	if err != nil {
+		return err
+	}
+	size := newObj.Spec.Resources.Requests[corev1.ResourceStorage]
+	if !existingObj.CreationTimestamp.IsZero() {
+		size = existingObj.Spec.Resources.Requests[corev1.ResourceStorage]
+	}
+	if err = waitForPVCReady(ctx, rclient, nsn, size); err != nil {
```
vrutkovs, line 47: We should be waiting for ready PVC only after DeletionTimestamp check? If VMCluster is being delete we don't need to wait for bound PVC

internal/controller/operator/factory/reconcile/pvc.go
```diff
@@ -20,24 +22,62 @@ import (
...
+		if err := rclient.Get(ctx, nsn, &pvc); err != nil {
+			if k8serrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, fmt.Errorf("cannot get PVC=%s: %w", nsn.String(), err)
+		}
+		if !pvc.DeletionTimestamp.IsZero() {
+			return true, nil
+		}
+		if len(pvc.Status.Capacity) == 0 {
```
vrutkovs, line 69: Could it be incomplete status? Not sure why we need this check

internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
vrutkovs, line 121: We should be waiting for status here. Doing this one by one would protect us from accidentally breaking all PVCs

internal/controller/operator/factory/reconcile/pvc.go
AndrewChubatiuk: there are two loops: one for creation/update, another one - waiting for PVC readiness. deletion may happen in each of these loops, that's why checking deletion timestamp in both, exiting without error and showing a warning. No further reconciliation is needed till user performs manual actions

internal/controller/operator/factory/reconcile/pvc.go
AndrewChubatiuk: it's needed for vm/vl/vt single, that are creating PVC before Deployment. In this case PVC creation will hang

internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
AndrewChubatiuk: updated

internal/controller/operator/factory/reconcile/pvc.go
vrutkovs: Ah, that makes sense

AndrewChubatiuk:
@vrutkovs could you please take a look?

PR #1985, opened 2026-03-18
test: add tests for useProxyProtocol setting

vrutkovs:
Ensure that we configure correct healthchecks for components with proxy protocol enabled on all deployment types.

Previously we had issues with that, hopefully https://github.com/VictoriaMetrics/operator/pull/1916 has fixed all of those


vrutkovs:
Curiously enough these tests are finding issues in 0.68 branch but not in 0.67 or latest master - somehow we regressed in LTS?

PR #1992, opened 2026-03-20
 Replace action versions with their commits' SHA for all workflows

arkid15r:


AndrewChubatiuk:
can we keep versions in comments at least?

vrutkovs:
I don't think we can keep them in sync (renovate can't do that at least)

arkid15r:
> I don't think we can keep them in sync (renovate can't do that at least)

Neither dependabot last time I checked.

PR #2006, opened 2026-03-30
defaults: do not use clusterversion for vl/vt requests load balancer tag

AndrewChubatiuk:
unset VL/VT clusterVersion for requests load balancer as it is not applicable for VMAuth

---
## Summary by cubic
Use the RequestsLoadBalancer ComponentVersion directly for VT/VL clusters, ignoring the clusterVersion. The VMAuth tag is now independent from the cluster tag; image tag still overrides, and if only clusterVersion is set we fall back to the default metrics version.

<sup>Written for commit 01d4189149952e6cec1f5d743f1910ec9d461b88. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/build/defaults.go
+++ b/internal/controller/operator/factory/build/defaults.go
@@ -598,7 +598,7 @@ func addVTClusterDefaults(objI any) {
 
 	if cr.Spec.RequestsLoadBalancer.Enabled {
 		cpLB := cp
-		cpLB.tag = setTag(cr.Spec.RequestsLoadBalancer.Spec.ComponentVersion, cp.tag)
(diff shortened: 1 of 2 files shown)


internal/controller/operator/factory/build/defaults.go
```diff
@@ -598,7 +598,7 @@ func addVTClusterDefaults(objI any) {
 
 	if cr.Spec.RequestsLoadBalancer.Enabled {
 		cpLB := cp
-		cpLB.tag = setTag(cr.Spec.RequestsLoadBalancer.Spec.ComponentVersion, cp.tag)
```
vrutkovs, line 601: Good catch, lets add a few unit tests to make sure we don't regress

Issue #1988, opened 2026-03-20
Apply generational sorting for VMAuth with VMDistributed


vrutkovs:
We sort zones so that failed or behind zones are updated first, so that functioning zones are left untouched.
However, the same order is used to VMAuth configuration. We should reverse that order in vmauth LB config, so that healthy, untouched zones would be prioritized - otherwise freshly updated, potentially unstable zone would be preferred.

AndrewChubatiuk:
instead of targets sorting vmauth configuration will contain only targets, that were upgraded by vmdistributed controller (i.e. target has VMDistributed owner reference). fix will be available in upcoming 0.68.4 and 0.69.0 releases

AndrewChubatiuk:
fix is available in 0.68.4

PR #2010, opened 2026-03-30
vmdistributed: filter vmauth targets by owner reference

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1988
downside of this solution: migration from distributed chart with a use of existing vmauth may lead to downtime till first zone becomes available after upgrade

---
## Summary by cubic
Filters VMAuth targets to include only VMAgents and VMClusters owned by the same VMDistributed. This prevents VMAuth from routing to unrelated resources.

- **Bug Fixes**
  - Added owner reference check when building VMAuth `DefaultTargetRefs` for VMAgents and VMClusters.
  - Updated tests to set owner references and validate the new filtering behavior.

<sup>Written for commit 983c6af8b8a40a63b69d347001b3ce8d6c50f8ed. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/vmdistributed/vmauth.go
+++ b/internal/controller/operator/factory/vmdistributed/vmauth.go
@@ -12,14 +12,24 @@ import (
 	"github.com/VictoriaMetrics/operator/internal/controller/operator/factory/build"
 )
 
-func vmClusterTargetRef(vmClusters []*vmv1beta1.VMCluster, excludeIds ...int) vmv1beta1.TargetRef {
+func hasOwnerReference(owners []metav1.OwnerReference, owner *metav1.OwnerReference) bool {
+	for i := range owners {
+		o := &owners[i]
+		if o.APIVersion == owner.APIVersion && o.Kind == owner.Kind && o.Name == owner.Name {
+			return true
+		}
+	}
+	return false
+}
+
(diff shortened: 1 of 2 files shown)


internal/controller/operator/factory/vmdistributed/vmauth.go
```diff
@@ -12,14 +12,24 @@ import (
...
+
+func vmClusterTargetRef(vmClusters []*vmv1beta1.VMCluster, owner *metav1.OwnerReference, excludeIds ...int) vmv1beta1.TargetRef {
 	var nsns []vmv1beta1.NamespacedName
 	for i := range vmClusters {
 		if slices.Contains(excludeIds, i) {
 			continue
 		}
 		vmCluster := vmClusters[i]
-		if vmCluster.CreationTimestamp.IsZero() {
+		if vmCluster.CreationTimestamp.IsZero() || !hasOwnerReference(vmCluster.OwnerReferences, owner) {
```
vrutkovs, line 32: That would stop VMAuth from routing to unrelated resources, but it won't change the order?

AndrewChubatiuk: right, I've checked communication, which initiated this issue and looks like this solution should cover that case.

PR #2013, opened 2026-03-30
reconcile: do not swallow error that may happen if STS PVC update failed

AndrewChubatiuk:
move PVC expansion out of STS reconcile to prevent from swallowing context deadline errors if PVC expansion took too long

---
## Summary by cubic
Move PVC expansion out of the StatefulSet reconcile retry loop and add conflict-safe PVC updates, so resize errors are returned instead of being swallowed. Tests now simulate in-progress and failing resizes and confirm `VMCluster` stays Expanding on retryable errors.

- **Bug Fixes**
  - Run PVC resize via `updateSTSPVC` after the retry block; retry PVC Get+Update with `retryOnConflict` and wait on the desired claim size.
  - Add `k8stools.ClientOpts.SkipPVCStatusUpdate`, pass it to `GetInterceptorsWithObjects`, and expose `GetTestClientWithOptsActionsAndObjects` for tests.
  - Update tests to cover retryable resize (Expanding status) and failing waits; adjust expected actions to include PVC Get before Update.

<sup>Written for commit a299dd8f8634596bd11c845b9f2f23b24fc310f2. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/reconcile/statefulset_pvc_expand_test.go
+++ b/internal/controller/operator/factory/reconcile/statefulset_pvc_expand_test.go
@@ -564,81 +564,72 @@ func Test_updateSTSPVC(t *testing.T) {
 	})
 
 	// failed with non-expandable sc
-	// BROKEN: unlimited number of retries here
-	// f(opts{
-	// 	sts: buildSTS(func(sts *appsv1.StatefulSet) {
-	// 		sts.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{
-	// 			{
-	// 				ObjectMeta: metav1.ObjectMeta{
-	// 					Name:   "vmselect-cachedir",
-	// 					Labels: map[string]string{"app": "vmselect"},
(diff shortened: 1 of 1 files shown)


internal/controller/operator/factory/reconcile/statefulset.go
```diff
@@ -153,6 +146,13 @@ func StatefulSet(ctx context.Context, rclient client.Client, newObj, prevObj *ap
 		return err
 	}
 
+	// check if pvcs need to resize
+	if len(existingObj.Spec.VolumeClaimTemplates) > 0 {
+		if err := updateSTSPVC(ctx, rclient, newObj, prevVCTs); err != nil {
```
vrutkovs, line 151: I think we should wrap in this `retryOnConflict` loop too?

vrutkovs: Actually, it would be best to handle retries within `updateSTSPVC`

PR #2031, opened 2026-04-03
test: speedup upgrade tests

AndrewChubatiuk:
## Summary by cubic
Speeds up and stabilizes upgrade e2e tests by consolidating rollout checks, auto-discovering expected workloads from VM CRDs (including logs/traces), and simplifying the Kind/test harness. Operator upgrades now use a full `OPERATOR_IMAGE` pullspec with an in-cluster Deployment for reliable, registry-agnostic rollouts.

- **Refactors**
  - Consolidated helpers into `getSnapshots`, `checkWorkloads`, and `getRolloutDiff`; `getApplications` derives expected `Deployment`/`StatefulSet`/`DaemonSet` from CRDs, covering agents, singles, alerts, anomalies, clusters, `VMDistributed` zones, and `VL*`/`VT*`.
  - Operator rollout via in-cluster `Deployment`: updates with `OPERATOR_IMAGE`, waits for `ObservedGeneration` and `ReadyReplicas`, ensures old pod is removed; supports extra envs (sets `VM_CONFIG_RELOADER_IMAGE`); tests randomized; 60m timeout.
  - Simplified Kind flow to `./kind.yaml`; removed OLM and local-registry logic/targets; tests pass `OPERATOR_IMAGE`; deleted `config/olm/*`.
  - Parallel-safe setup with Ginkgo `SynchronizedBeforeSuite`/`SynchronizedAfterSuite`; isolated, commented test cases for easier maintenance.
  - VTCluster storage pod spec simplified to reduce rollout diff noise.

- **Bug Fixes**
  - Fixed diff path generation for empty field names; added a deep inline-nesting test for embedded fields.
  - Treat `IsAlreadyExists` as conflict to ensure retries.
  - Skip creating `VMAuth` LB `DefaultTargetRefs` when no targets exist.

<sup>Written for commit a18c2d556b87f5b84ac42f58973138e45417a2df. Summary will update on new commits.</sup>

### DIFF
--- a/test/e2e/upgrade/upgrade_test.go
+++ b/test/e2e/upgrade/upgrade_test.go
@@ -426,23 +426,24 @@ var (
 	licenseKey          = os.Getenv("LICENSE_KEY")
 )
 
+type crVersionPair struct {
+	version string
+	cr      client.Object
+}
+
 type entry struct {
-	name     string
-	genDeps  func(string) []client.Object
-	crs      []client.Object
-	versions []string
-	envs     map[string]string
+	name    string
+	genDeps func(string) []client.Object
+	pairs   []crVersionPair
+	envs    map[string]string
 }
 
 func entries(es []entry) []TableEntry {
 	var result []TableEntry
 	for _, e := range es {
-		for _, v := range e.versions {
-			var objs []client.Object
-			for _, cr := range e.crs {
-				objs = append(objs, cr.DeepCopyObject().(client.Object))
-			}
-			result = append(result, Entry(fmt.Sprintf("from %s: %s", v, e.name), v, e.genDeps, objs, e.envs))
+		for _, p := range e.pairs {
+			obj := p.cr.DeepCopyObject().(client.Object)
+			result = append(result, Entry(fmt.Sprintf("from %s: %s (%T)", p.version, e.name, obj), p.version, e.genDeps, []client.Object{obj}, e.envs))
 		}
 	}
 	return result
@@ -549,6 +550,7 @@ func ensureNoPodRollout(version string, genDeps func(string) []client.Object, ob
 
 var _ = Describe("operator upgrade", Label("upgrade"), func() {
 	DescribeTable("should not rollout changes", ensureNoPodRollout, entries([]entry{
+		// nolint:dupl
 		{
 			name: "VMAgent/VLAgent",
 			genDeps: func(ns string) []client.Object {
@@ -594,22 +596,59 @@ var _ = Describe("operator upgrade", Label("upgrade"), func() {
 					},
 				}
 			},
-			crs: []client.Object{
-				with(vmagent),
-				with(vmagent, func(cr *vmv1beta1.VMAgent) {
+			pairs: []crVersionPair{
+				{version: "v0.68.0", cr: with(vmagent)},
+				{version: "v0.68.0", cr: with(vmagent, func(cr *vmv1beta1.VMAgent) {
 					cr.Spec.DaemonSetMode = true
-				}),
-				with(vmagent, func(cr *vmv1beta1.VMAgent) {
+				})},
+				{version: "v0.68.0", cr: with(vmagent, func(cr *vmv1beta1.VMAgent) {
 					cr.Spec.StatefulMode = true
-				}),
-				with(vlagent),
-				with(vlagent, func(cr *vmv1.VLAgent) {
+				})},
+				{version: "v0.68.0", cr: with(vlagent)},
+				{version: "v0.68.0", cr: with(vlagent, func(cr *vmv1.VLAgent) {
 					cr.Spec.K8sCollector.Enabled = true
 					cr.Spec.ServiceAccountName = "vlagent-collector"
-				}),
+				})},
+				{version: "v0.68.1", cr: with(vmagent)},
+				{version: "v0.68.1", cr: with(vmagent, func(cr *vmv1beta1.VMAgent) {
+					cr.Spec.DaemonSetMode = true
+				})},
+				{version: "v0.68.1", cr: with(vmagent, func(cr *vmv1beta1.VMAgent) {
+					cr.Spec.StatefulMode = true
+				})},
+				{version: "v0.68.1", cr: with(vlagent)},
+				{version: "v0.68.1", cr: with(vlagent, func(cr *vmv1.VLAgent) {
+					cr.Spec.K8sCollector.Enabled = true
+					cr.Spec.ServiceAccountName = "vlagent-collector"
+				})},
+				{version: "v0.68.2", cr: with(vmagent)},
+				{version: "v0.68.2", cr: with(vmagent, func(cr *vmv1beta1.VMAgent) {
+					cr.Spec.DaemonSetMode = true
+				})},
+				{version: "v0.68.2", cr: with(vmagent, func(cr *vmv1beta1.VMAgent) {
+					cr.Spec.StatefulMode = true
+				})},
+				{version: "v0.68.2", cr: with(vlagent)},
+				{version: "v0.68.2", cr: with(vlagent, func(cr *vmv1.VLAgent) {
+					cr.Spec.K8sCollector.Enabled = true
+					cr.Spec.ServiceAccountName = "vlagent-collector"
+				})},
+				{version: "v0.68.3", cr: with(vmagent)},
+				{version: "v0.68.3", cr: with(vmagent, func(cr *vmv1beta1.VMAgent) {
+					cr.Spec.DaemonSetMode = true
+				})},
+				{version: "v0.68.3", cr: with(vmagent, func(cr *vmv1beta1.VMAgent) {
+					cr.Spec.StatefulMode = true
+				})},
+				{version: "v0.68.3", cr: with(vlagent)},
+				{version: "v0.68.3", cr: with(vlagent, func(cr *vmv1.VLAgent) {
+					cr.Spec.K8sCollector.Enabled = true
+					cr.Spec.ServiceAccountName = "vlagent-collector"
+				})},
 			},
-			versions: []string{"v0.68.0", "v0.68.1", "v0.68.2", "v0.68.3"},
-		}, {
+		},
+		// nolint:dupl
+		{
 			name: "VMAlert/VMAuth/VMAlertmanager/VMAnomaly",
 			genDeps: func(ns string) []client.Object {
 				return []client.Object{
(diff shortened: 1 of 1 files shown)


internal/controller/operator/factory/vmdistributed/vmauth.go
```diff
@@ -94,8 +94,15 @@ func buildVMAuthLB(cr *vmv1alpha1.VMDistributed, vmAgents []*vmv1beta1.VMAgent,
 	}
 	var targetRefs []vmv1beta1.TargetRef
 	owner := cr.AsOwner()
-	targetRefs = append(targetRefs, vmAgentTargetRef(vmAgents, &owner, excludeIds...))
-	targetRefs = append(targetRefs, vmClusterTargetRef(vmClusters, &owner, excludeIds...))
+	if ref := vmAgentTargetRef(vmAgents, &owner, excludeIds...); len(ref.CRD.Objects) > 0 {
```
vrutkovs, line 97: I think this will limit the customization of `vmauth`, so we should document that only CRDs are supported

test/e2e/upgrade/upgrade_test.go
```diff
@@ -1491,588 +144,554 @@ var _ = Describe("operator upgrade", Label("upgrade"), func() {
...
+		})
+	}
+	wg.Wait()
 
-		restartManagerAndCleanup(ctx, k8sClient, namespace)
+	By("snapshotting child workload specs")
+	apps := getApplications(objs...)
+	podSpecs := getSnapshots(ctx, k8sClient, apps...)
+
+	updateOperator(ctx, k8sClient, "docker.io", currentVersion, namespace, envs)
```
vrutkovs:
```suggestion
	updateOperator(ctx, k8sClient, "quay.io", currentVersion, namespace, envs)
```
So that we wouldn't accidentally hit their limits

test/e2e/upgrade/utils.go
```diff
@@ -77,118 +79,138 @@ func operatorEnvVars(watchNamespace string) []corev1.EnvVar {
...
 	var result []corev1.EnvVar
+	maps.Copy(envs, extraEnvs)
 	for k, v := range envs {
 		result = append(result, corev1.EnvVar{Name: k, Value: v})
 	}
 	return result
 }
 
-func deployOldOperator(ctx context.Context, k8sClient client.Client, version, watchNamespace string) {
+func updateOperator(ctx context.Context, k8sClient client.Client, registry, version, watchNamespace string, envs map[string]string) {
```
vrutkovs: That would update operator to latest stable version, but not the current master version? So the tests would be failing only after the release

test/e2e/upgrade/upgrade_test.go
```diff
@@ -1491,588 +144,554 @@ var _ = Describe("operator upgrade", Label("upgrade"), func() {
...
-		Entry("from v0.68.1", "v0.68.1"),
-		// TODO[vrutkovs]: Regression?
-		Entry("from v0.68.2", "v0.68.2"),
-		Entry("from v0.68.3", "v0.68.3"),
-	)
+var _ = Describe("operator upgrade", Label("upgrade"), func() {
+	DescribeTable("should not rollout changes", ensureNoPodRollout, entries([]entry{
+		{
+			name: "VMAgent/VLAgent",
+			genDeps: func(ns string) []client.Object {
```
vrutkovs, line 556: Not sure why we need that - wouldn't the operator create these resources automatically?

test/e2e/upgrade/utils.go
```diff
@@ -77,118 +79,138 @@ func operatorEnvVars(watchNamespace string) []corev1.EnvVar {
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      "vm-operator",
-			Namespace: watchNamespace,
-		},
-		Spec: appsv1.DeploymentSpec{
-			Replicas: ptr.To[int32](1),
-			Selector: &metav1.LabelSelector{
-				MatchLabels: map[string]string{"app": "vm-operator"},
+	var dep appsv1.Deployment
+	if err := k8sClient.Get(ctx, nsn, &dep); err == nil {
```
vrutkovs, line 103: We're always in a random namespace, I think finding an existing deployment should be an error

test/e2e/upgrade/utils.go
```diff
@@ -23,25 +22,28 @@ import (
...
+func operatorEnvVars(watchNamespace string, extraEnvs map[string]string) []corev1.EnvVar {
 	envs := map[string]string{
-		"VM_CONTAINERREGISTRY":                           "quay.io",
 		"VM_VMALERTMANAGER_ALERTMANAGERDEFAULTBASEIMAGE": "prometheus/alertmanager",
 		"VM_ENABLEDPROMETHEUSCONVERTEROWNERREFERENCES":   "true",
 		"VM_GATEWAY_API_ENABLED":                         "true",
 		"VM_PODWAITREADYTIMEOUT":                         "20s",
 		"VM_PODWAITREADYINTERVALCHECK":                   "1s",
 		"VM_APPREADYTIMEOUT":                             "50s",
+		"VM_CONFIG_RELOADER_IMAGE":                       "quay.io/victoriametrics/operator:config-reloader-v0.68.3",
```
vrutkovs, line 44: I don;t think we need to specify this, let the operator use whatever is set by default?

test/e2e/upgrade/upgrade_test.go
```diff
@@ -1491,588 +144,554 @@ var _ = Describe("operator upgrade", Label("upgrade"), func() {
...
-		Entry("from v0.68.0", "v0.68.0"),
-		Entry("from v0.68.1", "v0.68.1"),
-		// TODO[vrutkovs]: Regression?
-		Entry("from v0.68.2", "v0.68.2"),
-		Entry("from v0.68.3", "v0.68.3"),
-	)
+var _ = Describe("operator upgrade", Label("upgrade"), func() {
+	DescribeTable("should not rollout changes", ensureNoPodRollout, entries([]entry{
+		{
+			name: "VMAgent/VLAgent",
```
vrutkovs, line 555: I think its best to split the cases so that we could easily comment out the failing cases and have a constent test case history

Makefile
```diff
@@ -283,7 +281,7 @@ build-installer: manifests generate kustomize ## Generate a consolidated YAML wi
 	$(KUSTOMIZE) build config/crd/overlay > dist/crd.yaml
 
 olm: operator-sdk opm yq docs
-	$(eval DIGEST = $(shell $(CONTAINER_TOOL) buildx imagetools inspect $(REGISTRY)/$(ORG)/$(REPO):$(TAG)-ubi --format "{{print .Manifest.Digest}}"))
+	$(eval DIGEST = $(shell $(CONTAINER_TOOL) buildx imagetools inspect $(REGISTRY)/$(ORG)/$(REPO):$(TAG) --format "{{print .Manifest.Digest}}"))
```
vrutkovs: This will break OpenShift operatorhub tests - they require UBI images

Makefile
```diff
@@ -303,13 +301,6 @@ olm: operator-sdk opm yq docs
 		bundle/$(VERSION)/manifests/victoriametrics-operator.clusterserviceversion.yaml
 	$(YQ) -i '.annotations."com.redhat.openshift.versions" = "v4.12-v4.21"' \
 		bundle/$(VERSION)/metadata/annotations.yaml
-	$(if $(findstring localhost,$(REGISTRY)), \
```
vrutkovs, line 306: not sure why is this being removed - lets move this to a separate PR?

test/e2e/upgrade/utils.go
AndrewChubatiuk: it will update to a version that is built by `make test-e2e`, which builds `docker.io/victoriametrics/operator:<OPERATOR_TAG>` image and loads to kind

test/e2e/upgrade/upgrade_test.go
AndrewChubatiuk: nope, VLAgent in namespaced mode requires RBAC to be configured manually

test/e2e/upgrade/utils.go
AndrewChubatiuk: it's needed for updating existing operator

test/e2e/upgrade/utils.go
AndrewChubatiuk: config reloader default version depends on operator version, which causes pods recreation

test/e2e/upgrade/upgrade_test.go
AndrewChubatiuk: it doesn't pull anything, it uses already loaded to kind locally built image

internal/controller/operator/factory/vmdistributed/vmauth.go
AndrewChubatiuk: it doesn't. this change is only done to avoid deploying when no backends are available, which will lead to distributed failure as vmauth requires backends. some tests have single zone (upgrade tests), without this change and vmauth enabled tests are constantly failing

Makefile
AndrewChubatiuk: this is a special case, which allows deploying olm package locally. removed it as currently without FBC it requires having local registry and opm ignores locally loaded images - always tries to pull it. this change doesn't impact release process

Makefile
AndrewChubatiuk:
```suggestion
	$(eval DIGEST = $(shell $(CONTAINER_TOOL) buildx imagetools inspect $(REGISTRY)/$(ORG)/$(REPO):$(TAG)-ubi --format "{{print .Manifest.Digest}}"))
```

internal/controller/operator/factory/vmdistributed/vmauth.go
vrutkovs: >some tests have single zone (upgrade tests), without this change and vmauth enabled tests are constantly failing

In that case the test needs to be fixed and a separe PR created for this - I think there might be a configuration where having no CRDs is valid

test/e2e/upgrade/upgrade_test.go
vrutkovs: ah, the test is expected to run with current operator built as `docker.io`? Perhaps we should be leave this as configurable?

test/e2e/upgrade/utils.go
```diff
@@ -77,118 +79,133 @@ func operatorEnvVars(watchNamespace string) []corev1.EnvVar {
...
+			return false
+		}
+		replicas := ptr.Deref(dep.Spec.Replicas, 0)
+		status := dep.Status
+		if status.ObservedGeneration < dep.Generation {
+			return false
 		}
-		if d.Status.ReadyReplicas < 1 {
-			return fmt.Errorf("operator not ready yet, readyReplicas=%d", d.Status.ReadyReplicas)
+		if status.ReadyReplicas != replicas || status.Replicas != replicas {
```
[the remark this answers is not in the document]
vrutkovs: We're waiting for new ReplicaSet to rollout AND the previous one to be downscaled. Existing code only waits for new one to rollout, so with one replica we declare operator finished only once the old replica has began to terminate. This is easy to verify by using invalid OPERATOR_TAG - the operator will start pods, they crash with ImagePullBackoff, but the test continues

test/e2e/upgrade/upgrade_test.go
vrutkovs: Updated it to accept OPERATOR_IMAGE instead (it would be constructed by the Makefile)

internal/controller/operator/factory/vmdistributed/vmauth.go
AndrewChubatiuk: https://github.com/VictoriaMetrics/operator/pull/2035

test/e2e/upgrade/upgrade_test.go
vrutkovs: Done in [a18c2d5](https://github.com/VictoriaMetrics/operator/pull/2031/commits/a18c2d556b87f5b84ac42f58973138e45417a2df)

PR #2035, opened 2026-04-07
vmdistributed: use default vmauth stub, when no healthy zones available

AndrewChubatiuk:
skip updating vmauth, when no healthy zones available since vmauth fails when no backends available

Alternatively makes sense to render some static content, when no backend set, but requires either adding a sidecar or adding this feature to vmauth

---
## Summary by cubic
Use a static stub backend for `VMAuth` in `vmdistributed` when no healthy zones are available so configs stay valid and the controller still updates `VMAuth`.

- **Bug Fixes**
  - Always populate `DefaultTargetRefs` for read/write; if no `VMCluster`/`VMAgent` targets exist, set `Static` to `http://127.0.0.1:9999`.
  - Add a reconcile test to verify `VMAuth` updates even with no healthy zones; update the changelog.

<sup>Written for commit aeee07754e7aaf2a10cd4b596751bcf96454ddad. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/vmdistributed/vmdistributed_reconcile_test.go
+++ b/internal/controller/operator/factory/vmdistributed/vmdistributed_reconcile_test.go
@@ -317,7 +317,7 @@ func Test_CreateOrUpdate_Actions(t *testing.T) {
 		},
 	})
 
-	// skip VMAuth update, when distributed setup has a single zone
+	// update VMAuth even, when no healthy zones available
 	f(args{
 		cr: &vmv1alpha1.VMDistributed{
 			ObjectMeta: objectMeta,
@@ -379,6 +379,11 @@ func Test_CreateOrUpdate_Actions(t *testing.T) {
 			{Verb: "Get", Kind: "VMCluster", Resource: vmClusterName},
 			{Verb: "Get", Kind: "VMAgent", Resource: vmAgentName},
 
+			// update LB config
+			{Verb: "Get", Kind: "VMAuth", Resource: vmAuthLBName},
+			{Verb: "Update", Kind: "VMAuth", Resource: vmAuthLBName},
+			{Verb: "Get", Kind: "VMAuth", Resource: vmAuthLBName},
+
 			// reconcile VMCluster
 			{Verb: "Get", Kind: "VMCluster", Resource: vmClusterName},
 			{Verb: "Update", Kind: "VMCluster", Resource: vmClusterName},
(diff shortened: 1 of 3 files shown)


internal/controller/operator/factory/vmdistributed/vmdistributed_reconcile_test.go
```diff
@@ -316,6 +316,83 @@ func Test_CreateOrUpdate_Actions(t *testing.T) {
 			{Verb: "Get", Kind: "VMAuth", Resource: vmAuthLBName},
 		},
 	})
+
+	// skip VMAuth update, when distributed setup has a single zone
```
vrutkovs: I think we want to throw a meaningful error here - "VMDistributed expects at least two zones to be set"

AndrewChubatiuk: this solution doesn't cover cases:
- adding existing setup to vmdistributed or creating a new one. since vmauth adds targets that have owner reference only upgrade of first zone leads to absent vmauth targets
- there's a feature request to be able to exclude zones from read and write. in case if only one zone left upgrade can also lead to no vmauth targets

vrutkovs: Ah yes, we need this check on create only then

vrutkovs (approved):

internal/controller/operator/factory/vmdistributed/vmdistributed_reconcile_test.go
```diff
@@ -316,6 +316,89 @@ func Test_CreateOrUpdate_Actions(t *testing.T) {
...
+			{Verb: "Update", Kind: "VMCluster", Resource: vmClusterName},
+			{Verb: "Get", Kind: "VMCluster", Resource: vmClusterName},
+
+			// reconcile VMAgent
+			{Verb: "Get", Kind: "VMAgent", Resource: vmAgentName},
+			{Verb: "Get", Kind: "VMAgent", Resource: vmAgentName},
+
+			// reconcile VMAuth
+			{Verb: "Get", Kind: "VMAuth", Resource: vmAuthLBName},
+			{Verb: "Update", Kind: "VMAuth", Resource: vmAuthLBName},
```
vrutkovs, line 398: We still update it here, so probably we need a better test:
* prepare VMDistributed w/ 2 zones
* delete both zones
* ensure that vmauth / vmagent sections have static stub address

AndrewChubatiuk: changed title, in tests, PR and changelog. solution is not different

PR #2053, opened 2026-04-10
feat: Add CLI tool for converting helm values

### LINKED ISSUE #1796: Helm-chart -> operator objects converter
helm-charts are great way to get started with VM installations and convenient way to customize the installation. They are however not helping much when it comes to operational tasks like preventing disruption on upgrade and orchestrating multiple cluster maintenance.

This is the area where operator excels, as it can customly handle changes from one state to the other. Its also designed to work in GitOps environment and  allows admins to separate configuration into several CRs with different area of concern.

We lack a convenient way of converting helm-based installations into operator-managed ones. It would be nice to have a tool, which can generate operator resources from values files. Its meant to be a one-shot tool and doesn't have to be precise, but it would be handy for the users who are migrating from PoC type of installation into a production-like system

vrutkovs:
Add a new CLI tool to convert a Helm values file into operator manifests.

TODO:
* [x] Add default values

Fixes https://github.com/VictoriaMetrics/operator/issues/1796


cmd/helm-converter/main.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"os"
+
+	k8syaml "sigs.k8s.io/yaml"
+
+	"github.com/VictoriaMetrics/operator/internal/converter"
+)
+
+var (
+	inputFile  = flag.String("input", "", "input file with helm values")
```
AndrewChubatiuk, line 14: there should be a requirement for users to keep in values file defaults as well

vrutkovs: Oof, that's unlikely to be followed often. I think the tool should be able to do that (assuming its the latest version of the chart)

vrutkovs: Implemented that

AndrewChubatiuk (approved):

PR #2085, opened 2026-04-21
renamed vmbackup and vmrestore ports

AndrewChubatiuk:
## Summary by cubic
Renamed `vmbackup` and `vmrestore` container port names to avoid collisions with `vmsingle`/`vmstorage`, fixing Kubernetes validation and Service issues. Added `VM_USE_OLD_BACKUP_RESTORE_PORT_NAMES` to keep old names during upgrades and avoid pod restarts.

- **Bug Fixes**
  - VMBackupManager port: `http-backup` (was `http`)
  - VMRestore port: `http-restore` (was `http`)
  - Applies to `vmcluster` and `vmsingle`

- **Migration**
  - Update Service/Ingress/probes to use `http-backup` and `http-restore` (no change if using port numbers)
  - To avoid restarts during upgrade, set `VM_USE_OLD_BACKUP_RESTORE_PORT_NAMES=true`; remove after migration

<sup>Written for commit 9892d8bdf1b64c438b51ae46967de24253a0653a. Summary will update on new commits.</sup>


vrutkovs (approved):

vrutkovs (commented):
I think this needs a way to revert it (via env var) so that the operator would not restart these pods outside of maintenance periods

PR #2097, opened 2026-04-23
Common labels

### LINKED ISSUE #2094: commonLabels and commonAnnotations
Unify existing options to set commonLabels and commonAnnotations

vrutkovs:
This ensures HTTPRoutes and PVCs include ManagedMetadata labels and annotations.

Fixes #2094 

TODO:
* [x] Implement passing commonLabel/managedMetadata in pod templates too


api/operator/v1beta1/vmcluster_types_test.go
```diff
@@ -4,7 +4,10 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
+
+	"github.com/VictoriaMetrics/operator/internal/config"
```
AndrewChubatiuk: let's avoid adding any `github.com/VictoriaMetrics/operator/internal/*` package in api. probably it can be done in defaults.go instead

AndrewChubatiuk (approved):
Overall looks good! could you please also add test to cover an attempt to overwrite standard labels using commonLabels?

internal/controller/operator/factory/build/defaults.go
```diff
@@ -48,11 +49,19 @@ func AddDefaults(scheme *runtime.Scheme) {
 	scheme.AddTypeDefaultingFunc(&vmv1.VMAnomaly{}, addVMAnomalyDefaults)
 	scheme.AddTypeDefaultingFunc(&vmv1beta1.VMServiceScrape{}, addVMServiceScrapeDefaults)
 	scheme.AddTypeDefaultingFunc(&vmv1alpha1.VMDistributed{}, addVMDistributedDefaults)
+
+	scheme.AddTypeDefaultingFunc(&appsv1.DaemonSet{}, addDefaultMetadata)
+	scheme.AddTypeDefaultingFunc(&corev1.ConfigMap{}, addDefaultMetadata)
+	scheme.AddTypeDefaultingFunc(&corev1.Namespace{}, addDefaultMetadata)
```
AndrewChubatiuk, line 57: not sure this is needed, we don't manage namespaces

AndrewChubatiuk: also don't understand why this function should be applied to all core k8s resources? isn't it enough to extend addDefaultMetadata to patch ManagedMetadata? also not sure if this should be applied to pods as well

vrutkovs: I'd rather add those preemptively, as some feature may start managing those resources later.

>also don't understand why this function should be applied to all core k8s resources?

My understanding is that if the object (`VMAgent`) creates a deployment and the service, we should make sure both have propagated labels and annotations. I'm not sure that omitting pods / child resources is sufficient for OPA.

> also not sure if this should be applied to pods as well

I think so, yeah

internal/controller/operator/factory/build/httproute.go
```diff
@@ -16,7 +16,7 @@ import (
 // HTTPRoute creates HTTPRoute object
 func HTTPRoute(cr builderOpts, port string, httpRoute *vmv1beta1.EmbeddedHTTPRoute) (*gwapiv1.HTTPRoute, error) {
 
-	lbls := labels.Merge(httpRoute.Labels, cr.SelectorLabels())
+	lbls := labels.Merge(labels.Merge(cr.FinalLabels(), httpRoute.Labels), cr.SelectorLabels())
```
AndrewChubatiuk: checked most of FinalLabels implementations, they already include SelectorLabels

internal/controller/operator/factory/vlsingle/vlsingle.go
```diff
@@ -48,8 +48,8 @@ func newPVC(r *vmv1.VLSingle) *corev1.PersistentVolumeClaim {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            r.PrefixedName(),
 			Namespace:       r.Namespace,
-			Labels:          labels.Merge(r.Spec.StorageMetadata.Labels, r.SelectorLabels()),
-			Annotations:     r.Spec.StorageMetadata.Annotations,
+			Labels:          labels.Merge(labels.Merge(r.FinalLabels(), r.Spec.StorageMetadata.Labels), r.SelectorLabels()),
+			Annotations:     labels.Merge(r.FinalAnnotations(), r.Spec.StorageMetadata.Annotations),
```
AndrewChubatiuk, line 52: not sure we need final annotations here. cfg.CommonAnnotations should be enough

internal/controller/operator/factory/build/httproute.go
vrutkovs: Good idea, not need to merge those again

AndrewChubatiuk:
@vrutkovs 
Thanks for a PR

PR #2099, opened 2026-04-24
vmdistributed: change default write requests load balancing policy from first_available to least_loaded

AndrewChubatiuk:
## Summary by cubic
Switches the default load balancing policy for write requests in `vmdistributed` from `first_available` to `least_loaded` to spread write load evenly across VMAgents. Read requests remain on `first_available`.

- **Bug Fixes**
  - Update `vmauth` "write" target to use `least_loaded`; retry codes unchanged.
  - Adjust tests to expect the new default and add explicit policies for write/read.
  - Add changelog entry.

<sup>Written for commit 6a9d422f3fbdf8598e73349c99c56b293b143e19. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/vmdistributed/vmauth.go
+++ b/internal/controller/operator/factory/vmdistributed/vmauth.go
@@ -78,7 +78,7 @@ func vmAgentTargetRef(vmAgents []*vmv1beta1.VMAgent, owner *metav1.OwnerReferenc
 	ref := vmv1beta1.TargetRef{
 		Name: "write",
 		URLMapCommon: vmv1beta1.URLMapCommon{
-			LoadBalancingPolicy: ptr.To("first_available"),
+			LoadBalancingPolicy: ptr.To("least_loaded"),
 			RetryStatusCodes:    []int{500, 502, 503},
 		},
 		Paths: []string{"/insert/.+", "/api/v1/write"},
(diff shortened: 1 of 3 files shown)


internal/controller/operator/factory/vmdistributed/vmauth.go
```diff
@@ -78,7 +78,7 @@ func vmAgentTargetRef(vmAgents []*vmv1beta1.VMAgent, owner *metav1.OwnerReferenc
 	ref := vmv1beta1.TargetRef{
 		Name: "write",
 		URLMapCommon: vmv1beta1.URLMapCommon{
-			LoadBalancingPolicy: ptr.To("first_available"),
+			LoadBalancingPolicy: ptr.To("least_loaded"),
```
vrutkovs, line 81: I think it needs to be complimented with removing exclude IDs from vmAgentTargetRef, so that we'd be writing in all zones (queues would grow when its being updated)

AndrewChubatiuk: vmagent is excluded as well in case if it can be restarted due to upgrade

vrutkovs (approved):

vrutkovs:
Thanks, could you backport it to 0.68.x too please?

PR #2100, opened 2026-04-24
vmdistributed: support trafficMode

### LINKED ISSUE #1995: vmdistributed: configure read-only and write-only zones
Moving from distributed helm chart to vmdistributed CRD. The ability to configure a zone as read-only or write-only was lost. Read-Only zones can be important for migrations or disasters like running AWS disk too close to EBS limits. Write-Only is very useful for maintenance tasks.

Not urgent, just capturing a feature that is useful.

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1995

---
## Summary by cubic
Adds per-zone traffic modes to `vmdistributed` to control read and write routing without removing a zone. `vmauth` now builds separate read and write targets and applies each zone’s mode during updates.

- **New Features**
  - Added `spec.zones[*].trafficMode` with `read-write` (default), `read-only`, `write-only`, `maintenance` (CRDs, defaults, docs updated).
  - `vmauth` LB: `read-only` zones are excluded from write targets (VMAgent); `write-only` from read targets (VMCluster); `maintenance` from both.
  - In `read-only` or `maintenance`, the zone’s VMInsert scales to 0 replicas to block writes.
  - Drains VMAgent queues only when a zone stops or starts accepting writes to avoid unnecessary waits around LB updates.

<sup>Written for commit a761f933bb8919e5f43be2a25686f82ba3c2b37d. Summary will update on new commits.</sup>

### DIFF
--- a/api/operator/v1alpha1/vmdistributed_types.go
+++ b/api/operator/v1alpha1/vmdistributed_types.go
@@ -82,11 +82,23 @@ type VMDistributedZoneCommon struct {
 	UpdatePause *metav1.Duration `json:"updatePause,omitempty"`
 }
 
+type VMDistributedTrafficMode string
+
+const (
+	VMDistributedTrafficModeReadWrite   VMDistributedTrafficMode = "read-write"
+	VMDistributedTrafficModeReadOnly    VMDistributedTrafficMode = "read-only"
+	VMDistributedTrafficModeWriteOnly   VMDistributedTrafficMode = "write-only"
+	VMDistributedTrafficModeMaintenance VMDistributedTrafficMode = "maintenance"
+)
+
 // +k8s:openapi-gen=true
 // VMDistributedZone defines items within a single zone to update.
 type VMDistributedZone struct {
 	// Name defines a name of zone, which can be used in zoneCommon spec as %ZONE%
 	Name string `json:"name"`
+	// TrafficMode defines allowed traffic mode for a zone: read-only, write-only, read-write, maintenance
+	// +kubebuilder:validation:Enum=read-only;write-only;read-write;maintenance
+	TrafficMode VMDistributedTrafficMode `json:"trafficMode,omitempty"`
 	// VMCluster defines a new inline or referencing existing one VMCluster
 	// +optional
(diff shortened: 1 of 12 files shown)


api/operator/v1alpha1/vmdistributed_types.go
```diff
@@ -84,11 +84,23 @@ type VMDistributedZoneCommon struct {
 	UpdatePause *metav1.Duration `json:"updatePause,omitempty"`
 }
 
+type VMDistributedTrafficMode string
+
+const (
+	VMDistributedTrafficModeReadWrite VMDistributedTrafficMode = "read-write"
+	VMDistributedTrafficModeReadOnly  VMDistributedTrafficMode = "read-only"
+	VMDistributedTrafficModeWriteOnly VMDistributedTrafficMode = "write-only"
+	VMDistributedTrafficModeNone      VMDistributedTrafficMode = "none"
```
vrutkovs: This is a bit confusing - `none` can be interpreted as "no mode set". Maybe `maintenance` mode instead?

internal/controller/operator/factory/vmdistributed/zone.go
```diff
@@ -102,8 +105,12 @@ func getZones(ctx context.Context, rclient client.Client, cr *vmv1alpha1.VMDistr
 		prevClusterSpec := vmCluster.Spec
 		vmCluster.Spec = *vmClusterSpec
 		rclient.Scheme().Default(&vmCluster)
+		if (z.TrafficMode == vmv1alpha1.VMDistributedTrafficModeReadOnly || z.TrafficMode == vmv1alpha1.VMDistributedTrafficModeMaintenance) && vmCluster.Spec.VMInsert != nil {
+			vmCluster.Spec.VMInsert.ReplicaCount = ptr.To(int32(0))
```
vrutkovs, line 114: Wouldn't that make vmagent start accumulating the queue for RO clusters? Also, needs a comment why we do this

vrutkovs (commented):
Looks good overall, but it needs a few more unit tests and e2e test preferably - and example / documentation update

vrutkovs (approved):

PR #2124, opened 2026-04-29
reconcile: prevent from staying in expanding state

AndrewChubatiuk:
## Summary by cubic
Always propagate reconcile errors from `reconcileAndTrackStatus` so the controller requeues and resources don’t get stuck in Expanding. Set `resultErr` for both Expanding and Failed outcomes, update tests to expect errors in retryable paths (conflict, wait interrupted/PVC resize), and add a `vmoperator` changelog entry.

<sup>Written for commit 9741a7effab6aceee3c9a6fba47e2c3764d7d872. Summary will update on new commits.</sup>


internal/controller/operator/controllers.go
```diff
@@ -345,8 +345,8 @@ func reconcileAndTrackStatus[T client.Object, ST reconcile.StatusWithMetadata[ST
 			resultStatus = vmv1beta1.UpdateStatusExpanding
 		} else {
 			resultStatus = vmv1beta1.UpdateStatusFailed
-			resultErr = err
 		}
+		resultErr = err
```
vrutkovs, line 349:
```suggestion
		resultStatus := vmv1beta1.UpdateStatusFailed
		if reconcile.IsRetryable(err) {
			resultStatus = vmv1beta1.UpdateStatusExpanding
		}
		resultErr = err
```

vrutkovs (commented):
LGTM, minor nitpick

internal/controller/operator/controllers.go
vrutkovs: I'll fix it in a follow-up PR - https://github.com/VictoriaMetrics/operator/pull/2140/commits

vrutkovs (approved):

PR #2131, opened 2026-05-02
fix(backup): stop passing deprecated -eula flag to vmbackupmanager

### LINKED ISSUE #1319: Remove usage of `-eula` flag
`-eula` flag have been deprecated last year and will be removed from VictoriaMetrics shortly.

Currently, `vmbackupmanaget` configuration still requires `acceptEULA` [flag](https://docs.victoriametrics.com/operator/api/#vmbackup-accepteula) in CR configuration.
It is needed to remove `-eula` flag usage in order to make sure there will be no incompatibility issues once flag will be removed.

alliasgher:
The `-eula` flag was deprecated in VictoriaMetrics and will be removed, causing a startup error in `vmbackupmanager` once that happens. Remove the two places in `build/backup.go` where the operator appends `-eula` to the container args.

The `AcceptEULA` field on the CR is intentionally left in place so existing configurations that set it don't require a breaking change.

Fixes #1319

---
## Summary by cubic
Stop passing the deprecated `-eula` flag to `vmbackupmanager` in backup and restore args so it starts cleanly on current and future VictoriaMetrics versions. The `AcceptEULA` CR field stays for backward compatibility but no longer changes container args.

<sup>Written for commit 1cd828d047d0d6d0f8afff9cfa7cb7bb8e76e852. Summary will update on new commits.</sup>

### DIFF
--- a/internal/controller/operator/factory/build/backup.go
+++ b/internal/controller/operator/factory/build/backup.go
@@ -55,9 +55,7 @@ func VMBackupManager(
 		fmt.Sprintf("-snapshot.createURL=%s", snapshotCreateURL),
 		fmt.Sprintf("-snapshot.deleteURL=%s", snapshotDeleteURL),
 	}
-	if cr.AcceptEULA {
-		args = append(args, "-eula")
-	}
+	// -eula flag was deprecated and removed from VictoriaMetrics; no longer passed.
 	if cr.LogLevel != nil {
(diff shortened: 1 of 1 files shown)


vrutkovs:
The bug was already fixed in https://github.com/VictoriaMetrics/operator/pull/2042 - this PR would ignore `AcceptEULA` setting which is not what we want

PR #2153, opened 2026-05-13
Speed up vmdistributed tests

vrutkovs:
* Don't wait for PQ check when VMAgent is just created
* Speed up metrics checks, replace `WaitForEmptyPQStub` with a stub in most cases


vrutkovs:
Doesn't seem to speed up anything :/

PR #2168, opened 2026-05-14
pin actions to full-length commit SHAs

rudransh-shrivastava:
Notes:
`actions/checkout` was downgraded by 1 commit due to version tag mismatch. The removed commit [0c366f](https://github.com/actions/checkout/commit/0c366fd6a839edf440554fa01a7085ccba70ac98) is a changelog documentation update with no corresponding release tag.

### DIFF
--- a/.github/workflows/crds.yaml
+++ b/.github/workflows/crds.yaml
@@ -16,14 +16,14 @@ jobs:
       name: docs
     steps:
       - name: Check out operator code
-        uses: actions/checkout@0c366fd6a839edf440554fa01a7085ccba70ac98
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
         with:
           repository: VictoriaMetrics/operator
           token: ${{ secrets.VM_BOT_GH_TOKEN }}
           path: __vm-operator-repo
 
       - name: Check out VM code
-        uses: actions/checkout@0c366fd6a839edf440554fa01a7085ccba70ac98
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
         with:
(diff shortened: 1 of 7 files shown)


vrutkovs:
>actions/checkout was downgraded by 1 commit due to version tag mismatch.

I think renovate would attempt to bring it back. Could you check their documentation if they have an option to use only tagged commit hashes?

rudransh-shrivastava:
> I think renovate would attempt to bring it back. Could you check their documentation if they have an option to use only tagged commit hashes?

Hi, The [documentation](https://docs.renovatebot.com/modules/manager/github-actions/#digest-pinning-and-updating) states:
> Actions pinned to a bare SHA without a version comment are disabled by default, because Renovate cannot determine which branch or tag the SHA belongs to. To enable updates, add a tag or branch name as a version comment, as shown above.

However, the docs seem wrong as earlier it tried to update the action like you describe (likely tracking the default branch?): https://github.com/VictoriaMetrics/operator/pull/2020

I think when version is specified using a comment, it tracks tags instead of latest commit. When version is not specified, it defaults to tracking untagged commits.

Do you think [helpers:pinGitHubActionDigestsToSemver](https://docs.renovatebot.com/presets-helpers/#helperspingithubactiondigeststosemver) is what we need?

rudransh-shrivastava:
Some additional context: Yesterday, I merged this PR in my fork and renovate was able to detect the versions. It seemed okay with `v6.0.2` and didn't make a PR for `actions/checkout` / `0c366f`


Before:
Screenshot 2026-05-14 at 2 12 09 PM

After:
image

vrutkovs:
Ah, great, thanks for looking into that. Feel free to make it ready for review to merge this

vrutkovs (approved):

vrutkovs:
Thank you for the contribution!

PR #2169, opened 2026-05-14
feat(vmrule): add patchMergeKey to Groups for kubectl/kustomize patches

### LINKED ISSUE #657: Add patchMergeKey to VMRule CRDs
The CRDs do not define patchMergeKey and patchStrategy for both group names and rule alert fields.

https://github.com/VictoriaMetrics/operator/blob/master/api/v1beta1/vmrule_types.go

The lack of merge keys makes it hard to adjust the rules and alerts with both kubectl patch and kustomize. This is especially prudent when you deploy the default rules with the stack chart, which can be quite noisy.

For example if one uses the stack chart deployed with flux, but wants to adjust some rules after the fact, they may used a kustomize postrenderer (or otherwise directly with kubectl patch) to try and and change a given rule (e.g. increasing the required alert time to 15m -> 60m):

```
   kind: VMRule    
   metadata:    
     name: stack-kubernetes-apps    
     namespace: victoria-metrics    
   spec:    
     groups:    
       - name: kubernetes-apps    
         rules:    
           - alert: KubeDeploymentGenerationMismatch                                              
             for: 60m 
```

Without the merge key the entire list of groups is overwritten, replacing everything with the list in the patch. The only current solution is to redefine the entirety of the Rules in your patch or to otherwise use a JSON patch with the numeric index of the group/rule you want to patch (which may change as new version of the charts are pushed). While in theory the chart could allow for rule customization, I would expect the CRDs to specify merging on group name and alert rule field.

This currently affects v0.34.0 of the operator and apiVersion operator.victoriametrics.com/v1beta1.

mvanhorn:
The VMRule CRD did not define `patchMergeKey`/`patchStrategy` on `spec.groups`, so `kubectl patch` and `kustomize` had to replace the entire group list to adjust any single group's rules. This made it painful to tweak a single chart-provided rule (e.g. raising an alert's `for:` from 60m to 90m) without redefining every other group in the patch.

This change adds name-based merging to `VMRuleSpec.Groups`:

- Source: 4-line marker block (`+patchMergeKey`, `+patchStrategy`, `+listType=map`, `+listMapKey=name`) + struct tag mirror, matching the existing `Conditions` precedent in `api/operator/v1beta1/vmextra_types.go:1384`.
- Generated: `config/crd/overlay/crd.yaml` and `crd.descriptionless.yaml` now carry `x-kubernetes-list-type: map` and `x-kubernetes-list-map-keys: [name]` on the groups field. Output of `make manifests`; no manual YAML edits.

Scoped to groups for this PR. The inner `Rules []Rule` list has the same pain in principle, but `Rule` carries either `Alert` or `Record` (never both required), so a single `listMapKey` is brittle there - happy to follow up in a separate PR if maintainers want a specific approach (e.g. compound keys via `+listMapKey` twice, or a synthetic merge field).

## Verification

- `make manifests` regenerated both CRD overlays with only the new map markers - no unrelated drift.
- `make api-gen` produced no further diff.
- `make docs` produced no further diff to `docs/api.md`.
- `make lint` clean (0 issues across the `api/...` and main modules).
- `go test ./operator/v1beta1/...` from `./api`: passing.

Closes #657

---
## Summary by cubic
Enable name-based merge for VMRule `spec.groups` so `kubectl patch` and `kustomize` can update a single group without replacing the whole list. This makes small tweaks (like changing an alert’s for: 60m → 90m) safer and easier.

- **New Features**
  - Added `+patchMergeKey=name`, `+patchStrategy=merge`, `+listType=map`, and `+listMapKey=name` to `VMRuleSpec.Groups`.
  - Regenerated CRDs to include `x-kubernetes-list-type: map` and `x-kubernetes-list-map-keys: [name]` for `groups`.

<sup>Written for commit b50f41d2c8c7a8299adbefce5334b4ef85b287fd. Summary will update on new commits.</sup>

### DIFF
--- a/api/operator/v1beta1/vmrule_types.go
+++ b/api/operator/v1beta1/vmrule_types.go
@@ -26,7 +26,11 @@ var initVMAlertTemplatesOnce sync.Once
 // VMRuleSpec defines the desired state of VMRule
 type VMRuleSpec struct {
 	// Groups list of group rules
-	Groups []RuleGroup `json:"groups"`
+	// +patchMergeKey=name
+	// +patchStrategy=merge
+	// +listType=map
+	// +listMapKey=name
+	Groups []RuleGroup `json:"groups" patchStrategy:"merge" patchMergeKey:"name"`
 }
 
 // RuleGroup is a list of sequentially evaluated recording and alerting rules.
--- a/config/crd/overlay/crd.descriptionless.yaml
+++ b/config/crd/overlay/crd.descriptionless.yaml
@@ -34036,6 +34036,9 @@ spec:
                   - rules
                   type: object
                 type: array
+                x-kubernetes-list-map-keys:
+                - name
+                x-kubernetes-list-type: map
             required:
             - groups
             type: object


vrutkovs (approved):
That makes sense, thank you for your contribution!

PR #2172, opened 2026-05-15
Halt emitting config-reloader signals when delay-interval is customized

### LINKED ISSUE #2171: config-reloader doesn't respect delay-interval
We're running a delay in goroutine, which doesn't halt other goroutines from running, so this interval is effectively a noop

vrutkovs:
Each reload called a separate goroutine, instead we should use delay-interval to prevent other goroutines from running.

Fixes #2171

### DIFF
--- a/cmd/config-reloader/main.go
+++ b/cmd/config-reloader/main.go
@@ -275,15 +275,14 @@ func (c *cfgWatcher) close() {
 }
 
 func (c *cfgWatcher) waitDelay(ctx context.Context) bool {
-	if *delayInterval <= 0 {
-		return true
-	}
-	t := time.NewTimer(*delayInterval)
-	defer t.Stop()
-	select {
-	case <-t.C:
-	case <-ctx.Done():
-		return false
+	if *delayInterval > 0 {
+		t := time.NewTimer(*delayInterval)
+		defer t.Stop()
+		select {
+		case <-t.C:
+		case <-ctx.Done():
+			return false
+		}
 	}
 	for {
 		select {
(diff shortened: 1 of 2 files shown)


cmd/config-reloader/main.go
```diff
@@ -251,26 +251,34 @@ func (c *cfgWatcher) start(ctx context.Context) {
 		for {
 			select {
 			case <-c.updates:
-				go func() {
```
AndrewChubatiuk, line 254: what do you think about goto-less approach?

```
for {
  select {
    case <-c.updates:
    case <-ctx.Done():
      return
    }
    if !c.waitDelay(ctx) {
      return
    }
    if err := c.reloader(ctx); err != nil {
      logger.Errorf("cannot trigger api reload: %s", err.Error())
      configLastReloadSuccess.Set(0)
      configReloadErrorsTotal.Inc()
      continue
    }
    configLastReloadSuccess.Set(1)
    configLastOkReloadTime.Set(uint64(time.Now().UnixMilli()))
    logger.Infof("reload config ok.")
  }
```

with `waitDelay` function
```
func (c *cfgWatcher) waitDelay(ctx context.Context) bool {
  if *delayInterval > 0 {
    t := time.NewTimer(*delayInterval)
    defer t.Stop()
    select {
    case <-t.C:
    case <-ctx.Done():
      return false
    }
  }
  for {
    select {
    case <-c.updates:
    default:
      return true
    }
  }
}
```

vrutkovs: It would still run in the same goroutine - this is why we need a goto to "exit" it. I'm also not happy with `goto` but the suggested code doesn't pass the tests

AndrewChubatiuk: i've updated code, it should pass

vrutkovs: Oh nevermind, I misread this - pushed it as a commit

AndrewChubatiuk (approved):

PR #2174, opened 2026-05-15
e2e: removed validating job

AndrewChubatiuk:
## Summary by cubic
Removed the in-cluster Job used for HTTP validation in e2e tests. We now call services via the API server proxy and use watch-based status/deletion waits that fail fast on errors to reduce flakiness and speed up runs.

- **Refactors**
  - Dropped `batch/v1` Job + `curl`; HTTP checks now use `rest.HTTPClientFor` (10s timeout) via the API server proxy, auto-POSTing when a payload is provided.
  - Added `GetRestConfig` and plumbed `rest.Config` (`k8sCfg`) into tests.
  - Implemented `buildServiceProxyURL` to reach in-cluster services without Pods.
  - Replaced polling `Eventually` with watch-based waits and deletion helpers (`expectStatusAfterAction`, `WatchUntilStatusReached/Seen`, `waitResourceDeleted`/`waitObjectDeleted`) with generation guards; watchers/status checks now fail fast on `UpdateStatusFailed`; set default polling interval to 2s.

<sup>Written for commit ad2d5e7f0cad50b068152627a57a2b3e239d1e6a. Summary will update on new commits. <a href="https://cubic.dev/pr/VictoriaMetrics/operator/pull/2174?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/test/e2e/utils_test.go
+++ b/test/e2e/utils_test.go
@@ -264,42 +264,67 @@ func mustGetFirstPod(ctx context.Context, rclient client.Client, obj client.Obje
 	return &pods[0]
 }
 
-//nolint:dupl,lll
-func waitResourceDeleted(ctx context.Context, rclient client.Client, nss types.NamespacedName, r client.Object) {
+func waitObjectDeleted(ctx context.Context, obj client.Object) {
 	GinkgoHelper()
+	nsn := types.NamespacedName{Name: obj.GetName(), Namespace: obj.GetNamespace()}
 	Eventually(func() error {
-		return rclient.Get(ctx, nss, r)
-	}, eventualDeletionTimeout).Should(MatchError(k8serrors.IsNotFound, "IsNotFound"))
+		return k8sClient.Get(ctx, nsn, obj.DeepCopyObject().(client.Object))
+	}, eventualDeletionTimeout).Should(MatchError(k8serrors.IsNotFound, "isNotFound"))
+}
+
+//nolint:dupl,lll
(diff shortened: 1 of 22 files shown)


test/e2e/utils_test.go
```diff
@@ -166,76 +168,52 @@ func expectHTTPRequestToSucceed(ctx context.Context, object httpRequestCRDObject
...
+	}
+	parts := strings.SplitN(u.Hostname(), ".", 3)
+	if len(parts) < 2 {
+		return "", fmt.Errorf("cannot parse namespace from service URL: %s", svcURL)
+	}
+	proxyPath := u.Path
+	if u.RawQuery != "" {
+		proxyPath += "?" + u.RawQuery
+	}
+	return fmt.Sprintf("%s/api/v1/namespaces/%s/services/%s:%s/proxy%s",
```
vrutkovs, line 188: I think we removed `nodes/proxy` RBAC permission from the operator anymore - would it still work?

AndrewChubatiuk: it uses services/proxy, which is already present in cluster-admin role, which is a default role for cli credentials generated by kind

test/e2e/vmanomaly_test.go
```diff
@@ -56,7 +57,6 @@ schedulers:
 
 var (
 	anomalyReadyTimeout  = 120 * time.Second
-	anomalyDeleteTimeout = 60 * time.Second
```
vrutkovs, line 59: I'm a bit concerned that we now use the same delete timeout for all objects, but probably its acceptable

vrutkovs (approved):

PR #2194, opened 2026-05-19
feat(api/operator): Add EnableServiceLinks field

AyCarlito:
This PR:
- Surface configuration of `EnableServiceLinks` in the `Deployment` spec.
- This field is optional, and enabled by default.
- When enabled, the Kubelet injects an environment variable for each active service in the namespace. With a sufficient number of services, this will cause [failures on pod start](https://dev.to/matthewdailey/post-mortem-kubernetes-pods-dont-start-because-of-too-many-services-1129).

---
## Summary by cubic
Add an optional `enableServiceLinks` flag across operator CRDs and pod templates to control Kubelet service env var injection. Defaults to true; set it to false to prevent pod start failures in namespaces with many Services.

- **New Features**
  - Added `enableServiceLinks` to `CommonAppsParams` (optional).
  - Controller wires the value to `pod.spec.enableServiceLinks`.
  - Updated all CRD schemas (`config/crd/overlay/*.yaml`), docs (`docs/api.md`), and changelog.
  - Added unit tests to verify nil/true/false propagation in `podtemplate`.

- **Migration**
  - No action needed. Set `enableServiceLinks: false` to disable service link env vars when desired.

<sup>Written for commit e215b804c62e0a52622dd4a0cda272d0ebd9caab. Summary will update on new commits. <a href="https://cubic.dev/pr/VictoriaMetrics/operator/pull/2194?utm_source=github">Review in cubic</a></sup>


internal/controller/operator/factory/build/deployment.go
```diff
@@ -21,6 +21,7 @@ func DeploymentAddCommonParams(dst *appsv1.Deployment, params *vmv1beta1.CommonA
 	dst.Spec.Template.Spec.HostNetwork = params.HostNetwork
 	dst.Spec.Template.Spec.DNSPolicy = params.DNSPolicy
 	dst.Spec.Template.Spec.DNSConfig = params.DNSConfig
+	dst.Spec.Template.Spec.EnableServiceLinks = params.EnableServiceLinks
```
vrutkovs: lets extend this to StatefulSets and DaemonSets as well

vrutkovs (commented):
Looks like a good start! Please also add unit tests to make sure its not accidentally removed later

internal/controller/operator/factory/build/deployment.go
AndrewChubatiuk: just rebase a branch, now defaults for podSpec are set in a single place

AndrewChubatiuk (approved):

AyCarlito:
Thanks for the help on this one! Can I ask what the typical release cadence is for this project and subsequently when these changes would be available? @AndrewChubatiuk @vrutkovs

PR #2203, opened 2026-05-21
ci: use digest instead of tag for image signing

kharf:
as tags are mutable, its better to use digests for signing and in future cosign will remove support for tag signing as seen in the build warnings: 

> WARNING: Image reference docker.io/victoriametrics/operator:v0.70.1 uses a tag, not a digest, to identify the image to sign.
    This can lead you to sign a different image than the intended one. Please use a
    digest (example.com/ubuntu@sha256:abc123...) rather than tag
    (example.com/ubuntu:latest) for the input to cosign. The ability to refer to
    images by tag will be removed in a future release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,6 +268,8 @@ publish:
 
 COSIGN ?= $(shell which cosign 2>/dev/null || echo $(COSIGN_BIN))
 
+digest = $(shell $(CONTAINER_TOOL) buildx imagetools inspect $(REGISTRY)/$(ORG)/$(REPO):$(1) --format "{{print .Manifest.Digest}}")
+
 .PHONY: cosign
 cosign: $(COSIGN_BIN)
 $(COSIGN_BIN): $(LOCALBIN)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,6 +9,11 @@ TAG ?= $(shell echo $$(git describe --long --all | tr '/' '-')$$( \
 	git diff-index --quiet HEAD -- || echo '-dirty-'$$( \
 		git diff-index -u HEAD -- ':!config' ':!docs' | openssl sha1 | cut -d' ' -f2 | cut -c 1-8)))
 OPERATOR_IMAGE ?= $(REGISTRY)/$(ORG)/$(REPO):$(TAG)
+OPERATOR_DIGEST      = $(shell $(CONTAINER_TOOL) buildx imagetools inspect $(REGISTRY)/$(ORG)/$(REPO):$(TAG) --format "{{print .Manifest.Digest}}")
```
AndrewChubatiuk: what do you think about using single macro instead?

```
digest = $(shell $(CONTAINER_TOOL) buildx imagetools inspect $(REGISTRY)/$(ORG)/$(REPO):$(1) --format "{{print .Manifest.Digest}}")
```

and then use it 

```
$(COSIGN) sign --yes $${registry}/$(ORG)/$(REPO)@$(call digest,$(TAG))
```

kharf: yea can do it. Should it be in lowercase like in your suggestion?

AndrewChubatiuk: other macros are in lowercase, makes sense to make this one in lowercase too

kharf: > other macros are in lowercase, makes sense to make this one in lowercase too

yea thought so as well

AndrewChubatiuk (approved):

PR #2207, opened 2026-05-22
reconcile: update statefulset status on success for OnDelete update strategy

### LINKED ISSUE #1242: StatefulSet using OnDelete update strategy have stale currentRevision
The vmoperator manages the StatefulSets belonging to VMClusters and VMAlertmanagers using `OnDelete` `updateStrategy.type`. In such case, K8s statefulset controller does not update the `status.currentRevision` to the value of `status.updateRevision` when the last pod was successfully deleted and hence updated to the current controllerrevision. This cause stale information to be exposed in the status block, and moreover the `currentReplicas` to be missing:

```
status:
  availableReplicas: 6
  collisionCount: 0
  currentRevision: vmstorage-skr-697c996f7b
  observedGeneration: 38
  readyReplicas: 6
  replicas: 6
  updateRevision: vmstorage-skr-74c4bd7697
  updatedReplicas: 6
```

vmoperator could patch the statefulset/status subresource to have the `currentRevision` equal to the `updateRevision` when the last pod is deleted.

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/1242

---
## Summary by cubic
Update StatefulSet status for OnDelete strategy by patching status.currentRevision to match status.updateRevision and syncing status.currentReplicas after pods converge or when no updates are needed. Prevents stale status in monitoring and fixes #1242.

- **Bug Fixes**
  - Patch StatefulSet status after rollout completes and when no update is needed.
  - Set status.currentRevision and status.currentReplicas to reflect the actual state.
  - Add RBAC to get/update/patch apps/statefulsets/status for vlagent, vmagent, vmalertmanager, vmanomaly, and vmcluster; update changelog.

<sup>Written for commit 60f4e06faf587d1eacb3e5ed7531e44df3849939. Summary will update on new commits. <a href="https://cubic.dev/pr/VictoriaMetrics/operator/pull/2207?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/internal/controller/operator/factory/reconcile/statefulset.go
+++ b/internal/controller/operator/factory/reconcile/statefulset.go
@@ -212,13 +212,32 @@ type rollingUpdateOpts struct {
 	delete         bool
 }
 
+// patchSTSCurrentRevision patches statefulset status.currentRevision to match status.updateRevision
+// after all pods are updated. This is needed because Kubernetes does not update currentRevision
+// for OnDelete strategy, leaving stale data visible to monitoring tools.
+// See https://github.com/VictoriaMetrics/operator/issues/1242
+func patchSTSCurrentRevision(ctx context.Context, rclient client.Client, nsn types.NamespacedName, updateRevision string, replicas int32) error {
+	var sts appsv1.StatefulSet
+	if err := rclient.Get(ctx, nsn, &sts); err != nil {
+		return fmt.Errorf("cannot get statefulset for status update: %w", err)
+	}
+	if sts.Status.CurrentRevision == updateRevision {
+		return nil
+	}
+	logger.WithContext(ctx).Info(fmt.Sprintf("updating statefulset=%s/%s status currentRevision from %q to %q", nsn.Namespace, nsn.Name, sts.Status.CurrentRevision, updateRevision))
+	sts.Status.CurrentRevision = updateRevision
+	// currentReplicas is not updated with OnDelete strategy too
+	sts.Status.CurrentReplicas = replicas
+	if err := rclient.Status().Update(ctx, &sts); err != nil {
+		return fmt.Errorf("cannot update statefulset=%s/%s currentRevision status: %w", nsn.Namespace, nsn.Name, err)
+	}
+	return nil
+}
+
 // we perform rolling update on sts by manually evicting pods one by one or in batches
 // we check sts revision (kubernetes controller-manager is responsible for that)
 // and compare pods revision label with sts revision
 // if it doesn't match - updated is needed
-//
-// we always check if sts.Status.CurrentRevision needs update, to keep it equal to UpdateRevision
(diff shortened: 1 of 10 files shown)


internal/controller/operator/factory/reconcile/statefulset.go
```diff
@@ -212,13 +212,31 @@ type rollingUpdateOpts struct {
...
+// patchSTSCurrentRevision patches statefulset status.currentRevision to match status.updateRevision
+// after all pods are updated. This is needed because Kubernetes does not update currentRevision
+// for OnDelete strategy, leaving stale data visible to monitoring tools.
+// See https://github.com/VictoriaMetrics/operator/issues/1242
+func patchSTSCurrentRevision(ctx context.Context, rclient client.Client, nsn types.NamespacedName, updateRevision string, replicas int32) error {
+	var sts appsv1.StatefulSet
+	if err := rclient.Get(ctx, nsn, &sts); err != nil {
+		return fmt.Errorf("cannot get statefulset for status update: %w", err)
+	}
+	if sts.Status.CurrentRevision == updateRevision {
```
vrutkovs, line 224: We should exit early if the strategy is not `OnDelete`

internal/controller/operator/factory/reconcile/statefulset.go
```diff
@@ -212,13 +212,31 @@ type rollingUpdateOpts struct {
...
+		return nil
+	}
+	logger.WithContext(ctx).Info(fmt.Sprintf("updating statefulset=%s/%s status currentRevision from %q to %q", nsn.Namespace, nsn.Name, sts.Status.CurrentRevision, updateRevision))
+	sts.Status.CurrentRevision = updateRevision
+	sts.Status.CurrentReplicas = replicas
```
vrutkovs, line 230: Not sure why are we using desired number of replicas here - my understanding we should read `status.updatedReplicas`, which corresponds to the replicas from `updateRevision`

vrutkovs (changes requested):

internal/controller/operator/factory/reconcile/statefulset.go
AndrewChubatiuk: this function is executed inside performRollingUpdateOnSts, which in turn is only called when strategy is set to OnDelete

internal/controller/operator/factory/reconcile/statefulset.go
AndrewChubatiuk: status is only updated in case of success or if no actions are needed, which means statefulset is in desired state, where replicas count is equal to expected and all of them are healthy

vrutkovs: Add a comment about this so it won't be forgotten

vrutkovs (approved):

vrutkovs (approved):

PR #2218, opened 2026-05-27
chore: add version check to crds workflow

vrutkovs:
Make sure that LTS releases don't create PRs for helm-chart to update the CRDs

### DIFF
--- a/.github/workflows/crds.yaml
+++ b/.github/workflows/crds.yaml
@@ -32,7 +32,6 @@ jobs:
 
       - name: Check operator chart app version
         id: chart
-        continue-on-error: true
         env:
           RELEASE_VERSION: ${{ github.event.release.tag_name }}
         run: |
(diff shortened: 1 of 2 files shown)


AndrewChubatiuk (approved):

.github/workflows/crds.yaml
```diff
@@ -30,7 +30,35 @@ jobs:
...
+          RELEASE_VERSION: ${{ github.event.release.tag_name }}
+        run: |
+          chart_app_version=$(yq -r '.appVersion // ""' charts/victoria-metrics-operator/Chart.yaml)
+          if [ -z "$chart_app_version" ]; then
+            echo "failed to read chart appVersion"
+            exit 1
+          fi
+
+          set +e
+          python3 ../__vm-operator-repo/hack/compare_versions.py "$chart_app_version" "$RELEASE_VERSION"
```
AndrewChubatiuk: can this script just produce stdout value and store it to >> $GITHUB_OUTPUT?
then it chould be used in subsequent steps `if` section

vrutkovs: That was my first option but then I thought that some stdout may also leak there?

AndrewChubatiuk: only data explicitly written to GITHUB_OUTPUT is a step output, it doesn't capture whole stdout.

PR #2220, opened 2026-05-27
.github: enforce least permissions for github action jobs

rudransh-shrivastava:
Scope permissions to job level and set `permissions: {}` at workflow level.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -9,8 +9,7 @@ on:
       - 'README.md'
       - '.github/workflows/docs.yaml'
 
-permissions:
-  contents: write
+permissions: {}
 
 jobs:
   update-docs:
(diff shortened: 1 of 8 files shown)


.github/workflows/docs.yaml
```diff
@@ -9,15 +9,16 @@ on:
...
+permissions: {}
 
 jobs:
   update-docs:
     runs-on: ubuntu-latest
     environment:
       name: docs
       url: https://docs.victoriametrics.com/operator
+    permissions:
+      contents: read
```
vrutkovs: Lets reuse VM_BOT_GH_TOKEN so that we wouldn't have another read-only one?

.github/workflows/sandbox.yaml
```diff
@@ -15,13 +15,13 @@ on:
...
-permissions:
-  contents: read
-  pull-requests: write
+permissions: {}
 
 jobs:
   update:
     name: Update version on sandbox
+    permissions:
+      contents: read
```
vrutkovs: Same here - bot token instead of a new one

.github/workflows/docs.yaml
rudransh-shrivastava: Done!

.github/workflows/sandbox.yaml
rudransh-shrivastava: Done!

AndrewChubatiuk (approved):

Issue #916, opened 2024-04-03
Indtroduce metrics endpoint for config-reloader container. 


f41gh7:
I'd like to have the following metrics:

1) reload call errors to the vm endpoint.
2) last reload
3) reload rate
3) any watch errors

f41gh7:
Must be fixed at [v0.43.0](https://github.com/VictoriaMetrics/operator/releases/tag/v0.43.0) release

PR #2228, opened 2026-05-29
config-reloader: fix success timestamp metric units

immanuwell:
Fixes `configreloader_last_reload_success_timestamp_seconds`.

The metric name says seconds, but the reloader was writing `time.Now().UnixMilli()`.
So after any successful reload the exported value is 1000x too big. Kinda busted for dashboards and alerts.

Repro:
1. On `master`, run `go test ./cmd/config-reloader -run TestCfgWatcherSuccessTimestampUsesSeconds -count=1`
2. It fails with a value like `1780068551994`
3. This patch switches to `time.Now().Unix()` and adds the test, so it passes

Related: #916

### DIFF
--- a/cmd/config-reloader/main.go
+++ b/cmd/config-reloader/main.go
@@ -264,7 +264,7 @@ func (c *cfgWatcher) start(ctx context.Context) {
 				continue
 			}
 			configLastReloadSuccess.Set(1)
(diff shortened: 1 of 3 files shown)


AndrewChubatiuk (approved):

AndrewChubatiuk:
Thanks for contribution!

PR #2232, opened 2026-05-31
use multiple secrets if scrape config exceeds K8s secret limit

### LINKED ISSUE #1048: Operator is unable to create secret for vmagent when compressed config "vmagent.yaml.gz" is larger than 1MB
I am using latest build of the operator (v0.46.4) and vmagent (v1.102.0). I have lots of various endpoints to scrape. So Operator is trying to create secret for config "vmagent.yaml.gz" but it is larger than 1MB and managed AKS Kubernetes won't allow to create secret larger than 1MB (Ref-https://kubernetes.io/docs/concepts/configuration/secret/#working-with-secrets)

Error
` Warning  ReconcilationError  <unknown>  victoria-metrics-operator  Secret "vmagent-vmagent-0" is invalid: data: Too long: must have at most 1048576 bytes`

`NAME        SHARDS COUNT   REPLICA COUNT   STATUS
vmagent-0                  1               failed
vmagent-1                  1               failed
vmagent-2                  1               failed`

I also tried by enabling shards but no luck.

It is similar issue like https://github.com/prometheus-operator/prometheus-operator/issues/4702

I think fix will be like https://github.com/prometheus-operator/prometheus-operator/pull/4427 where operator creating shardSecret.

AndrewChubatiuk:
vmagent and vmsingle: use additional secrets if default scrape config size exceeds K8s secret limit. fixes #1048

---
## Summary by cubic
Splits oversized scrape configs for `vmagent` and `vmsingle` across multiple gzip-compressed Secrets, loading overflow jobs via `scrape_config_files`. `config-reloader` now supports paired `--watched-dir`/`--target-dir` to decompress and atomically sync files to an EmptyDir so readers never see partial state.

- **New Features**
  - `vmagent`/`vmsingle`: When compressed config exceeds the budget, scrape jobs are split into numbered gzip Secrets. The main config adds `scrape_config_files` only when overflow exists; stale extra Secrets are cleaned up. Pods mount `sc-files-out` for decompressed job files, and per-bucket Secret volumes are paired with `--watched-dir`/`--target-dir`.
  - `config-reloader`: New `--target-dir` paired by index with `--watched-dir` (counts must match). On init and change, auto-detects gzip, decompresses, and atomically swaps the target directory; performs an initial sync on startup to pre-populate targets.
  - Operator config: `VM_CONFIG_DATA_BUDGET_BYTES` (default 512KB) caps compressed data per Secret/ConfigMap. Used for scrape config bucketing and `vmalert` rule ConfigMap splitting.

- **Refactors**
  - Removed `VMRule` size validation tied to a single ConfigMap; rules are now split by the configured budget during reconcile.

<sup>Written for commit c6f3221a378abec0480a28207faba6b8fd1887d7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VictoriaMetrics/operator/pull/2232?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


AndrewChubatiuk:
@cubic-ai-dev

cmd/config-reloader/file_watch.go
```diff
@@ -115,28 +116,36 @@ func readFileContent(src string) ([]byte, error) {
 }
 
 type dirWatcher struct {
-	dirs sets.Set[string]
-	wg   sync.WaitGroup
-	w    *fsnotify.Watcher
+	// srcToTarget: source dir → target dir; empty target means reload-only, no copy.
+	srcToTarget map[string]string
```
vrutkovs: Lets move this to a separate `SrcTargetPairs`?

cmd/config-reloader/file_watch.go
```diff
@@ -115,28 +116,36 @@ func readFileContent(src string) ([]byte, error) {
...
+	wg          sync.WaitGroup
+	w           *fsnotify.Watcher
 }
 
-func newDirWatchers(dirs []string) (*dirWatcher, error) {
-	dws := sets.New[string]()
+// newDirWatchers watches dirs; targetDirs is paired by index with dirs (dirs[i] → targetDirs[i]).
+// If targetDirs is non-empty it must have the same length as dirs.
+// A watched dir with no corresponding target only triggers a reload on change.
+func newDirWatchers(dirs []string, targetDirs []string) (*dirWatcher, error) {
```
vrutkovs, line 185: Use `map[string]string` (or better `SrcTargetPairs`) here? I think that would allow us to avoid checks like `len(targetDirs) != len(dirs)`

cmd/config-reloader/file_watch.go
```diff
@@ -235,17 +249,88 @@ func (dw *dirWatcher) start(ctx context.Context, updates chan struct{}) {
...
-
 				}
 			}
 		}
 	}()
 }
 
+// copyDirToTarget reads files from srcDir, decompresses any gzip content, writes to targetDir,
+// and removes any files in targetDir that no longer exist in srcDir.
+func copyDirToTarget(srcDir, targetDir string) error {
```
vrutkovs: Since we're removing files from targetDir its more of a `sync` than `copy`

internal/controller/operator/factory/vmscrapes/extra_config.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// Shared constants for overflow scrape config Secrets (VMAgent and VMSingle).
+const (
+	// ExtraConfigFilename is the key in each overflow Secret holding the gzip-compressed job list.
+	ExtraConfigFilename = "jobs.yaml"
+	// ExtraConfigLabelKey is set on overflow Secrets for label-selector-based cleanup.
+	ExtraConfigLabelKey = "operator.victoriametrics.com/scrape-config-extra"
+	// MaxCompressedBucketSize is 90% of the Kubernetes Secret data size limit.
+	MaxCompressedBucketSize = corev1.MaxSecretSize * 9 / 10
```
vrutkovs: This assumes metadata takes <10% of the data, but it may easily be overflown (i.e. large `last-applied` annotation).

I think the approach should be recursive here:
* Build one large secret in this function with argument `parts=1`
* Generate necessary config (i.e. in vmagent path)
* Check the generated config size (including existing metadata if available). If its smaller than `corev1.MaxSecretSize` we're happy
* If we're smaller we call a function to call function with `parts=2`, create two secrets, check their size - and keep dividing the larger secrets

Hopefully this won't be too CPU intensive since we'll be running more compressions here. Maybe we can build metadata + file overheard and use it instead of 90% threshold? Since we're doing splitting perhaps we could omit the compression entirely?

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -1267,18 +1315,35 @@ func createOrUpdateScrapeConfig(ctx context.Context, rclient client.Client, cr,
...
+	// Assemble main config: global section + optional scrape_config_files glob + first bucket's jobs.
+	mainCfg := cfgBase
+	if extraCount > 0 {
+		mainCfg = append(mainCfg, yaml.MapItem{
+			Key:   "scrape_config_files",
+			Value: []string{vmscrapes.ExtraConfigFilesGlob},
+		})
+	}
+	mainCfg = append(mainCfg, yaml.MapItem{Key: "scrape_configs", Value: buckets[0]})
+	generatedConfig, err := yaml.Marshal(mainCfg)
```
vrutkovs, line 1345: We should be checking its size here

internal/controller/operator/factory/vmscrapes/extra_config.go
AndrewChubatiuk: even with metadata i would keep 90% threshold to not cause issues for kyverno users

internal/controller/operator/factory/vmagent/vmagent.go
AndrewChubatiuk: if metadata size is taken into accounts for buckets calculation checking size here is redundant

internal/controller/operator/factory/vmscrapes/extra_config.go
AndrewChubatiuk: updated implementation

cmd/config-reloader/file_watch.go
vrutkovs: We should be passing a list of `dirPair`'s here. This would eliminate the len check

cmd/config-reloader/file_watch.go
```diff
@@ -235,17 +253,90 @@ func (dw *dirWatcher) start(ctx context.Context, updates chan struct{}) {
...
-
 				}
 			}
 		}
 	}()
 }
 
+// syncDirToTarget reads files from srcDir, decompresses any gzip content, writes to targetDir,
+// and removes any files in targetDir that no longer exist in srcDir.
+func syncDirToTarget(srcDir, targetDir string) error {
```
vrutkovs: Pass `dirPair`?

cmd/config-reloader/file_watch.go
```diff
@@ -235,17 +253,90 @@ func (dw *dirWatcher) start(ctx context.Context, updates chan struct{}) {
...
+	if err := os.MkdirAll(targetDir, 0755); err != nil {
+		return fmt.Errorf("cannot create target dir %s: %w", targetDir, err)
+	}
+	srcEntries, err := os.ReadDir(srcDir)
+	if err != nil {
+		return fmt.Errorf("cannot read source dir %s: %w", srcDir, err)
```
vrutkovs: No point in creating a targetDir if source dir is inaccessible - this check should be earlier

cmd/config-reloader/file_watch.go
vrutkovs: Also this is not atomic for multiple files - we write them one by one, instead we should be creating a tmp dir, write files there - and rename it from tmp to `targetDir` atomically. This would also avoid the part where we cleanup targetDir from the files no longer required there

cmd/config-reloader/file_watch.go
```diff
@@ -235,17 +253,90 @@ func (dw *dirWatcher) start(ctx context.Context, updates chan struct{}) {
...
+	}
+	srcNames := make(map[string]struct{}, len(srcEntries))
+	for _, e := range srcEntries {
+		if e.IsDir() || strings.HasPrefix(e.Name(), "..") {
```
vrutkovs: Not sure why we need `HasPrefix` - we would already skip k8s projected dirs in `IsDir` check

cmd/config-reloader/file_watch.go
```diff
@@ -235,17 +253,90 @@ func (dw *dirWatcher) start(ctx context.Context, updates chan struct{}) {
...
+			continue
+		}
+		srcNames[e.Name()] = struct{}{}
```
vrutkovs: Lets use k8s sets here

cmd/config-reloader/main.go
```diff
@@ -46,6 +44,8 @@ var (
 		"watched-dir", "directory to watch non-recursively")
 	rulesDir = flagutil.NewArrayString(
 		"rules-dir", "the same as watched-dir, legacy")
+	targetDir = flagutil.NewArrayString(
+		"target-dir", "when provided, must have the same count as --watched-dir; each non-empty value causes files from the corresponding watched directory to be decompressed (if needed) and written to that target directory on change")
```
vrutkovs:
```suggestion
		"target-dir", "when provided, must have the same number of argument as --watched-dir; each non-empty value causes files from the corresponding watched directory to be decompressed (if needed) and written to that target directory on change")
```

internal/controller/operator/factory/vmsingle/vmsingle.go
```diff
@@ -667,15 +712,15 @@ func CreateOrUpdateScrapeConfig(ctx context.Context, rclient client.Client, cr *
...
 	ac := getAssetsCache(ctx, rclient, cr)
-	if err := createOrUpdateScrapeConfig(ctx, rclient, cr, prevCR, childObject, ac); err != nil {
+	if _, err := createOrUpdateScrapeConfig(ctx, rclient, cr, prevCR, childObject, ac); err != nil {
 		return err
 	}
 	return nil
 }
 
-func createOrUpdateScrapeConfig(ctx context.Context, rclient client.Client, cr, prevCR *vmv1beta1.VMSingle, childObject client.Object, ac *build.AssetsCache) error {
+func createOrUpdateScrapeConfig(ctx context.Context, rclient client.Client, cr, prevCR *vmv1beta1.VMSingle, childObject client.Object, ac *build.AssetsCache) (int, error) {
```
vrutkovs, line 722: We should unify this with vmagent's `createOrUpdateScrapeConfig`

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -174,7 +175,7 @@ func CreateOrUpdate(ctx context.Context, cr *vmv1beta1.VMAgent, rclient client.C
 
 	if prevCR != nil {
 		var err error
-		prevAppTpl, err = newK8sApp(prevCR, ac)
+		prevAppTpl, err = newK8sApp(prevCR, ac, 0)
```
vrutkovs, line 178: We always generate a 0 overflow for previous template - wouldn't that cause a reload every time we sync? Perhaps we should be storing `extraCount` in status to prevent that?

internal/controller/operator/factory/vmagent/vmagent.go
vrutkovs: Do we have a mechanism for a manual override of the bucketing - in case we didn't take all factors in account and it still overflows?

internal/controller/operator/factory/vmsingle/vmsingle.go
AndrewChubatiuk: it's used at VMAgent's newPodSpec to generate flags properly, need first to simplify secrets loading for remote write, then we can make this function shared for both

cmd/config-reloader/file_watch.go
AndrewChubatiuk: `..data` is a symlink, not a directory — IsDir() returns false for it. Without the prefix check, `os.ReadFile("..data")` would follow the symlink to the timestamped directory and fail with EISDIR on every Kubernetes secret mount.

internal/controller/operator/factory/vmagent/vmagent.go
AndrewChubatiuk: added global variable VM_CONFIG_DATA_BUDGET_BYTES to override default limit

internal/controller/operator/factory/vmsingle/vmsingle.go
```diff
@@ -125,22 +126,22 @@ func CreateOrUpdate(ctx context.Context, cr *vmv1beta1.VMSingle, rclient client.
 	var prevDeploy *appsv1.Deployment
 	if prevCR != nil {
 		var err error
-		prevDeploy, err = newDeploy(ctx, prevCR)
+		prevDeploy, err = newDeploy(ctx, prevCR, 0)
```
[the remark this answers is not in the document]
AndrewChubatiuk: prevDeploy with extraConfigSecretCount=0 has no real effect. reconcile.Deployment only reads prevObj for ObjectMeta and pod template annotations — both are derived from the CR, not from the overflow count. The actual spec diff that decides whether to roll out is against the live cluster object, not against prevDeploy.

internal/controller/operator/factory/vmagent/vmagent.go
AndrewChubatiuk: only prevAppTpl metadata is used in reconcile, this makes no impact

vrutkovs: So far its only used in `spec` but I worry that it would leak into metadata (or just spec metadata) later. I think we'll guard ourselves from future issue by adding a single status field.

I agree its a minor thing though, no restarts happening now

cmd/config-reloader/file_watch.go
```diff
@@ -210,10 +276,13 @@ func (dw *dirWatcher) start(ctx context.Context, updates chan struct{}) {
...
 		return true, nil
 	}
-	for dir := range dw.dirs {
-		if _, err := updateCache(dir); err != nil {
+	for _, p := range dw.pairs {
+		if _, err := updateCache(p.src); err != nil {
 			logger.Errorf("cannot update dir cache during start: %s", err)
 		}
+		if err := p.sync(); err != nil {
+			logger.Errorf("cannot copy dir %s to target on start: %s", p.src, err)
```
vrutkovs, line 284: We should exit here, otherwise we'll be sending `update` signal event when update / sync fail

AndrewChubatiuk: fixed

vrutkovs (approved):

Issue #1681, opened 2025-12-09
vmalert restarts when applying ~1400 alerts via VMRule resources


patsevanton:
### Describe the bug

When applying approximately 1400 alerting rules (distributed across multiple VMRule CRDs, each containing 200 alerts), vmalert pod restarts. The restart occurs around the 7th VMRule file (containing alerts 1201–1400). This behavior prevents the reliable creation of large-scale alerting rule sets in Kubernetes environments using the VictoriaMetrics Operator.

### To Reproduce

Deploy VictoriaMetrics stack with operator (e.g., using victoria-metrics-k8s-stack Helm chart).

Use the performance test repository: https://github.com/patsevanton/performance-test-alerts-victoriametrics

Apply VMRule files sequentially (each with 200 alerts).

Observe vmalert pod restarting around the 7th file (after ~1200-1400 alerts).

Example VMRule structure (each file contains 200 similar alerts):
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMRule
metadata:
  name: loadtest-alerts-1
spec:
  groups:
    - name: loadtest-generated-1
      interval: 30s
      rules:
        - alert: LoadTestAlert_1_1
          expr: vector(1)
          for: 15s
          labels:
            severity: warning
          annotations:
            summary: Load Test Alert 1-1
        # ... 199 more alerts
```

### Version

```
helm list -n vmks
NAME	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART                            	APP VERSION
vmks	vmks     	1       	2025-12-09 19:57:07.582483153 +0600 +06	deployed	victoria-metrics-k8s-stack-0.65.1	v1.131.0 
```

### Logs

From vmalert pod events:
```
0s          Normal    Created                  pod/vmalert-vmks-victoria-metrics-k8s-stack-dd75cf485-77f86      Created container config-reloader
0s          Normal    Started                  pod/vmalert-vmks-victoria-metrics-k8s-stack-dd75cf485-77f86      Started container config-reloader
0s          Normal    ScalingReplicaSet        deployment/vmalert-vmks-victoria-metrics-k8s-stack               Scaled down replica set vmalert-vmks-victoria-metrics-k8s-stack-995cd7594 to 0 from 1
[...]

PR #2235, opened 2026-06-01
vmalert: compress rules configmaps to minimize amount of volumes and decrease probability of pod restart

AndrewChubatiuk:
related issue https://github.com/VictoriaMetrics/operator/issues/1681


AndrewChubatiuk:
@cubic-ai-dev

internal/controller/operator/factory/vmalert/vmalert.go
```diff
@@ -306,11 +321,40 @@ func newPodSpec(cr *vmv1beta1.VMAlert, ruleConfigMapNames []string, ac *build.As
...
 	if !cr.IsUnmanaged() {
 		crc := build.ConfigReloaderContainer(false, cr, crMounts, nil)
+		// rules-out write-side: not in crMounts to avoid --watched-dir causing reload loops.
+		crc.VolumeMounts = append(crc.VolumeMounts, corev1.VolumeMount{
+			Name:      "rules-out",
+			MountPath: vmAlertRulesOutDir,
+		})
+		for i := range ruleConfigMapNames {
+			crc.Args = append(crc.Args,
+				fmt.Sprintf("--target-dir=%s", path.Join(vmAlertRulesOutDir, fmt.Sprintf("rules-src-%d", i))),
```
vrutkovs: I think we need an interface around adding watch / target dirs instead of assembling cmdline on the fly - we need to make sure `crMounts` don't include this dir already

internal/controller/operator/factory/build/child_object.go
```diff
@@ -170,3 +171,38 @@ func (co *ChildObjects[T]) forEachCollectSkipOn(apply func(s T) error, shouldIgn
...
+	data, err := yaml.Marshal(items)
+	if err != nil {
+		return nil, fmt.Errorf("yaml marshal: %w", err)
+	}
+	compressed, err := GzipConfig(data)
+	if err != nil {
+		return nil, fmt.Errorf("gzip: %w", err)
+	}
+	n := len(compressed)
+	if n <= limit || len(items) == 1 {
```
vrutkovs:
```suggestion
	if n <= limit {
```

Don't attempt to create a single over the limit item

vrutkovs: We should be returning error here for recursion to start breaking this item down iiuc?

vrutkovs (changes requested):

vrutkovs (approved):

vrutkovs (approved):

PR #2242, opened 2026-06-03
vmcluster: support storage node discovery

### LINKED ISSUE #2236: VMCluster: support SRV-based storage node discovery for vminsert/vmselect
## Summary

`vminsert` and `vmselect` both accept `-storageNode=dnssrv+<srv-record>` (with `-storageNode.discoveryInterval`) which lets the binary dynamically discover storage pods via DNS SRV records and refresh on an interval. The operator currently generates a **static** `-storageNode=<podname>` list per replica from `AvailableStorageNodeIDs`, which means scaling `vmstorage.replicaCount` requires a full vminsert/vmselect rolling restart before the new storage pods receive any traffic.

Request: expose an opt-in SRV-discovery mode in the VMCluster CRD that:

1. Skips generation of the static `-storageNode=<pod>...` list
2. Emits `-storageNode=dnssrv+_<port-name>._tcp.<vmstorage-svc>.<ns>.svc.cluster.local` (and the same for vmselect→vmstorage)
3. Sets `-storageNode.discoveryInterval` (configurable, default e.g. `30s`)

## Why

We hit this today on a 16 → 20 vmstorage scale-up:

- New vmstorage pods (16-19) came Ready promptly, but vminsert kept routing only to 0-15 because its `-storageNode=` argv was frozen from process startup.
- Since the operator's vminsert reconcile is serialized behind the vmstorage rolling update, vminsert didn't get a new template until **~2.5h** after the spec change.
- During that window, load that should have spread across 20 storage pods was concentrated on 15 (one was in mid-restart, four were unreachable from vminsert).
- vminsert saturated at `vm_concurrent_insert_current = 4000/4000` on every pod, Alloy backpressure peaked at ~580M pending samples.

Workaround was to manually patch the vminsert Deployment with the new `-storageNode=` list and bump replicas, forcing a rolling restart of vminsert ahead of the operator. SRV discovery would avoid both the manual intervention and the restart entirely — vmstorage scale-up becomes zero-downtime for the write path.

## API sketch

```yaml
spec:
  vmstorage:
    replicaCount: 20
  vminsert:
    replicaCount: 10
    storageNodeDiscovery:           # <-- new
      mode: SRV                     # SRV | Static (default Static, current behavior)
      interval: 30s                 # maps to -storageNode.discoveryInterval
  vmselect:
    storageNodeDiscovery:
      mode: SRV
      interval: 30s
```

When `mode: SRV`, the operator:
- Skips the existing per-replica `-storageNode=<pod>` generation
- Emits a single `-storageNode=dnssrv+_<port-name>._tcp.<vmstorage-svc>.<ns>.svc.cluster.local`
- Emits `-storageNode.discoveryInterval=<interval>`

The vmstorage headless Service already exposes named ports (`vminsert: 8400`, `vmselect: 8401`) — the existing SRV records resolve all pods, no Service changes needed.

## Compatibility

- Default `mode: Static` preserves today's behavior. Existing clusters are unchanged unless they opt in.
- Switching `Static → SRV` triggers a one-time rolling restart of vminsert/vmselect (template change). After that, future scale-ups of vmstorage are zero-downtime.

Happy to take a stab at the PR if maintainers are aligned on the API shape.

AndrewChubatiuk:
fixes #2236

---
## Summary by cubic
Adds automatic vmstorage discovery to VMCluster. vminsert and vmselect can now resolve storage nodes via DNS SRV with optional refresh interval and address filter to avoid restarts when storage scales. Requires an enterprise license.

- **New Features**
  - New `spec.discovery` with `enabled`, `interval`, and `filter`; per-component overrides via `spec.vminsert.discovery` and `spec.vmselect.discovery`.
  - Controller sets `-storageNode=srv+<headless-svc>.<namespace>[.svc.<domain>]:<port>` and adds `-storageNode.discoveryInterval`/`-storageNode.filter` when provided; otherwise uses static pod addresses.
  - Validation: license required when discovery is enabled (global or per-component); validates `interval` duration and `filter` regexp; disallows `vmstorage.maintenanceInsertNodeIDs`/`maintenanceSelectNodeIDs` when the corresponding component uses discovery (allowed if that component overrides discovery to disabled).
  - CRD, API docs, changelog, and `vmcluster` docs updated with examples and guidance on filter matching against full pod addresses.

- **Migration**
  - No change by default.
  - To enable, set `spec.discovery.enabled: true` (or per component) and optionally set `interval`/`filter`.
  - Ensure a valid license is configured before enabling.

<sup>Written for commit 896c750de3c2fe68f4273592c720de112c963812. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VictoriaMetrics/operator/pull/2242?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


api/operator/v1beta1/vmcluster_types.go
```diff
@@ -86,6 +87,40 @@ type VMClusterSpec struct {
...
+type VMClusterDiscovery struct {
+	// Enabled turns on automatic vmstorage node discovery via DNS SRV records.
+	// This is an enterprise feature and requires a valid license key.
+	// See https://docs.victoriametrics.com/victoriametrics/cluster-victoriametrics/#automatic-vmstorage-discovery
+	Enabled bool `json:"enabled"`
+	// Interval is the interval for refreshing the vmstorage node list resolved from DNS SRV records.
+	// The minimum supported value is 1s.
+	// Defaults to 2s if not set.
+	// +optional
+	// +kubebuilder:validation:Pattern:="^[0-9]+(ms|s|m|h)$"
```
vrutkovs: Milliseconds is just unrealistic. IMO seconds are also useless here

docs/resources/vmcluster.md
```diff
@@ -630,6 +629,93 @@ Possible configuration options for backup crd can be found at [link](https://doc
...
+
+The `discovery` field can be set at the cluster level (applies to both `vminsert` and `vmselect`) or overridden per component.
+
+**`spec.discovery` fields:**
+
+| Field      | Description |
+|------------|-------------|
+| `enabled`  | Enables automatic vmstorage node discovery via DNS SRV records. |
+| `interval` | How often to refresh the list of storage nodes. Minimum `1s`, defaults to `2s`. |
+| `filter`   | Optional regexp to filter discovered storage addresses. Only matching addresses are used. |
```
vrutkovs, line 648: Is it a filter on IP addresses or hostnames?

api/operator/v1beta1/vmcluster_types.go
```diff
@@ -726,6 +769,39 @@ func (cr *VMCluster) Validate() error {
...
 	}
+	var vminsertDiscovery, vmselectDiscovery *VMClusterDiscovery
+	if cr.Spec.VMInsert != nil {
+		vminsertDiscovery = cr.Spec.VMInsert.Discovery.OrDefault(cr.Spec.Discovery)
+	}
+	if cr.Spec.VMSelect != nil {
+		vmselectDiscovery = cr.Spec.VMSelect.Discovery.OrDefault(cr.Spec.Discovery)
+	}
+	if d := vminsertDiscovery; d != nil && d.Enabled {
+		if !cr.Spec.License.IsProvided() {
```
vrutkovs, line 799: We duplicate license check here and later in `vmselectDiscovery` - and  in VMAnomaly. Lets use a single error message for all cases?

api/operator/v1beta1/vmcluster_types.go
```diff
@@ -726,6 +769,39 @@ func (cr *VMCluster) Validate() error {
...
+			return fmt.Errorf("discovery requires a valid license key, see https://docs.victoriametrics.com/victoriametrics/enterprise/")
+		}
+		if d.Filter != "" {
```
vrutkovs: Should we validate the interval here too?

docs/resources/vmcluster.md
AndrewChubatiuk: it's a filter for hostnames

api/operator/v1beta1/vmcluster_types.go
AndrewChubatiuk: will keep seconds

api/operator/v1beta1/vmcluster_types.go
AndrewChubatiuk: added validation

api/operator/v1beta1/vmcluster_types.go
AndrewChubatiuk: done

vrutkovs (approved):

PR #2251, opened 2026-06-05
internal/controller/vmanomaly: sync configuration with latest vmanomaly releases

zekker6:
- add support of "infer_every" and "exact" parameters - https://docs.victoriametrics.com/anomaly-detection/components/scheduler/#configuration-parameters
- add support of "connection_retry_attempts" - https://docs.victoriametrics.com/anomaly-detection/components/writer/#config-parameters
- add support of "use_reader_connection_settings" parameter - https://docs.victoriametrics.com/anomaly-detection/components/server/#parameters
- add support of "logger_levels" parameter  - https://docs.victoriametrics.com/anomaly-detection/components/settings/#logger-levels
- add support of "offset" parameter - https://docs.victoriametrics.com/anomaly-detection/components/reader/#per-query-parameters

- fix "global_smoothing" parameter marshaling to yaml - https://docs.victoriametrics.com/anomaly-detection/components/models/#online-seasonal-quantile


api/operator/v1/vmanomaly_types.go
```diff
@@ -275,6 +282,10 @@ type VMAnomalyServerSpec struct {
 	// UIDefaultState defines default query state for anomaly UI
 	// +optional
 	UIDefaultState string `json:"uiDefaultState,omitempty" yaml:"ui_default_state,omitempty"`
+	// UseReaderConnectionSettings when set to true, anomaly UI reuses connection settings
+	// (credentials, TLS, etc.) from the reader configuration to connect to datasources
+	// +optional
+	UseReaderConnectionSettings bool `json:"useReaderConnectionSettings,omitempty" yaml:"use_reader_connection_settings,omitempty"`
```
vrutkovs, line 288: That would default it to explicit `false`, do we want to distinguish unset / true / false here? In that case it's best be done via `*bool`

AndrewChubatiuk (approved):

api/operator/v1/vmanomaly_types.go
fred-navruzov: for now, we can't think of real need to distinguish between false/unset for our boolean args, so it should be safe to use current explicit false behavior, @zekker6 , wdyt?

zekker6: It's an implicit false for vmanomaly itself, there is no distinction between unset / false. So I thinkg that's fine to keep as a `bool`

internal/controller/operator/factory/vmanomaly/config/config_test.go
```diff
@@ -367,11 +376,16 @@ models:
...
     z_threshold: 2.5
 schedulers:
-  scheduler_1m:
-    class: scheduler.periodic.PeriodicScheduler
-    fit_every: 2m
+  scheduler_backtesting:
+    class: backtesting
     fit_window: 3h
-    infer_every: 1m
+    from_iso: 0001-01-01T00:00:00Z
```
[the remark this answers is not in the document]
zekker6: Seems to be worth fixing, but I'll address this in a separate PR as the source of this is not related to current set of changes.

vrutkovs (approved):

PR #2253, opened 2026-06-06
fix: allow watched files with '..' in their names

immanuwell:
config-reloader had a sneaky edge case.

It skipped any watched path containing `..`, so a normal file like `rules..yaml` got treated like kube hidden entries and no reload fired. This is real world stuff too, ConfigMap keys may contain `.` so that filename is valid.

This patch only skips basename entries that start with `..`, like `..data`, and keeps regular files working.

Repro
1. On current `master`, run `go test ./cmd/config-reloader -run TestDirWatcherProcessesRegularFilesWithDoubleDotsInName -count=1` and it fails.
2. Watch a directory with a regular file named `rules..yaml`, edit it, reload is not triggered.

Checks
- `go test ./cmd/config-reloader -count=1`
- `go test $(go list ./... | grep -v /e2e)`
- `make lint`

---
## Summary by cubic
Fixes `config-reloader` so files with '..' in their names (e.g., rules..yaml) trigger reloads. Only Kubernetes synthetic entries whose basename starts with '..' (e.g., ..data) are ignored.

- **Bug Fixes**
  - Skip only paths where `filepath.Base(path)` starts with `..`; do not skip regular files containing `..`.
  - Add tests to confirm reloads for `rules..yaml` and no reloads for `..data`.

<sup>Written for commit 44f2ecdea6cc4e8414819d0e852329dfe9551582. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VictoriaMetrics/operator/pull/2253?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


vrutkovs (approved):
LGTM, thank you for your contribution. Could you also add a CHANGELOG entry?

PR #2258, opened 2026-06-08
fix: add listMapKey to VMRule rules for structured server-side apply merge

### LINKED ISSUE #657: Add patchMergeKey to VMRule CRDs

mvanhorn:
## Summary

Adds a `+listMapKey=alert` marker to the `VMRule` `rules` array so Kubernetes server-side apply merges rule entries by key instead of replacing the whole list, preventing rules from clobbering each other on partial applies.

## Why this matters

`VMRule.spec.groups[].rules` was a plain list with no `listMapKey`, so a server-side apply that touched one rule replaced the entire `rules` array, dropping rules managed by other appliers (#657). Marking the list with its identifying key lets the API server do a structured merge. The change is in `api/operator/v1beta1/vmrule_types.go` with the regenerated CRD in `config/crd/overlay/crd.yaml`, plus a unit test and `docs/api.md` / `docs/CHANGELOG.md` updates.

## Testing

- Added a unit test asserting the merge-key marker is present on the rules field.
- `go build` and `go vet` pass on the `api` module; the CRD overlay regenerates cleanly.

Closes #657

---
## Summary by cubic
Make `VMRule.spec.groups[].rules` a map-style list keyed by `record` and `alert` so server-side apply merges rules by key instead of replacing the list. Adds XOR validation and defaults for `record`/`alert` to prevent invalid specs and rule clobbering (fixes #657).

- **Bug Fixes**
  - Marked `RuleGroup.rules` with `+listType=map`, `+listMapKey=record`, `+listMapKey=alert`; set `Rule.record`/`Rule.alert` defaults to `""`; regenerated CRDs with `x-kubernetes-list-type`, `x-kubernetes-list-map-keys`, and defaults; updated `docs/api.md` and `docs/CHANGELOG.md`.
  - Enforced validation: error if both `record` and `alert` are set, or neither.
  - Added tests for CRD schema markers and JSON field names, JSON round-trip/validation, composite-key SSA merges for rules, and traditional SMP merge-by-name behavior for groups.

<sup>Written for commit 7d0ebe1d35633c642bcfe3ed021dc580ebef9ab1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/VictoriaMetrics/operator/pull/2258?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/api/operator/v1beta1/vmrule_types.go
+++ b/api/operator/v1beta1/vmrule_types.go
@@ -101,9 +101,11 @@ type RuleGroup struct {
 type Rule struct {
 	// Record represents a query, that will be recorded to dataSource
 	// +optional
+	// +kubebuilder:default=""
 	Record string `json:"record,omitempty" yaml:"record,omitempty"`
 	// Alert is a name for alert
 	// +optional
+	// +kubebuilder:default=""
 	Alert string `json:"alert,omitempty" yaml:"alert,omitempty"`
 	// Expr is query, that will be evaluated at dataSource
 	// +optional
@@ -187,6 +189,14 @@ func (cr *VMRule) Validate() error {
 			return fmt.Errorf("duplicate group name: %s", errContext)
 		}
 		uniqNames.Insert(group.Name)
+		for j, rule := range group.Rules {
+			if rule.Record != "" && rule.Alert != "" {
+				return fmt.Errorf("rule at group %s index %d has both record and alert set", group.Name, j)
+			}
+			if rule.Record == "" && rule.Alert == "" {
+				return fmt.Errorf("rule at group %s index %d has neither record nor alert set", group.Name, j)
+			}
+		}
 		groupBytes, err := yaml.Marshal(group)
 		if err != nil {
 			return fmt.Errorf("cannot marshal %s, err: %w", errContext, err)
--- a/api/operator/v1beta1/vmrule_types.go
+++ b/api/operator/v1beta1/vmrule_types.go
@@ -36,7 +36,11 @@ type RuleGroup struct {
 	// evaluation interval for group
 	// +optional
 	Interval string `json:"interval,omitempty" yaml:"interval,omitempty"`
-	// Rules list of alert rules
+	// Rules list of alert rules.
+	// Rules are merged by record and alert fields.
+	// +listType=map
+	// +listMapKey=record
+	// +listMapKey=alert
 	Rules []Rule `json:"rules"`
 	// Limit the number of alerts an alerting rule and series a recording
 	// rule can produce
--- a/api/operator/v1beta1/vmrule_types_test.go
+++ b/api/operator/v1beta1/vmrule_types_test.go
@@ -6,16 +6,17 @@ import (
 	"errors"
 	"io"
 	"os"
+	"reflect"
 	"strings"
 	"testing"
-
-	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v2"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	strategicpatch "k8s.io/apimachinery/pkg/util/strategicpatch"
+	yamlutil "k8s.io/apimachinery/pkg/util/yaml"
 )
 
 func TestVMRuleValidate(t *testing.T) {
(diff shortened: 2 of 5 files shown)


AndrewChubatiuk:
hey @mvanhorn 
thanks for PR
could you clarify a purpose of added tests cases? they seem to be irrelevant

api/operator/v1beta1/vmrule_types.go
```diff
@@ -41,7 +41,11 @@ type RuleGroup struct {
 	// evaluation interval for group
 	// +optional
 	Interval string `json:"interval,omitempty" yaml:"interval,omitempty"`
-	// Rules list of alert rules
+	// Rules list of alert rules.
+	// Rules are merged by record and alert fields.
+	// +listType=map
```
vrutkovs, line 41: I think this would allow us to create entries which are both record rules and alerts - but that would be rejected by vmalert.

We should trap such cases in `Validate()`:
```
for j, rule := range group.Rules {
    if rule.Record != "" && rule.Alert != "" {
        return fmt.Errorf("rule at group %s index %d has both record and alert set", group.Name, j)
    }
    if rule.Record == "" && rule.Alert == "" {
        return fmt.Errorf("rule at group %s index %d has neither record nor alert set", group.Name, j)
    }
}
```

api/operator/v1beta1/vmrule_types_test.go
```diff
@@ -183,3 +194,158 @@ spec:
             description: "Service nginx on env test accepted {{$labels.requests}} requests in the last 5 minutes"`,
 	})
 }
+
+func TestVMRuleRulesRoundTripAndValidate(t *testing.T) {
```
vrutkovs, line 231: These are useful unit tests for existing behaviour, but we should add some more for `listMapKey` behaviour, i.e. "Two rules with same alert name but different expr"

vrutkovs (changes requested):

api/operator/v1beta1/vmrule_types.go
AndrewChubatiuk: one of keys is allowed to be empty, there's another issue: vmalert allows to have rules with identical record or alert property and with this change such requirement appears on operator

AndrewChubatiuk (approved):

api/operator/v1beta1/vmrule_types_test.go
AndrewChubatiuk: added some cases

vrutkovs (approved):

mvanhorn:
Thanks for landing this, @vrutkovs. Adding the listMapKey marker to the VMRule rules array makes server-side apply merge correctly.

mvanhorn:
Thank you @vrutkovs. The listMapKey marker on VMRule rules should stop server-side apply from clobbering the whole list.

AndrewChubatiuk:
@mvanhorn
this PR was reverted in 0.73.1 as it breaks k8s-stack installation, that has many default rules with duplicated names

mvanhorn:
Thanks for the heads up, and sorry for the breakage. I dug into whether a composite list-map key could make this safe and concluded it cannot, so I will not attempt a re-fix.

The core problem: Kubernetes list-map keys must be direct scalar fields with a unique composite, but VMAlert's actual rule identity (HashRule) includes the full labels map. Two rules can share the same alert name and expr and differ only in labels like severity, which is exactly the k8s-stack pattern that broke. Labels cannot be a list-map key, and making mutable fields like expr part of identity turns edits into delete/add and can append a duplicate when a chart changes an expression. A validating webhook could only reject those valid rule sets, not merge them.

So keeping rules atomic in v1beta1 is the right call. If per-rule patching ever matters enough, it would need a new API version with a required stable rule id and a migration story. Apologies again for the churn, and thanks for the quick revert.

PR #2264, opened 2026-06-09
CI workflow speedup

vrutkovs:
Some more tricks to cache more things.

Upgrade workflow run: https://github.com/VictoriaMetrics/operator/actions/runs/27329419701


AndrewChubatiuk:
why have you removed cleanup step? understand it can take some time, but we've reached machine disk size limit already

vrutkovs:
Oh, okay, I though its some artifact from previous times

AndrewChubatiuk:
can you make a change to trigger updated pipelines?

Makefile
```diff
@@ -169,7 +169,7 @@ test-e2e: load-kind ginkgo crust-gather mirrord
 		--output-interceptor-mode=none \
 		-procs=$(E2E_TESTS_CONCURRENCY) \
 		-randomize-all \
-		-timeout=60m \
+		-timeout=90m \
```
AndrewChubatiuk: why not just switch to vm-runner?

vrutkovs: right, lets do that

AndrewChubatiuk (approved):

PR #2268, opened 2026-06-10
.github: add explicit persist-credentials for checkout action

rudransh-shrivastava:
Add explicit `persist-credentials` as it's `true` by default.

cc: @arkid15r


vrutkovs:
A good explanation at why we need that - https://github.com/actions/checkout/issues/485

vrutkovs (approved):

PR #2302, opened 2026-06-17
test: run "Build & Test" on all changes

vrutkovs:
This test is now required, so we should be running it on every change

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -4,16 +4,6 @@ on:
   push:
     branches:
       - master
-    paths:
-      - '**'
-      - '!docs/**'
-      - '!*.md'
-      - '!.github/**'
-      - '.github/workflows/main.yaml'
-      - '.github/workflows/docs.yaml'
-      - '!codespell/**'
-      - 'LICENSE'
-      - 'PROJECT'
   pull_request:
     branches:
       - master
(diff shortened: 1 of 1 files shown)


AndrewChubatiuk:
@vrutkovs why do we need to run tests on markdown changes?
CI probably should run for generated files only, but they can be added to gitignore instead

vrutkovs:
This test is now required to pass so that we could queue PRs (I asked Roman to set this in the repo changes). Its relatively cheap - 3-4 mins so worth having, as it may prevent us from merging bad PRs or force pushes (I'm famous for doing that). I wish we could also add "wait for e2e to complete if its running", but I don't think Github has that feature.

I don't mind reverting that if that prevents us from merging useful changes.

AndrewChubatiuk (dismissed):

AndrewChubatiuk:
if we're removing this filters, they should be moved to e2e part as it now only skips pipeline for root Makefile, but now e2e runs for all files except root Makefile

vrutkovs:
ah, yes, good idea

.github/workflows/main.yaml
```diff
@@ -98,6 +77,10 @@ jobs:
             src:
               - '**'
               - '!Makefile'
+              - '!docs/**'
+              - '!*.md'
+              - '!.github/**'
+              - '!codespell/**'
```
AndrewChubatiuk, line 84: all workflows except .github/workflows/main.yaml can also be excluded

AndrewChubatiuk (approved):

PR #2307, opened 2026-06-18
fix: use tcp startup probes for tls-backed workloads

### LINKED ISSUE #1824: Add TCP startup checks

immanuwell:
Fixes #1824

When a managed app enables TLS on its HTTP endpoint, the operator still renders default `httpGet` app probes. kubelet hits that endpoint without a TLS handshake, so health checks can fail even when the pod is fine.

This keeps plain HTTP behavior as-is, and switches the default app probe to a `tcpSocket` `startupProbe` only for TLS-backed endpoints when no custom probes are set. Custom probes still win, no funny business.

Repro:
1. Create a `VMAuth` with `spec.extraArgs.tls: "true"` and no custom probes.
2. Check the rendered pod spec. It gets default `httpGet` probes on the app port.
3. Run it with TLS or mTLS. kubelet probe requests fail, pod looks sick for no good reason.

Checks:
- `go test ./internal/controller/operator/factory/...`
- `make test`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,7 @@ aliases:
 ## tip
 
 * BUGFIX: [config-reloader](https://docs.victoriametrics.com/operator/): fix possible panic on Secret watch events when the informer's local cache fell out of sync and Kubernetes delivered a stale tombstone entry instead of the Secret object. The config-reloader now unwraps tombstones correctly and logs an error for any other unexpected types.
+* BUGFIX: [vmoperator](https://docs.victoriametrics.com/operator/): switch default app probes to `tcpSocket` `startupProbe` when TLS is enabled on the managed HTTP endpoint. This avoids broken kubelet `httpGet` checks against TLS and mTLS-protected workloads. See [#1824](https://github.com/VictoriaMetrics/operator/issues/1824).
 
 * Dependency: [vmoperator](https://docs.victoriametrics.com/operator/): Updated default versions for VL apps to [v1.51.0](https://github.com/VictoriaMetrics/VictoriaLogs/releases/tag/v1.51.0).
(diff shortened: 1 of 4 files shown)


AndrewChubatiuk:
@immanuwell @vrutkovs 
why tcp liveness and readiness checks are not used for TLS endpoints by default?
why they are just disabled?

immanuwell:
@AndrewChubatiuk I didn't switch TLS endpoints to default TCP livenessProbe/readinessProbe because, per the discussion in https://github.com/VictoriaMetrics/operator/issues/1824 these probes only verify that the listener is accepting connections, not that the app is actually healthy or meaningfully ready

so the agreed direction was to use a TCP startupProbe for the default TLS/mTLS case, and not keep steady-state default liveness/readiness checks with weaker semantics

custom probes still override this, so users who want TCP liveness/readiness can set them explicitly

that's my understanding of the agreed direction, maybe I'm wrong

AndrewChubatiuk (approved):

AndrewChubatiuk:

vrutkovs:
I think this also needs an upgrade test?

PR #2309, opened 2026-06-19
fix: remove orphaned vmsingle feature configmaps

immanuwell:
bug:
if a `VMSingle` had `inlineRelabelConfig`, `relabelConfig`, or `streamAggrConfig` before, and later those fields are removed, the generated ConfigMaps just hang around. stale stuff, not referenced anymore.

fix:
clean orphaned relabel and stream aggr ConfigMaps during `VMSingle` reconcile.

repro:
1. Create a `VMSingle` with `spec.inlineRelabelConfig` or `spec.streamAggrConfig`.
2. Wait until operator creates the generated ConfigMap.
3. Remove those fields from the `VMSingle` spec.
4. Reconcile again. before this PR, the old ConfigMaps stay there.
5. With this PR, they get cleaned up.

tests:
`go test ./internal/controller/operator/factory/vmsingle/... ./internal/controller/operator/factory/finalize/... ./internal/controller/operator/factory/reconcile/...`


AndrewChubatiuk (approved):
Thanks for contribution!

PR #2311, opened 2026-06-22
docs: add docs about HPA and VPA

vrutkovs:
Add doc pages about operator integrations with HPA and VPA

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -453,16 +453,203 @@ It can be simplified with cert-manager and kustomize command:
 kustomize build config/deployments/webhook/
 ```
 
-## Requirements
+### Requirements
 
 - Valid certificate with key must be provided to operator
 - Valid CABundle must be added to the `ValidatingWebhookConfiguration`
 
 
-## Useful links
+### Useful links
 
 - [k8s admission webhooks](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/)
 
+## HPA 
+VictoriaMetrics operator can create and manage Kubernetes [HorizontalPodAutoscaler](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) objects for selected VictoriaMetrics components. Configure HPA in VictoriaMetrics custom resources instead of creating `HorizontalPodAutoscaler` objects manually - the operator needs to know that a workload is autoscaled so it can preserve autoscaler-managed replica counts during reconciliation.
+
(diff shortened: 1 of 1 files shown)


AndrewChubatiuk:
why is it a part of integrations? HPA is a part of k8s

docs/integrations/hpa.md
```diff
@@ -0,0 +1,92 @@
...
+- `VMDistributed` components that expose `hpa` in their component spec.
+
+## Configuration
+
+The `hpa` section embeds Kubernetes `autoscaling/v2` HorizontalPodAutoscaler spec fields:
+
+- `minReplicas` - minimum number of replicas. Defaults to `1` when omitted.
+- `maxReplicas` - maximum number of replicas.
+- `metrics` - metric sources used for scaling decisions.
+- `behaviour` - scaling behavior configuration.
```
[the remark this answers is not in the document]
vrutkovs: oh, its `behavior` in https://kubernetes.io/docs/reference/kubernetes-api/autoscaling/horizontal-pod-autoscaler-v2/#HorizontalPodAutoscalerSpec, but `behaviour` in https://docs.victoriametrics.com/operator/api/#embeddedhpa

vrutkovs:
My thinking was that its a separate integrations like Istio. Good point on HPA though, not sure if I would like to separate them though

AndrewChubatiuk:
there're some basic concepts, which are applicable for all resources like useStrictSecury, serviceSpec, some global env variables. probably makes sense keep them, HPA and VPA close to each other. most likely as separate pages, but in a common section

vrutkovs:
For now moved this to Configuration, later we may want to have a separate page for this - along with common labels / annotations feature, what do you think?

docs/configuration.md
```diff
@@ -453,21 +453,208 @@ It can be simplified with cert-manager and kustomize command:
...
+
+- VPA `Auto` and `Recreate` modes can evict pods to apply recommendations. Configure PodDisruptionBudgets and replicas for availability.
+- Use `updateMode: Off` when you need recommendations only.
+- HPA and VPA may interact unexpectedly when both manage the same workload.
 
 ---
 
 The following legacy links are retained for historical reference.
 
-###### List of command-line flags
```
AndrewChubatiuk, line 471: please keep it as it is, 6-th level headers are excluded from TOC and used only for backward compatibility

vrutkovs: Done

AndrewChubatiuk (approved):

PR #2313, opened 2026-06-22
vmcluster: support storage node pools

### LINKED ISSUE #741: Multiple statefulsets for vmstorage nodes on vmcluster
Currently operator creates one statefulset for all vmstorage nodes and scales it accordingly. This approach is good until something bad happens to one of storage nodes. For example:
It is not possible to turn off one of vmstorage nodes (for restore purpose)
We cannot keep the corrupted vmstorage for diagnosis while adding another one to keep cluster running.
We cannot change resource limits seperately. This is specially needed during incidents for temporarily adjustments. Moreover, we have seen that some storages behave differently in our cluster, this can happen because of different computing infrastructure or maybe a flaw in load balancing in shard ingestion or search loads.
We propose that the operator deploy each storage as one separate sts. In this way maintenances would be simpler, and also there would be a way to set different resoruce configuration on vmcluster CRD object.
You might have concerns about the simplicity. We can think of a more felxible representation of vmcluster. If vmoperator can support different nodepools in defining vmcluster it can keep it simple to new users as well as allowing people for more advanced designs.
For example:

```
vmstorage:
  nodepools:
    - name: pool1
       replicaCount: 2
       resources: {}
       tolerations: {}
       nodeSelector {}
    - name: pool2
       replicaCount: 1
       resources: {}
       tolerations: {}
       nodeSelector {}
```

### LINKED ISSUE #2574: Question about vmstorage groups setup with distinct replicationFactor in multi-level vmselect
Does the operator support setup [vmstorage groups for multi-level vmselect](https://docs.victoriametrics.com/victoriametrics/cluster-victoriametrics/#vmstorage-groups-at-vmselect)?
I need to define distinct `replicationFactor` per each group. As example here:
```
 -replicationFactor=g1:2
 -storageNode=g1/host1,g1/host2
 -replicationFactor=g2:1
 -storageNode=g2/host4
```

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/741

Adds pools to VMCluster, allowing to run multiple isolated groups of vmstorage nodes (and optionally dedicated vminsert nodes) within a single cluster. Each pool is queried by vmselect as a named storage group, enabling multi-retention.

Each entry in spec.pools has a name, an optional vmstorage override (merges onto the top-level spec.vmstorage), and an optional vminsert override (merges onto the top-level spec.vminsert).
When any pool is defined, the top-level vmstorage StatefulSet is not deployed; pools replace it entirely.
The top-level vminsert Deployment is deployed as a shared insert only when no pool defines its own dedicated vminsert.
vmstorage.retentionPeriod inside a pool overrides the cluster-level spec.retentionPeriod for that pool, enabling per-pool retention.

Pool resources are named `vm<component>-<cluster>-<pool>` (e.g. vmstorage-prod-hot).
An `app.kubernetes.io/pool=<name>` label is added to both the StatefulSet/Deployment selector and pod template labels, keeping per-pool selectors disjoint.


internal/controller/operator/factory/vmcluster/vmcluster_pools.go
```diff
@@ -0,0 +1,173 @@
...
+			return fmt.Errorf("vmstorage pdb: %w", err)
+		}
+	}
+	if err := createOrUpdateVMStorage(ctx, rclient, view, prevView, pool.Name, owner); err != nil {
+		return fmt.Errorf("vmstorage: %w", err)
+	}
+	if err := createOrUpdateVMStorageService(ctx, rclient, view, prevView, owner, pool.Name); err != nil {
+		return fmt.Errorf("vmstorage service: %w", err)
+	}
+	if err := createOrUpdateVMStorageHPA(ctx, rclient, view, prevView, pool.Name); err != nil {
```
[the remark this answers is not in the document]
AndrewChubatiuk: recheck if it's still relevant

AndrewChubatiuk:
@vrutkovs need your feedback on this

vrutkovs:
I'd like to have this merged after release is cut, but in general it looks okay

AndrewChubatiuk:
sure, no rush, just wanted to get some feedback

api/operator/v1beta1/vmcluster_types.go
```diff
@@ -100,6 +100,18 @@ type VMClusterSpec struct {
 	// See https://docs.victoriametrics.com/victoriametrics/cluster-victoriametrics/#automatic-vmstorage-discovery
 	// +optional
 	Discovery *VMClusterDiscovery `json:"discovery,omitempty"`
+
+	// Pools defines named groups of vmstorage (and optionally vminsert) components.
+	// Each pool gets its own StatefulSet and headless Service named <component>-<cluster>-<pool>.
+	// Top-level vmstorage and vminsert specs act as defaults; pool specs override them field-by-field.
+	// vmselect queries all pools using the pool name as a storage group name (-storageNode=<pool>/<addr>).
+	// When pools are defined the top-level vmstorage is not deployed; pools replace it entirely.
+	// The top-level vminsert is deployed as a shared insert across all pools only when no pool
```
vrutkovs:
```suggestion
	// The top-level vminsert is deployed as a shared insert group across all pools only when no pool
```

We're not deploying a single vminsert instance, right?

api/operator/v1beta1/vmcluster_types.go
```diff
@@ -742,6 +768,54 @@ func (cr *VMCluster) GetRemoteWriteURL() string {
...
+	}
+	if vms.RollingUpdateStrategyBehavior != nil {
+		if err := vms.RollingUpdateStrategyBehavior.Validate(); err != nil {
+			return err
+		}
+	}
+	return vms.Validate()
+}
+
+func (vmi *VMInsert) validate() error {
```
vrutkovs, line 835: No license verification? Seems we're only doing this for discovery, but imo its worth having for every type of vminsert

api/operator/v1beta1/vmcluster_types.go
```diff
@@ -867,34 +910,66 @@ func (cr *VMCluster) Validate() error {
...
-		if err := vmselectDiscovery.validate(); err != nil {
-			return fmt.Errorf("vmselect: %w", err)
+	}
+	if err := vmselectDiscovery.validate(cr.Spec.License); err != nil {
+		return fmt.Errorf("vmselect: %w", err)
+	}
+
+	poolNames := make(map[string]struct{}, len(cr.Spec.Pools))
+	for i, pool := range cr.Spec.Pools {
+		if !poolNameRe.MatchString(pool.Name) {
```
vrutkovs: Use `IsDNS1123Subdomain` here? Not sure why we are duplicating `kubebuilder:validation:Pattern` (and not enforcing the length)

docs/CHANGELOG.md
```diff
@@ -18,14 +18,14 @@ aliases:
 
 * FEATURE: [vmoperator](https://docs.victoriametrics.com/operator/): add `networkPolicy` field to all supported CRDs (`VMSingle`, `VMAgent`, `VMAlert`, `VMAlertmanager`, `VMAuth`, `VLSingle`, `VLAgent`, `VTSingle`, `VMAnomaly`, and all cluster sub-components). When set, the operator creates and manages a `NetworkPolicy` resource that restricts ingress/egress to the component's pods. See [#2977](https://github.com/VictoriaMetrics/helm-charts/issues/2977).
 * BUGFIX: [vmoperator](https://docs.victoriametrics.com/operator/): grant the operator's `ClusterRole` permissions on `networking.k8s.io/networkpolicies`. The `networkPolicy` feature above was shipped without this grant, so the operator would fail with a `Forbidden` error creating/updating/deleting `NetworkPolicy` objects on a real cluster.
+* FEATURE: [vmcluster](https://docs.victoriametrics.com/operator/resources/vmcluster/): add `spec.pools` support for multi-retention and tiered-storage setups. Each named pool runs isolated `vmstorage` nodes and optionally a dedicated `vminsert`, all queried transparently by `vmselect` as named storage groups. See [#741](https://github.com/VictoriaMetrics/operator/issues/741).
```
vrutkovs: Group features together

internal/controller/operator/factory/vmcluster/vmcluster_test.go
```diff
@@ -858,6 +859,169 @@ func TestCreateOrUpdate(t *testing.T) {
 				"managed-by":                  "vm-operator",
 			}, svc.Labels)
 		}})
+
+	// pools: two pools with shared vminsert — pool STSes get pool label, instance label stays the cluster name, top-level vmstorage not created
```
vrutkovs, line 901: We should be verifying the storageNode args for vminserts in all tests

api/operator/v1beta1/vmcluster_types.go
```diff
@@ -1231,3 +1306,35 @@ func (cr *VMAuthLoadBalancerSpec) UseTLS() bool {
...
+
+// VMClusterPool defines a named group of vmstorage (and optionally vminsert) components
+// within a VMCluster. Each pool gets its own StatefulSet and headless Service.
+// +k8s:openapi-gen=true
+type VMClusterPool struct {
+	// Name is the unique identifier for this pool within the cluster.
+	// Used as a suffix for generated resource names and as a storage group name in vmselect.
+	// Must be a lowercase alphanumeric DNS label; hyphens allowed in the interior.
+	// +kubebuilder:validation:Pattern:="^[a-z0-9]([a-z0-9-]*[a-z0-9])?$"
+	// +kubebuilder:validation:MaxLength=32
```
vrutkovs: Cluster name is unlimited, why would we limit it here? VMStorage names are very easy to overflow btw, this should be limited to 16 to make "vmstorage-" STS fit in 63 chars max

internal/controller/operator/factory/vmcluster/vmcluster_reconcile_test.go
```diff
@@ -167,15 +167,6 @@ func Test_CreateOrUpdate_Actions(t *testing.T) {
...
-				// VMSelect
-				{Verb: "Get", Kind: "StatefulSet", Resource: vmselectName},
-				{Verb: "Create", Kind: "StatefulSet", Resource: vmselectName},
-				{Verb: "Get", Kind: "StatefulSet", Resource: vmselectName}, // wait for ready
-				{Verb: "Get", Kind: "Service", Resource: vmselectName},
-				{Verb: "Create", Kind: "Service", Resource: vmselectName},
-				{Verb: "Get", Kind: "VMServiceScrape", Resource: vmselectName},
-				{Verb: "Create", Kind: "VMServiceScrape", Resource: vmselectName},
-
 				// VMInsert
```
vrutkovs: Lets comment here that pool logic has changed the order of reconciliation and vmselect is now checked later

internal/controller/operator/factory/vmcluster/vmcluster_test.go
```diff
@@ -858,6 +859,169 @@ func TestCreateOrUpdate(t *testing.T) {
...
+			var topSts appsv1.StatefulSet
+			err := rclient.Get(ctx, types.NamespacedName{Namespace: cr.Namespace, Name: cr.PrefixedName(vmv1beta1.ClusterComponentStorage)}, &topSts)
+			assert.True(t, k8serrors.IsNotFound(err), "top-level vmstorage STS must not exist when pools are defined")
+			// shared vminsert must still be created (no pool has a dedicated insert)
+			var dep appsv1.Deployment
+			assert.NoError(t, rclient.Get(ctx, types.NamespacedName{Namespace: cr.Namespace, Name: cr.PrefixedName(vmv1beta1.ClusterComponentInsert)}, &dep))
+		},
+	})
+
+	// pools: pool with dedicated vminsert — pool insert created, top-level vminsert skipped
```
vrutkovs, line 960: Wouldn't that leave the second pool outside of the ingestion path? I think it would be more explicit to require defining vminsert for every pool

api/operator/v1beta1/vmcluster_types.go
AndrewChubatiuk: this is needed to validate enterprise version dependent features, so in vminsert this check is redundant

internal/controller/operator/factory/vmcluster/vmcluster_test.go
AndrewChubatiuk: updated tests

PR #2315, opened 2026-06-22
fix: rename TracesVertsion to TracesVersion in BaseOperatorConf

immanuwell:
`BaseOperatorConf` has `TracesVertsion` (note the swapped `rt`) instead of `TracesVersion`. 
Its just a naming fix.

Reproduce by grepping the struct:

```
grep -n "Version\b" internal/config/config.go
```

You'll see all the `*Version` fields line up except `TracesVertsion` on line 114.


AndrewChubatiuk (approved):
Thanks for contribution!

PR #2322, opened 2026-06-23
refactor(e2e): use direct Entry(...) calls in upgrade_test.go

vrutkovs:
This allows us to use "FEntry" to focus on particular type for testing


AndrewChubatiuk (approved):

vrutkovs:
This makes test last longer and its harder to distinguish which test is running. Closing this

PR #2326, opened 2026-06-24
introduce useLegacyNaming boolean to use legacy naming (compatible with charts) for operator resources

AndrewChubatiuk:
related issues [#3021](https://github.com/VictoriaMetrics/helm-charts/issues/3021), [#2755](https://github.com/VictoriaMetrics/helm-charts/issues/2755)


api/operator/v1/vlagent_types.go
```diff
@@ -92,7 +92,12 @@ type VLAgentSpec struct {
 
 	// ServiceAccountName is the name of the ServiceAccount to use to run the pods
 	// +optional
-	ServiceAccountName         string `json:"serviceAccountName,omitempty"`
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+
+	// UseLegacyNaming uses standalone Helm chart naming for managed resources:
+	// the CR name is used directly instead of the default "<type>-<name>" convention.
+	// +optional
+	UseLegacyNaming            bool `json:"useLegacyNaming,omitempty"`
```
vrutkovs, line 100: Legacy here is too vague - afaik operator didn't change the name schema recently, so imo `UseHelmChartNaming` or `UseSuffixNamingSchema` is better

vrutkovs (commented):
Lets update `helm-converter.md` to mention this flag (possibly extend it into proper migration procedure)

api/operator/v1/vlagent_types.go
AndrewChubatiuk: imo word legacy makes it clear that final destination is naming convention, which is used by operator

AndrewChubatiuk: we can add opposite flag to charts to allow users to generate resources with naming, which is already compatible with default operator naming

vrutkovs: Okay, I don't feel too strongly about this, `Legacy` is acceptable

Helm chart change would be useful - but it may cause issues for upgraded installations, I don't think we can reliably separate those two?

AndrewChubatiuk: for helm charts we can keep current behaviour by default, but makes sense to mention it everywhere in docs that it's suggested to set that toggle

AndrewChubatiuk: updated converter's docs and implementation, also added a [PR for helm-charts](https://github.com/VictoriaMetrics/helm-charts/pull/3030)

vrutkovs (approved):

PR #2335, opened 2026-06-25
feat: support vmsingle in vmdistributed

### LINKED ISSUE #2267: Created Distributed VMSingle resource
A user would like to something similar to the VMDistributed CR, but does not want the performance overhead and complexity of VMCluster. The ideal solution would be a something similar to the VMDistributed CR, but uses VMSingle instead of VMCluster for managing the Victoriametrics instance. Currently the workaround for this is to deploy multiple VMSingle, VMAgent, and a VMAuth resources, but this more YAML for the user to orchestrate, and cannot managed using the k8s-stack helm chart.

vrutkovs:
Fixes #2267 

Pretty much the same as VMCluster implementation, also we allow mixing VMCluster and VMSingle zones - although its not possible to set both in `zoneCommon`


api/operator/v1alpha1/vmdistributed_types.go
```diff
@@ -136,6 +144,24 @@ func (z *VMDistributedZone) VMClusterName(cr *VMDistributed) string {
...
+		return z.VMSingle.Name
+	case len(cr.Spec.ZoneCommon.VMSingle.Name) > 0:
+		return strings.ReplaceAll(cr.Spec.ZoneCommon.VMSingle.Name, ZonePlaceholder, z.Name)
+	default:
+		return fmt.Sprintf("%s-%s", cr.Name, z.Name)
+	}
+}
+
+// UseVMSingle returns true if the zone should use VMSingle instead of VMCluster
+func (z *VMDistributedZone) UseVMSingle(cr *VMDistributed) bool {
```
AndrewChubatiuk: let's make VMDistributedSpec something like BackendType instead of this condition, this will prevent from having mixed zones and also make it clear how to perform validation and what should user expect in result

internal/controller/operator/factory/reconcile/vmsingle.go
```diff
@@ -0,0 +1,68 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	vmv1beta1 "github.com/VictoriaMetrics/operator/api/operator/v1beta1"
+	"github.com/VictoriaMetrics/operator/internal/controller/operator/factory/logger"
+)
+
+// VMSingle performs an update or create of VMSingle and waits till it's ready
+func VMSingle(ctx context.Context, rclient client.Client, newObj, prevObj *vmv1beta1.VMSingle, owner *metav1.OwnerReference) error {
```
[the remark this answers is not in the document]
vrutkovs: I intentionally copied the reconciler code so that we'd be sure we didn't break VMSingle reconciliation. The followup will unify them

api/operator/v1alpha1/vmdistributed_types.go
vrutkovs: > this will prevent from having mixed zones

I'm not against those, but okay, lets start with basics first and add mixed zones if requested later

api/operator/v1alpha1/vmdistributed_types.go
```diff
@@ -456,10 +506,20 @@ func (cr *VMDistributed) UnmarshalJSON(src []byte) error {
...
 	clusters := sets.New[string]()
+	singles := sets.New[string]()
 	agents := sets.New[string]()
 	spec := cr.Spec
+	isVMSingle := cr.Spec.BackendType == VMDistributedBackendTypeVMSingle
 	hasCommonVMInsert := cr.Spec.ZoneCommon.VMCluster.Spec.VMInsert != nil
 	hasCommonVMSelect := cr.Spec.ZoneCommon.VMCluster.Spec.VMSelect != nil
+	hasCommonVMSingle := cr.Spec.ZoneCommon.VMSingle.Name != "" || cr.Spec.ZoneCommon.VMSingle.Spec != nil
+	hasCommonVMCluster := cr.Spec.ZoneCommon.VMCluster.Name != "" || !equality.Semantic.DeepEqual(cr.Spec.ZoneCommon.VMCluster.Spec, vmv1beta1.VMClusterSpec{})
+	if isVMSingle && hasCommonVMCluster {
```
AndrewChubatiuk, line 517: can you move this to validate?

api/operator/v1alpha1/vmdistributed_types.go
```diff
@@ -456,10 +506,20 @@ func (cr *VMDistributed) UnmarshalJSON(src []byte) error {
...
+		return fmt.Errorf("backendType=VMSingle is incompatible with zoneCommon.vmcluster configuration")
+	}
+	if !isVMSingle && hasCommonVMSingle {
```
AndrewChubatiuk, line 520: can you move this to validate?

api/operator/v1alpha1/vmdistributed_types.go
```diff
@@ -469,14 +529,34 @@ func (cr *VMDistributed) Validate() error {
 			return fmt.Errorf("spec.zones[%d].name=%s is duplicated, zone names must be unique", i, zone.Name)
 		}
 		zones.Insert(zone.Name)
-		clusterName := zone.VMClusterName(cr)
-		agentName := zone.VMAgentName(cr)
-		if len(clusterName) > 0 {
-			if clusters.Has(clusterName) {
-				return fmt.Errorf("spec.zones[%d].vmcluster.name=%s is already added in a different zone", i, clusterName)
+		if isVMSingle {
+			if zone.VMCluster.Name != "" || !equality.Semantic.DeepEqual(zone.VMCluster.Spec, vmv1beta1.VMClusterSpec{}) {
```
AndrewChubatiuk, line 533: it can be also moved to validate

internal/controller/operator/factory/build/defaults.go
AndrewChubatiuk, line 89: vmsingle license should be also added

internal/controller/operator/factory/vmdistributed/vmauth.go
```diff
@@ -60,6 +60,43 @@ func vmClusterTargetRef(vmClusters []*vmv1beta1.VMCluster, owner *metav1.OwnerRe
...
+			Namespace: vmSingle.Namespace,
+		})
+	}
+	ref := vmv1beta1.TargetRef{
+		Name: "read",
+		URLMapCommon: vmv1beta1.URLMapCommon{
+			LoadBalancingPolicy: ptr.To("first_available"),
+			RetryStatusCodes:    []int{500, 502, 503},
+		},
+		Paths: []string{"/api/v1/.+", "/flags", "/metrics"},
```
AndrewChubatiuk, line 85: why flags and metrics are only exposed for vmsingle?

vrutkovs: Right, I forgot to copy this to VMCluster (although I'm not sure if we really need this at all)

AndrewChubatiuk (approved):

PR #2339, opened 2026-06-27
fix: preserve webhook method with reload auth key

immanuwell:
`config-reloader` was forcing `POST` whenever `--reload-url-auth-key` was set, so `--webhook-method` got ignored. That flag combo is kinda busted on current `master`.

This keeps the configured method and sends `authKey` in the query string, so it matches VictoriaMetrics reload auth handling and does the obvious thing.

Repro:
1. Expose a reload endpoint that expects `PUT` and reads `authKey` from query args.
2. Run `config-reloader` with `--webhook-method=PUT` and `--reload-url-auth-key=secret-value`.
3. Current `master` sends `POST`, not `PUT`.
4. This patch sends `PUT` with `?authKey=secret-value`.

Checks:
- `make test`
- `make lint`

### DIFF
--- a/cmd/config-reloader/main.go
+++ b/cmd/config-reloader/main.go
@@ -240,26 +240,25 @@ type reloader struct {
 func (r *reloader) reload(ctx context.Context) error {
 	configReloadsTotal.Inc()
 	authKey := reloadURLAuthKey.Get()
-
-	var resp *http.Response
+	reloadTarget := *reloadURL
 	if len(authKey) > 0 {
-		formData := url.Values{
-			"authKey": []string{authKey},
-		}
-		var err error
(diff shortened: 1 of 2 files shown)


AndrewChubatiuk:
VictoriaMetrics can [parse value](https://github.com/VictoriaMetrics/VictoriaMetrics/blob/1b0e843e8fcbe316656bcb0a0c6468d1299e1222/lib/httpserver/httpserver.go#L478) from both query string and body, I'd prefer to keep it as is. @vrutkovs wdyt?

immanuwell:
> VictoriaMetrics can parse value from both query string and body, I'd prefer to keep it as is

agreed, so I am going to close this PR

PR #2343, opened 2026-06-29
chore: use docker as for vmanomaly renovate version tracking

vrutkovs:
vmanomaly repo is private, but renovate could use docker tags (similar to helm-chart)


AndrewChubatiuk (approved):

AndrewChubatiuk:
@vrutkovs could you please also add test/Makefile to config? now we have different versions of ginkgo binary and library due to this reason

PR #2355, opened 2026-07-01
fix(vmdistributed): parse exact metric names when polling queues

immanuwell:
small parser fix for vmdistributed queue polling.

old code matched metric names by substring. that can pick up `# HELP` and `# TYPE` lines when `-metrics.exposeMetadata` is enabled, and it can also grab similarly prefixed metrics first. kinda nasty, easy to miss.

this patch parses `/metrics` line by line and only reads exact metric-name matches.

repro:
1. enable `-metrics.exposeMetadata` on vmagent.
2. hit `/metrics` and look for `vm_persistentqueue_bytes_pending`.
3. before this patch the parser may match metadata lines first, or a same-prefix metric first.
4. after this patch it reads only real samples for the exact metric.

added regression tests for metadata lines and same-prefix metrics.

### DIFF
--- a/internal/controller/operator/factory/vmdistributed/util.go
+++ b/internal/controller/operator/factory/vmdistributed/util.go
@@ -39,40 +39,54 @@ func fetchMetricValues(ctx context.Context, httpClient *http.Client, url, metric
 }
 
 func unmarshalMetric(values map[string]float64, s, name, dimension string) error {
-	for {
-		idx := strings.Index(s, name)
-		if idx == -1 {
-			break
-		}
-		s = s[idx+len(name):]
-		s = skipLeadingWhitespace(s)
-		n := strings.IndexByte(s, '{')
-		var dimValue string
-		if n >= 0 {
-			// Tags found. Parse them.
+	for len(s) > 0 {
+		line := s
+		if n := strings.IndexByte(s, '\n'); n >= 0 {
+			line = s[:n]
 			s = s[n+1:]
+		} else {
+			s = ""
+		}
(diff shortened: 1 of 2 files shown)


internal/controller/operator/factory/vmdistributed/util_test.go
```diff
@@ -38,6 +38,25 @@ test_metric{path="no path"} 24
 `, "test_metric", "path", map[string]float64{
 		"no path": 24,
 	})
+
+	t.Run("ignores similarly prefixed metrics", func(t *testing.T) {
```
AndrewChubatiuk: please remove t.Run and put test name into comments as in examples above

internal/controller/operator/factory/vmdistributed/util.go
```diff
@@ -39,40 +39,54 @@ func fetchMetricValues(ctx context.Context, httpClient *http.Client, url, metric
...
-		s = s[idx+len(name):]
-		s = skipLeadingWhitespace(s)
-		n := strings.IndexByte(s, '{')
-		var dimValue string
+		line = strings.TrimSpace(line)
+		if len(line) == 0 || strings.HasPrefix(line, "#") {
+			continue
+		}
+
+		n := strings.IndexAny(line, "{ \t\r")
```
vrutkovs, line 55: Not sure what is this for - we want to support labelless metrics? I think later on we require `{` to find tags

AndrewChubatiuk (approved):

AndrewChubatiuk:
@immanuwell, thanks a lot for contribution!

PR #2358, opened 2026-07-02
add openshift compatibility mode

AndrewChubatiuk:
related issue https://github.com/VictoriaMetrics/helm-charts/issues/2971.
suppose it makes more sense to implement it in operator instead of charts, but also added mount of certificate to victoria-metrics-single and victoria-metrics-agent charts as well


docs/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ aliases:
...
 * FEATURE: [vmdistributed](https://docs.victoriametrics.com/operator/resources/vmdistributed/): support using `VMSingle` backends alongside `VMCluster` backends in separate zones.
+* FEATURE: [vmoperator](https://docs.victoriametrics.com/operator/): add `VM_OPENSHIFT_COMPATIBILITY` env var (`OpenshiftCompatibilityMode` config field, default `auto`) to enable OpenShift-specific compatibility features for managed pods. In `auto` mode the operator detects OpenShift at startup via the `security.openshift.io/v1` API group. Currently covers automatic mounting of the `openshift-service-ca.crt` ConfigMap into `VMAgent` and `VMSingle` pods at `/etc/ssl/certs/openshift-service-ca/service-ca.crt`, allowing scrape targets signed with the OpenShift service signing CA to be verified without per-target TLS config. Supported values: `auto`, `enabled`, `disabled`. See [#2971](https://github.com/VictoriaMetrics/helm-charts/issues/2971).
```
```suggestion
* FEATURE: [vmoperator](https://docs.victoriametrics.com/operator/): add `VM_OPENSHIFT_COMPATIBILITY` env var (`OpenshiftCompatibilityMode` config field, default `auto`) to enable OpenShift-specific compatibility features for managed pods. In `auto` mode the operator detects OpenShift at startup via presence of `security.openshift.io/v1` API group. Currently covers automatic mounting of the `openshift-service-ca.crt` ConfigMap into `VMAgent` and `VMSingle` pods at `/etc/ssl/certs/openshift-service-ca/service-ca.crt`, so that targets signed with the OpenShift service signing CA could be verified without additional TLS config. Supported values: `auto`, `enabled`, `disabled`. See [#2971](https://github.com/VictoriaMetrics/helm-charts/issues/2971).
```

internal/controller/operator/factory/build/security.go
```diff
@@ -52,14 +54,104 @@ var (
...
+// restricted-v2 SCC assigns UIDs/GIDs from the namespace-allocated range.
+// Returns nil when OpenShift compatibility is not active or sc is nil.
+func WarnOpenShiftSecurityContext(sc *vmv1beta1.SecurityContext) []string {
+	if !IsOpenShiftCompatibilityActive() || sc == nil || sc.PodSecurityContext == nil {
+		return nil
+	}
+	psc := sc.PodSecurityContext
+	var warnings []string
+	if psc.RunAsUser != nil {
+		warnings = append(warnings, "spec.securityContext.runAsUser is explicitly set; on OpenShift this value must fall within the namespace-allocated UID range or pod admission will fail — consider removing it and letting the SCC assign the UID")
```
vrutkovs:
```suggestion
		warnings = append(warnings, "spec.securityContext.runAsUser is explicitly set; in OpenShift setting  this value may require additional SCC permission - or fall within the namespace-allocated UID range otherwise pod admission will fail; consider removing it and letting the controller assign the UID")
```

internal/controller/operator/factory/build/security.go
```diff
@@ -52,14 +54,104 @@ var (
...
+	}
+	if psc.RunAsGroup != nil {
+		warnings = append(warnings, "spec.securityContext.runAsGroup is explicitly set; on OpenShift this value must fall within the namespace-allocated GID range or pod admission will fail — consider removing it and letting the SCC assign the GID")
```
vrutkovs:
```suggestion
		warnings = append(warnings, "spec.securityContext.runAsGroup is explicitly set; in OpenShift setting  this value may require additional SCC permissions - or fall within the namespace-allocated GID range otherwise pod admission will fail; consider removing it and letting the controller assign the GID")
```

internal/controller/operator/factory/build/security.go
```diff
@@ -52,14 +54,104 @@ var (
...
+	}
+	if psc.FSGroup != nil {
+		warnings = append(warnings, "spec.securityContext.fsGroup is explicitly set; on OpenShift this value must fall within the namespace-allocated GID range or pod admission will fail — consider removing it and letting the SCC assign the fsGroup")
```
vrutkovs:
```suggestion
		warnings = append(warnings, "spec.securityContext.fsGroup is explicitly set; in OpenShift setting  this value may require additional SCC permissions - or fall within the namespace-allocated supplemental group range otherwise pod admission will fail; consider removing it and letting the controller assign the fsGroup")
```

internal/controller/operator/factory/build/security.go
```diff
@@ -52,14 +54,104 @@ var (
...
+		w = append(w, WarnOpenShiftSecurityContext(spec.RequestsLoadBalancer.Spec.SecurityContext)...)
+	}
+	return w
+}
+
+// WarnOpenShiftSecurityContext returns warning messages for each security context field
+// that is explicitly set but would cause pod admission failures on OpenShift when the
+// restricted-v2 SCC assigns UIDs/GIDs from the namespace-allocated range.
+// Returns nil when OpenShift compatibility is not active or sc is nil.
+func WarnOpenShiftSecurityContext(sc *vmv1beta1.SecurityContext) []string {
```
vrutkovs, line 124:
```suggestion
func WarnOpenShiftPodSecurityContext(sc *vmv1beta1.SecurityContext) []string {
```

We should be checking container security contexts too

vrutkovs (changes requested):

internal/controller/operator/factory/build/security.go
AndrewChubatiuk: vmv1beta1.SecurityContext historically includes both container and pod security contexts

vrutkovs: Yes, but we are checking just PodSecurityContext, not ContainerSecurityContext - this is taken into scheduling account too

AndrewChubatiuk: PodSecurityContext's RunAsUser and RunAsGroup are used for both Container and Pod security context

vrutkovs: Lets warn here if those are set?

AndrewChubatiuk: it's already there

vrutkovs (approved):

PR #2364, opened 2026-07-03
vmagent and vmsingle: updated RBAC

AndrewChubatiuk:
updated VMAgent and VMSingle RBAC:
  - with watchNamespace set create role/rolebinding per each namespace
  - remove configmap permissions
  - limit secrets permissions to secret with resource prefixed name and in a certain namespace


internal/controller/operator/factory/vmagent/rbac.go
```diff
@@ -13,96 +13,107 @@ import (
...
+// namespaces is the list of watched namespaces; an empty slice means cluster-wide access.
+func createK8sAPIAccess(ctx context.Context, rclient client.Client, cr, prevCR *vmv1beta1.VMAgent, namespaces []string) error {
+	if len(namespaces) == 0 {
 		if err := ensureCRExist(ctx, rclient, cr, prevCR); err != nil {
 			return fmt.Errorf("cannot ensure state of vmagent's cluster role: %w", err)
 		}
 		if err := ensureCRBExist(ctx, rclient, cr, prevCR); err != nil {
 			return fmt.Errorf("cannot ensure state of vmagent's cluster role binding: %w", err)
 		}
+		// Secrets access must be namespace-scoped even in cluster-wide mode to prevent
```
vrutkovs, line 99: This is a breaking functional change - for instance, you can no longer reuse the same license across multiple namespaces

internal/controller/operator/factory/vmagent/rbac.go
```diff
@@ -155,47 +166,66 @@ func buildCR(cr *vmv1beta1.VMAgent) *rbacv1.ClusterRole {
...
+		o := cr.AsOwner()
+		owner = &o
+	}
+	return reconcile.RoleBinding(ctx, rclient, rb, prevRB, owner)
 }
 
-func buildRole(cr *vmv1beta1.VMAgent) *rbacv1.Role {
-	return &rbacv1.Role{
+// buildRole builds the Role for ns.
+// In cr's own namespace: includes the scoped secrets rule (for config-reloader) plus SD rules,
```
vrutkovs, line 199: That would also be a functionally breaking change - a single VMAgent can't scrape targets across namespaces with secrets?

internal/controller/operator/factory/vmagent/rbac.go
```diff
@@ -155,47 +166,66 @@ func buildCR(cr *vmv1beta1.VMAgent) *rbacv1.ClusterRole {
...
+// with an owner reference so K8s GC cleans it up.
+// In any other watched namespace: SD rules only, no owner reference.
```
vrutkovs, line 201: This would be problematic for garbage collector

internal/controller/operator/factory/vlagent/rbac.go
```diff
@@ -14,21 +14,21 @@ import (
...
-			"list",
-			"watch",
-		},
-		Resources: []string{
-			"pods",
-			"namespaces",
-			"nodes",
-		},
+		Verbs:     []string{"get", "list", "watch"},
+		Resources: []string{"pods", "namespaces"},
```
vrutkovs: That would still be a cluster-wide rule to access all pods in all namespaces

AndrewChubatiuk: yes, just removed nodes permissions if it's not needed. 

actually removed these changes, nodes permissions are required unconditionally

internal/controller/operator/factory/vmagent/rbac.go
AndrewChubatiuk: it's not, operator collects secrets from all scrape configurations and remote writes and writes them all together into a single secret. for license operator just mounts a secret to a pod, it means it already should be in the same namespace. no changes here

internal/controller/operator/factory/vmagent/rbac.go
AndrewChubatiuk: watch namespace supports a list of namespaces, but currently vmagent can only access secrets in a namespace it resides, which is not valid.

internal/controller/operator/factory/vmagent/rbac.go
AndrewChubatiuk: well, same as other resources, namespaced resources have owner references, so even for some reason they were not removed they will be purged during CR removal

internal/controller/operator/factory/vmagent/rbac.go
vrutkovs: Oh, secrets are packed into config.gz now, no need to maintain access to them anymore

vrutkovs (approved):
Thank you for working on this!

Issue #1219, opened 2025-01-14
Possible bug in ServiceMonitor → VMServiceScrape conversion (regex field incorrectly interpreted as array)


rassvetalov:
We are experiencing an issue with the VictoriaMetrics Operator converting a Prometheus Operator ServiceMonitor resource into a VMServiceScrape. Our environment details:
	•	Kubernetes version: 1.29
	•	VictoriaMetrics Operator version: :v0.51.3
	•	Installation method: Helm 

Description of the problem

We have a ServiceMonitor resource that includes metricRelabelings and relabelings with regex and replacement fields. When the VictoriaMetrics Operator attempts to convert this ServiceMonitor into a VMServiceScrape, we receive the following error in the operator logs:
`{"level":"error","ts":"2025-01-14T19:43:42Z","logger":"controller.PrometheusConverter","msg":"cannot update VMServiceScrape","vmservicescrape":"karpenter","namespace":"kube-system","error":"VMServiceScrape.operator.victoriametrics.com \"karpenter\" is invalid: [spec.endpoints[0].relabelConfigs[0].regex: Invalid value: \"array\": spec.endpoints[0].relabelConfigs[0].regex in body must be of type string: \"array\", spec.endpoints[0].metricRelabelConfigs[0].regex: Invalid value: \"array\": spec.endpoints[0].metricRelabelConfigs[0].regex in body must be of type string: \"array\"]"}`
It appears that the operator interprets the regex field as an array, even though in our YAML we have tried putting it as a string

Steps to reproduce
	1.	Install Prometheus Operator and VictoriaMetrics Operator in the same cluster.
	2.	Deploy the following ServiceMonitor (simplified example):
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: karpenter
  namespace: kube-system
spec:
  endpoints:
    - interval: 30s
      port: http-metrics
      path: /metrics
      metricRelabelings:
        - action: drop
          regex: ^karpenter_cloudprovider_instance.*
          sourceLabels:
            - __name__
        - action: replace
          replacement: 'itprod'
          targetLabel: team
      relabelings:
        - action: replace
          regex: ^(.*)$
          replacement: $1
          sourceLabels:
            - __meta_kubernetes_pod_node_name
          targetLabel: nodename
        - action: replace
          sourceLabels:
            - __meta_kubernetes_service_name
          targetLabel: app
  namespaceSelector:
    matchNames:
      - kube-system
[...]

PR #2373, opened 2026-07-07
fix: preserve single regex values in converted ServiceMonitors

immanuwell:
Related to #1219

`StringOrArray` was marshaling nested single values as JSON arrays, so converted `ServiceMonitor` objects could send `regex: ["..."]` instead of `regex: "..."`. That makes `VMServiceScrape` validation blow up.

This switches `MarshalJSON` to a value receiver, so converted relabel regex fields stay strings. Tiny fix, low risk.

Repro:
1. Install Prometheus Operator and VictoriaMetrics Operator.
2. Create a `ServiceMonitor` with single-string `regex` in `metricRelabelings` or `relabelings`.
3. Current `master` can reject the generated `VMServiceScrape` with `regex ... must be of type string: "array"`.
4. This branch keeps `regex` as a JSON string, so the converted object goes through.

Tests:
`go test ./internal/controller/operator/factory/converter`
`cd api && go test ./operator/v1beta1`

### DIFF
--- a/api/operator/v1beta1/vmextra_types.go
+++ b/api/operator/v1beta1/vmextra_types.go
@@ -1021,19 +1021,17 @@ func (m *StringOrArray) UnmarshalJSON(data []byte) error {
 	}
 }
 
-// MarshalJSON implements json.Marshaller interface
-func (m *StringOrArray) MarshalJSON() ([]byte, error) {
-	if m == nil {
-		return nil, nil
-	}
-	ms := *m
+// MarshalJSON implements json.Marshaller interface.
+// Use a value receiver so nested struct fields of type StringOrArray
+// are serialized as strings when they contain a single element.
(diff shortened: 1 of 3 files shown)


api/operator/v1beta1/vmextra_types.go
```diff
@@ -1021,19 +1021,17 @@ func (m *StringOrArray) UnmarshalJSON(data []byte) error {
...
 }
 
-// MarshalJSON implements json.Marshaller interface
-func (m *StringOrArray) MarshalJSON() ([]byte, error) {
-	if m == nil {
-		return nil, nil
-	}
-	ms := *m
+// MarshalJSON implements json.Marshaller interface.
+// Use a value receiver so nested struct fields of type StringOrArray
```
vrutkovs, line 1025: This would break `omitempty` behaviour - instead of `nil` empty fields would receive `""`

PR #2379, opened 2026-07-08
vmrule: remove patch merge key from rules

AndrewChubatiuk:
reverts https://github.com/VictoriaMetrics/operator/pull/2258

multiple k8s-stack default rules are failing as it's a common practice there to use a same rule name to define different thresholds for different severities


vrutkovs (approved):

vrutkovs:
We probably should also update documentation to mention why an idea in  #2258 is useful and would break k8s-stack

PR #2386, opened 2026-07-08
{vm,vl,vt}cluster: fixed volume names while using useLegacyNaming

### LINKED ISSUE #2385: BUG REPORT : useLegacyNaming doesn't handle correctly PVC naming for vlstorage inside VLCluster object
Hi all,

Target : migration from Helm deployed victoria-logs-cluster to operator managed VLCluster without disruption

Issue : naming for vlselect and vlinsert deployment is correct and all objects are handled correctly as it seems but with the problem with PVC cannot confirm because operator failed at vlstorage stage

Descrpition : 
problem with PVC name : iwith helm the pvc are build with the name : vlstorage-volume-<app_name>-vlstorage-<idx_sts> and operator even with useLegacyNaming flag build the PVC name like : vlstorage-db-<app_name>-vlstorage-<idx_sts>

Attended result : when useLegacyNaming is set PVC names are the same than with helm

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/2385

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -13,6 +13,8 @@ aliases:
 
 ## tip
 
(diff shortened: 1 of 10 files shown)


docs/CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ aliases:
 
 ## tip
```
vrutkovs, line 15:
```suggestion
### API breaking change

`GetStorageVolumeName` / `GetCacheMountVolumeName` moved from component types to cluster types
```

vrutkovs (commented):
LGTM, we should note the breaking change in API

PR #2393, opened 2026-07-09
vmalertmanager: default storage fsGroup so silences persist

### LINKED ISSUE #2392: vmalertmanager: silences not persisted on fresh PVC because pod fsGroup is not defaulted
**Problem**

A VMAlertmanager with persistent storage but no `useStrictSecurity` and no user-provided `spec.securityContext` gets a StatefulSet pod with no `securityContext` at all. The alertmanager process runs as uid 65534 (nobody) in the default images, but a freshly provisioned PersistentVolume is owned by root, so the process cannot write the notification log and silences files under `--storage.path`. Silences are silently lost across pod restarts, and depending on the CSI driver you get permission-denied on the data directory.

**Reproduction**

1. Create a VMAlertmanager with `spec.storage` set (a real StorageClass, fresh PVC) and default security settings (no `useStrictSecurity`, no `spec.securityContext`).
2. Create a silence via the UI/API.
3. Delete the pod so it reschedules onto the same PVC.
4. The silence is gone. Inspecting the pod shows no `fsGroup`, and the alertmanager data dir is root-owned and not group-writable by 65534.

**Why fsGroup matters**

`fsGroup` makes the kubelet recursively chown/chmod the mounted volume to the supplemental group so a non-root process can write to it. The operator already defaults `fsGroup: 65534` when `useStrictSecurity: true` (via `getDefaultPodSecurityContext`), but the default path (`addStrictSecuritySettingsToPod`) returns nil, so the common default deployment silently has no fsGroup.

**Our production context**

We run a fleet of VMAlertmanager instances and hit this in production. We currently patch `fsGroup: 65534` onto every VMAlertmanager pod securityContext manually to keep silence persistence working.

**Relevant code**

- `internal/controller/operator/factory/build/security.go` — `addStrictSecuritySettingsToPod` returns nil for the default case
- `internal/controller/operator/factory/vmalertmanager/statefulset.go` — `newStsForAlertManager` sets no securityContext of its own

The same "no default fsGroup on PVC" pattern also applies to VMSingle, VMStorage/VMSelect (vmcluster), and vmagent statefulStorage, but this issue is scoped to VMAlertmanager since silence loss is the most severe and silent failure mode.

**Proposed fix**

When a VMAlertmanager has persistent storage and neither `useStrictSecurity` nor a user `spec.securityContext` is set, default the pod securityContext to `fsGroup: 65534` (with `fsGroupChangePolicy: OnRootMismatch` where supported). Skip under OpenShift, where the assigned SCC manages fsGroup. Any user-provided securityContext or `useStrictSecurity: true` takes precedence unchanged.

A PR implementing this (fix + unit tests + changelog) is ready.

aloks1701:
**What**

Default the VMAlertmanager pod `securityContext.fsGroup` to `65534` when persistent storage is configured and neither `useStrictSecurity` nor a user `securityContext` is set.

Fixes #2392

**Why**

With default security settings the VMAlertmanager pod gets no securityContext, so the alertmanager process (uid 65534 in the default images) cannot write nflog/silences to a freshly provisioned volume. Silences are silently lost on pod restart. `useStrictSecurity: true` already defaults `fsGroup: 65534`; this brings the same protection to the common default deployment when a PVC is in use. We hit this on a production fleet and currently patch fsGroup manually.

**How**

- New helper `build.AddDefaultStorageFSGroupToPod` (in `factory/build/security.go`) reuses the existing `65534` group constant, `IsFSGroupChangePolicySupported`, and OpenShift detection. It only acts when the pod securityContext is still nil, `useStrictSecurity` is false, and no user securityContext is set; it is a no-op under OpenShift.
- Called from `newStsForAlertManager` guarded by `cr.Spec.Storage != nil`.
- Backward compatible: user securityContext and `useStrictSecurity: true` are unchanged.

**Scope**

VMAlertmanager only. The same default-fsGroup gap exists for VMSingle/VMStorage/VMSelect/vmagent statefulStorage and can be a follow-up if this approach is accepted.

**Testing**

- New `TestNewStsForAlertManagerStorageFSGroup` covering: default+PVC gets fsGroup 65534; no storage keeps no securityContext; user securityContext untouched; `useStrictSecurity: true` unchanged.
- `go build ./...`, `go vet ./...` clean; touched package tests pass; changelog entry added. No API types changed, so no codegen needed.


vrutkovs (approved):

PR #2394, opened 2026-07-09
vmalert: ignore alerting rules when no notifiers set

### LINKED ISSUE #2388: VMAlert requires spec.notifier even without alert rules
Hi,

with operator version victoriametrics/operator:v0.66.1 (helm chart victoria-metrics-operator-0.57.1) the manifest below was working fine.

With the latest helm chart which comes with victoriametrics/operator:v0.73.1 it fails:

> Warning   ReconciliationError   vmalert/vmalert  cannot generate prev deploy spec: cannot generate new spec for vmalert: no notifiers found, properly configure selectors or static notifiers using spec.notifiers or spec.notifier

even though the documentation says notification is only required when there are alert rules (sorry couldn't find the source for the docs, so screenshot, it's just from the vmalertspec docs):

Image

so my workaround is to set a dummy url for now.

Manifest to reproduce the issue:

```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: vmtest
---
kind: VMSingle
metadata:
  name: i1
  namespace: vmtest
---
kind: VMAlert
metadata:
  name: vmalert
  namespace: vmtest
spec:
  datasource:
    url: http://vmsingle-i1.vmtest.svc:8428
  remoteWrite:
    url: http://vmsingle-i1.vmtest.svc:8428
#  notifier:
#    url: http://127.0.0.1:28080/does_not_exist
  evaluationInterval: 30s
  ruleSelector: {}
---
kind: VMRule
metadata:
  name: foo1
  namespace: vmtest
spec:
  groups:
  - name: foo1
    rules:
    - record: foo1
      expr: vector(1)
---
kind: VMRule
metadata:
  name: foo2
  namespace: vmtest
spec:
  groups:
  - name: foo2
    rules:
    - record: foo2
      expr: vector(2)
---
kind: VMRule
metadata:
  name: foo3
  namespace: vmtest
spec:
  groups:
  - name: foo3
    rules:
    - record: foo3
      expr: vector(3)
```

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/2388

PR ignores VMRule configurations if VMAlert has not notifiers set


internal/controller/operator/vmrule_controller.go
```diff
@@ -122,7 +122,7 @@ func (r *VMRuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 			}
 		}
 
-		if _, configErr := vmalert.CreateOrUpdateRuleConfigMaps(ctx, r, item, &instance); configErr != nil {
+		if _, _, configErr := vmalert.CreateOrUpdateRuleConfigMaps(ctx, r, item, &instance); configErr != nil {
```
vrutkovs: We should add more detailed checks for cases when we create a vmrule with no alerting and add it later - this should reject this change when notifier is not set

AndrewChubatiuk: i would rather make logic stable, if vmalert has no notifier it just ignores all alerting rules, don't want to have a difference, which depends on order of resources creation

vrutkovs (approved):

internal/controller/operator/vmrule_controller.go
vrutkovs: SGTM

PR #2396, opened 2026-07-10
fix(config): return error on malformed label without '=' instead of panicking

AruneshDwivedi:
Labels.Set split a flag value on commas then on '=' and indexed sp[1], so a label value without an '=' (e.g. a bare key) made sp have one element and the code panicked with index out of range. Return a clear error for malformed key=value pairs instead of crashing at startup. Added a unit test covering valid and malformed input.


AndrewChubatiuk (approved):
LGTM! Could you also add a changelog entry?

vrutkovs (approved):

PR #2399, opened 2026-07-13
vmauth: check config reload happened before completing reconcile

AndrewChubatiuk:
- check config reload event happened before marking reconcile as complete for vmauth
- force waiting for vmauth config reload event in vmdistributed
- use synctest to improve tests speed
- scrape config reloader metrics


AndrewChubatiuk:
@cubic-ai-dev

api/operator/v1beta1/vmagent_types.go
```diff
@@ -283,11 +283,25 @@ func (cr *VMAgent) GetReloadURL(host string) string {
...
+
 // GetReloaderParams implements reloadable interface
 func (cr *VMAgent) GetReloaderParams() *CommonConfigReloaderParams {
 	return &cr.Spec.CommonConfigReloaderParams
 }
 
+// GetConfigReloadMetricNames returns metrics, that vmagent exposes for config-reload completion check
+func (cr *VMAgent) GetConfigReloadMetricNames() []string {
+	return []string{
+		"vm_promscrape_config_last_reload_success_timestamp_seconds",
```
vrutkovs: We may want to add those as consts and reuse them here - it would prevent us from making typos here

internal/controller/operator/factory/reconcile/config_reload.go
```diff
@@ -0,0 +1,115 @@
...
+	configReloadHTTPTimeout  = 5 * time.Second
+)
+
+// WaitForConfigReload verifies, by polling each ready replica's own
+// config-reload-completion metric, that a config change made at or after
+// `after` has actually been picked up by the running process - not just
+// that the merged config was written to a Secret. It's a no-op unless
+// WaitForConfigReload is enabled on the CR and the CR type reports a
+// config-reload-completion metric.
+func WaitForConfigReload(ctx context.Context, rclient client.Client, cr configReloadWaitable, after time.Time) error {
```
vrutkovs: We should pass metric names here (and default to `cr.GetConfigReloadMetricNames` if unset). This would enable us to check exact metric reload time (in case multiple events are fired within one second). I think this is what https://github.com/VictoriaMetrics/operator/pull/2399/changes#r3569190534 is about too

api/operator/v1beta1/vmagent_types.go
AndrewChubatiuk: there's only one metric `vm_promscrape_config_last_reload_success_timestamp_seconds`, which is used in multiple places. not sure consts make sense

internal/controller/operator/factory/reconcile/config_reload.go
AndrewChubatiuk: sorry, don't get your idea

api/operator/v1beta1/vmagent_types.go
vrutkovs: Right, in that case its probably extra. I wonder if we have e2e tests for these to prevent us from referencing metric names which don't exist

AndrewChubatiuk: i can add a warning to logs that metric doesn't exist and exit from check

internal/controller/operator/factory/reconcile/config_reload.go
vrutkovs: If I read this correctly whenever one secret is updated we'll check for all metrics - although we could check for just one metric in some cases. It could help us when multiple metrics are being updated simultaneously and I think there is a situation where we'll issue an extra reload

internal/controller/operator/factory/vmagent/vmagent.go
```diff
@@ -162,19 +171,40 @@ func CreateOrUpdate(ctx context.Context, cr *vmv1beta1.VMAgent, rclient client.C
...
+		changedAssets = append(changedAssets, vmv1beta1.ConfigReloadAssetMain)
+	}
+	if relabelChanged {
+		changedAssets = append(changedAssets, vmv1beta1.ConfigReloadAssetRelabel)
+	}
+	if streamAggrChanged {
+		changedAssets = append(changedAssets, vmv1beta1.ConfigReloadAssetStreamAggr)
+	}
+	if len(changedAssets) > 0 {
+		if err := reconcile.WaitForConfigReload(ctx, rclient, cr, writeStartedAt, changedAssets...); err != nil {
```
vrutkovs: Do we make sure that VMAgent has pods? I think this would timeout if first reconcile didn't have pods created

AndrewChubatiuk: right, updated code

AndrewChubatiuk:
@cubic-ai-dev

AndrewChubatiuk:
@cubic-ai-dev

internal/controller/operator/factory/reconcile/config_reload.go
```diff
@@ -0,0 +1,105 @@
...
+	client.Object
+	SelectorLabels() map[string]string
+}
+
+var (
+	configReloadWaitInterval = 2 * time.Second
+	configReloadWaitTimeout  = 60 * time.Second
+	configReloadHTTPTimeout  = 5 * time.Second
+	configReloaderMetricsURL = func(podIP string) string {
+		return fmt.Sprintf("http://%s:%d/metrics", podIP, build.ConfigReloaderDefaultPort)
```
vrutkovs: This is IPv4-only, use `net.JoinHostPort` instead

internal/podutil/util_test.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	// ready=true: included
+	f(ptr.To(true), []string{"http://10.0.0.1:8080/metrics"})
+
+	// ready=false: excluded
+	f(ptr.To(false), nil)
+
+	// ready=nil (unknown): EndpointSlice semantics treat this as ready, so it must be included
+	f(nil, []string{"http://10.0.0.1:8080/metrics"})
```
vrutkovs, line 55: Lets add ipv6 example (could be in a followup)

vrutkovs (approved):

PR #2401, opened 2026-07-13
feat: support full image reference via image.ref for digest pinning

rrhubenov:
## Summary

Adds an optional `ref` field to the shared `image` spec used across all operator CRs.

Previously the `image` spec only accepted `repository` + `tag`, which assembles the container image as `repository:tag`. This form **cannot reference an image by digest** (`repository@sha256:<digest>`) — there was no supported way to pin images by digest.

When `ref` is set, it is used verbatim as the full image reference and takes precedence over `repository`/`tag`, so images can now be pinned by digest, e.g. `victoriametrics/vmsingle@sha256:<digest>`. When `ref` is set, the global container registry prefix, tag/version defaulting and enterprise suffix are skipped so the reference is used as-is.

Image string assembly is centralized in a new `Image.Reference()` helper, replacing the duplicated `fmt.Sprintf("%s:%s", ...)` sites.

This is fully backward-compatible: `ref` is optional and existing `repository`/`tag` specs behave exactly as before.

## Test plan

- [x] `make generate manifests docs` produces only the additive `ref` field
- [x] `go build ./...` and `go vet ./...` pass in both modules
- [x] New unit test `TestImage_Reference` covers repository+tag, ref-with-tag, ref-with-digest, and ref precedence
- [x] Existing factory package tests pass


vrutkovs:
I'd prefer `tag` to support refs natively, so that we could backport it as a bugfix to LTS. @AndrewChubatiuk wdyt?

AndrewChubatiuk:
> I'd prefer tag to support refs natively, so that we could backport it as a bugfix to LTS
 
agree, just add some check, if tag matches digest pattern use `@` for concatenation, if not `:`. we already have too many properties for setting version

rrhubenov:
can do

rrhubenov:
@vrutkovs @AndrewChubatiuk new approach has been implemented

api/operator/v1beta1/vmextra_types.go
```diff
@@ -1697,12 +1697,29 @@ type ManagedObjectsMetadata struct {
...
+	// (e.g. "sha256:<hex>"); a digest is joined to the repository with "@"
+	// instead of ":", allowing images to be pinned by digest.
 	Tag string `json:"tag,omitempty"`
 	// PullPolicy describes how to pull docker image
 	PullPolicy corev1.PullPolicy `json:"pullPolicy,omitempty"`
 }
 
+// imageDigestTagRegexp matches an image digest reference of the form
+// "algorithm:hex", e.g. "sha256:abc123...", per the OCI/docker reference grammar.
+var imageDigestTagRegexp = regexp.MustCompile(`^[a-z0-9]+(?:[.+_-][a-z0-9]+)*:[0-9a-fA-F]{32,}$`)
```
AndrewChubatiuk:
```suggestion
var imageDigestTagRegexp = regexp.MustCompile(`^[a-zA-Z0-9-_+.]+:[a-fA-F0-9]{32,}$`)
```

AndrewChubatiuk (approved):
LGTM! Thanks for improvement!

vrutkovs (approved):

PR #2411, opened 2026-07-16
removed library/ component from result image, that doesn't contain / in repo name

### LINKED ISSUE #2409: Operator rewrites declarative image paths in an inconsistent manner
When instantiating vmagent custom resources, if the spec contains `image.repository` then the operator adds a `library` prefix to the container's repository string https://github.com/VictoriaMetrics/operator/blob/master/internal/controller/operator/factory/build/container.go#L245 . This causes several inconsistencies:

* Organisations using cloud-hosted private registries would want to use `[registry url]/[organisation]` as the global registry with the operator chart's `global.image.registry` value, and `[image name]` as the `image.repository` in the vmagent custom resource's spec. In this scenario, instead of `[registry url]/[organisation]/[image name]` the vmagent's image is rewritten as `[registry url]/[organisation]/library/[image name]`
* By comparison, the operator does **not** add the `library` infix for the `config-reloader` images. 
* When setting `global.image.registry`, the `image.registry` parameter in the vmagent's spec is ignored.  This means that one cannot override it as a workaround for the agent's image by setting the registry to `[registry url]` and the image to `[organisation]/[image name]`. 

I believe the assumption from #1816 about requiring the `library` infix when `image.repository` does not contain `/` is incorrect, and that the default mode should be to not add the infix **unless** the registry matches a certain pattern.

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/2409


internal/controller/operator/factory/build/container_test.go
```diff
@@ -283,7 +283,6 @@ func TestFormatContainerImage(t *testing.T) {
 	// replace registry
 	f("private.github.io", "my-private.registry/victoria-metrics/storage", "my-private.registry/victoria-metrics/storage")
 	f("docker.proxy.org/public-repo-proxy-docker-hub", "docker.proxy.org/public-repo-proxy-docker-hub/victoriametrics/vmauth", "docker.proxy.org/public-repo-proxy-docker-hub/victoriametrics/vmauth")
```
vrutkovs, line 285:
```suggestion
	f("docker.proxy.org/public-repo-proxy-docker-hub", "docker.proxy.org/public-repo-proxy-docker-hub/victoriametrics/vmauth", "docker.proxy.org/public-repo-proxy-docker-hub/victoriametrics/vmauth")
	// library image
	f("docker.io/library/", "alpine", "docker.io/library/alpine")
```

Lets replace the test case with the expected way to use "library" images

vrutkovs (approved):

PR #2413, opened 2026-07-17
VLDistributed implementation

### LINKED ISSUE #1960: Create VL Distributed Resource
To avoid having to create multiple PRs when updating Clusters that are deployed in Multiple AZs it would be create if there was a distributed resource similar to the VMDistributed for logs so upgrades would only require upgrading 1 PR, and minimal human interaction.

vrutkovs:
Similar to `VMDistributed` this implements an object to control multiple region-separated cluster.

The PR abstracts metrics handling, but the main control flow is duplicated - in case some resource-specific changes would be required. Later we can abstract zone handling 

Fixes #1960


internal/podutil/prometheus.go
```diff
@@ -1,43 +1,11 @@
...
-	"net/http"
 	"strconv"
 	"strings"
-
-	vmv1alpha1 "github.com/VictoriaMetrics/operator/api/operator/v1alpha1"
-	"github.com/VictoriaMetrics/operator/internal/controller/operator/factory/build"
-	"github.com/VictoriaMetrics/operator/internal/controller/operator/factory/k8stools"
 )
 
-func fetchMetricValues(ctx context.Context, httpClient *http.Client, url, metricName, dimension string) (map[string]float64, error) {
```
AndrewChubatiuk, line 16: in [this PR](https://github.com/VictoriaMetrics/operator/pull/2399) moved it to internal/podutil, going to use it later for [migration tool](https://github.com/VictoriaMetrics/operator/pull/2412) as well

AndrewChubatiuk: same is about endpoint slices discovery

vrutkovs: moved to `internal/podutil`

internal/controller/operator/factory/vmdistributed/zone_test.go
```diff
@@ -201,6 +202,13 @@ func TestGetZones(t *testing.T) {
 }
 
 func TestWaitForEmptyPQ(t *testing.T) {
+	// Pre-compute the expected backend URL for VMCluster named "test" in ns "default"
```
vrutkovs, line 204: This may be useful for setups where users decided to hide the url (I think its on by default?), but I'm not sure if its a good idea - hashing format may change?

internal/controller/operator/factory/reconcile/reconcile.go
```diff
@@ -213,13 +217,19 @@ func waitForStatus[T client.Object, ST StatusWithMetadata[STC], STC any](
 		if lastStatus != nil && !limiter.Throttle() {
 			logger.WithContext(ctx).V(1).Info(fmt.Sprintf("waiting for %T=%s to be ready, current status: %s", obj, nsn.String(), string(lastStatus.UpdateStatus)))
 		}
+		if lastStatus != nil && lastStatus.UpdateStatus == vmv1beta1.UpdateStatusFailed && isCluster {
```
AndrewChubatiuk, line 220: why is this only for cluster?

internal/controller/operator/factory/vldistributed/vmauth.go
```diff
@@ -0,0 +1,143 @@
+package vldistributed
```
AndrewChubatiuk, line 1: looks like whole file can be moved to shared code and be used by both distributed specs

internal/controller/operator/factory/vldistributed/zone.go
```diff
@@ -0,0 +1,446 @@
+package vldistributed
```
AndrewChubatiuk, line 1: this whole file also looks like candidate to be shared between different distributed CRs

internal/controller/operator/factory/reconcile/reconcile.go
vrutkovs: There is no failure condition for VMAuth, for instance

internal/controller/operator/factory/vldistributed/zone.go
vrutkovs: Agree, but I'd prefer to optimize this later - I suspect there might be VM / VL specifics here

AndrewChubatiuk (approved):

internal/config/config.go
```diff
@@ -143,6 +143,9 @@ type BaseOperatorConf struct {
 	// remove the pod from rotation before the process shuts down.
 	// Set to false before upgrading the operator to prevent rolling updates on existing resources.
 	EnableDefaultPreStopHook bool `default:"true" env:"VM_ENABLE_DEFAULT_PRESTOP_HOOK"`
+	// Disables automatic -remoteWrite.showURL=true injection for VMDistributed-managed VMAgents.
+	// Use it during upgrades from operator versions that did not set this flag to avoid pod rollouts.
+	VMDistributedDisableRemoteWriteShowURL bool `default:"false" env:"VM_VMDISTRIBUTED_DISABLE_REMOTE_WRITE_SHOW_URL"`
```
AndrewChubatiuk: I'm afraid of this param as expect requests for global defaults for other flags as well and why spec.zoneCommon doesn't work for this purpose? also why this vmagent param is specific to distributed CR? i would rather avoid adding it at all

vrutkovs: This is "don't configure vmagent to show URL", so it needs to be at VMDistributed level. Let me play around with this some more - I think its not needed

PR #2427, opened 2026-07-21
reconcile: log ConfigMap changed keys and sizes instead of full values

### LINKED ISSUE #2426: reconcile: ConfigMap data_diff logs full values, producing multi-MB log lines
### Is your feature request related to a problem?

The ConfigMap reconcile logs the full data diff at INFO with no size bound:

```go
// internal/controller/operator/factory/reconcile/configmap.go:52 (v0.72.0, identical on master)
logger.WithContext(ctx).Info(fmt.Sprintf("updating ConfigMap %s", newCM.Name), "data_diff", dataDiff, "bin_data_diff", binDataDiff)
```

`diffDeep` records the complete desired/current values through its go-cmp `Reporter` (`toInterface` returns `v.Interface()` verbatim), so for large ConfigMaps (in our case VMAgent relabel/scrape configs and rule files) a single log line reaches several MB. Concretely we see 2-5 MB single lines from `victoriametrics/operator:v0.72.0`.

This breaks downstream log pipelines that enforce per-line limits. We ship cluster logs to VictoriaLogs: vlinsert skips lines above `-insert.maxLineSizeBytes` (default 2MiB), and when the skipped line sits mid-request in an Elasticsearch `_bulk` payload, the create/payload framing desynchronizes and the entire batch is rejected — unrelated pods' log lines are dropped together with the operator's diff line. (Somewhat ironically, the operator's own logs are the ones breaking VictoriaLogs ingestion.)

There is currently no way to bound this: the operator uses controller-runtime zap (no argument-length cap; `-loggerMaxArgLen` from VictoriaMetrics `lib/logger` is only compiled into the config-reloader binary), and the only chart-level knob is `--zap-log-level=error`, which silences all reconcile visibility.

The Secret reconcile already handles this exact concern deliberately — `secret.go` computes the diff but logs only a boolean summary (`data_changed=%t`) instead of the payload. Two options, either would resolve it:

1. Adopt the Secret pattern for ConfigMap: log the changed key names (and value sizes), not the values themselves.
2. Alternatively, cap per-value size in `diff.go`'s `Report`/`toInterface` (truncate with an explicit marker), so all `*_diff` fields are bounded.

Option 1 keeps the current usefulness (which keys changed) while making the line size independent of ConfigMap content.

- `--zap-log-level=error`: loses all reconcile INFO logging; too blunt.
- Truncating in the log collector: mutates log content in transit and has to be maintained per-pipeline; the payload-free log message at the source is strictly better.

### Additional information

- Observed on `victoriametrics/operator:v0.72.0` (chart `victoria-metrics-k8s-stack 0.85.x`); the logging site is unchanged in `v0.73.1` and master.
- The diff logging was introduced in v0.55.0 (#1271) and reformatted in v0.69.0 (#1821); neither added a size bound.
- Happy to send a PR for option 1 if the approach is acceptable.

dberkerdem:
### Why

Fixes #2426.

The ConfigMap reconcile logs the full `data_diff`/`bin_data_diff` at INFO with no size bound, so for large ConfigMaps (VMAgent relabel/scrape configs, rule files) a single log line reaches several MB. This breaks downstream log pipelines that enforce per-line limits — in our case VictoriaLogs itself: vlinsert skips lines above `-insert.maxLineSizeBytes`, and a skipped line mid-`_bulk`-payload desynchronizes the framing and rejects the whole batch.

Approach 1 from the issue, as confirmed by @AndrewChubatiuk: keep logging which keys changed, drop the values.

### What changed

- `ConfigMap` reconcile now logs the changed key names with the value byte size on each side, instead of the full values. Example line:

  Before:
  ```
  updating ConfigMap name=monitoring/vmagent-config data_diff={"data['scrape.yml']":{"++":"<2MB of yaml>","--":"<2MB of yaml>"}} ...
  ```
  After:
  ```
  updating ConfigMap name=monitoring/vmagent-config data_diff=["data['scrape.yml'](current=2097152B,desired=2097201B)"] bin_data_diff=[]
  ```

- The changed-key summary is computed directly from the desired/current maps (new `changedKeysWithSizes` helper) rather than post-processing the `diffDeep` output. Two reasons:
  - go-cmp descends into `[]byte` per element, so a changed `binaryData` value would otherwise produce one diff entry per changed byte index — still unbounded.
  - it also skips the go-cmp reflection walk over megabytes of ConfigMap data on every reconcile.

  Semantics of the update decision are unchanged (nil and empty maps compare equal, matching the previous `cmpopts.EquateEmpty` behavior). `diffDeep` itself and all its other callers are untouched; `Secret` already logs only `data_changed=%t`.

- Changelog entry under `tip`.

### Test evidence

- `go test ./internal/controller/operator/factory/reconcile/` — 38 tests pass, including new `TestChangedKeysWithSizes` (added/changed/removed keys, nil-vs-empty maps, binary data).
- `bin/golangci-lint run ./internal/controller/operator/factory/reconcile/...` — 0 issues.
- `go build ./...`, `go vet`, `gofmt` clean.


AndrewChubatiuk (approved):

AndrewChubatiuk:
hey @dberkerdem 
could you please sign your commit?

dberkerdem:
> hey @dberkerdem could you please sign your commit?

done!

PR #2438, opened 2026-07-23
operator: added Ignored status for child objects, do not mark object as failed if it was applied on at least one parent

### LINKED ISSUE #2432: VMRule status not updated when no longer in ruleSelector
The status of a VMRule will go from null to 'operational' when it's label is updated to match the rule selector. However, when the VMRule no longer matches the rule selector the status stays 'operational' despite the corresponding configuration getting removed from the rulesfiles.

The VMRule status shouldn't say 'operational' when it isn't.

AI says other VM resource statuses are susceptible to this same issue, but I haven't confirmed that. I did see the VMRule issue irl.

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/2432

- added Ignored status in child object was not picked by any parent
- do not fail child if it was applied on at least one parent


AndrewChubatiuk:
@cubic-ai-dev recheck

internal/controller/operator/factory/vmanomaly/config/config.go
```diff
@@ -93,7 +93,7 @@ func (pos *ParsedObjects) UpdateStatusesForChildObjects(ctx context.Context, rcl
 		return nil
 	}
 	parentObject := fmt.Sprintf("%s.%s.vmanomaly", cr.Name, cr.Namespace)
-	return reconcile.StatusForChildObjects(ctx, rclient, parentObject, []*vmv1.VMAnomalyConfig{childObject})
+	return reconcile.StatusForChildObject(ctx, rclient, parentObject, childObject)
```
vrutkovs: We return `nil` when `childObject == nil` before we get a chance to set status. Similar to other call sites we should run `StatusForChildObjects` with `pos.configs.All()` when its nil

AndrewChubatiuk: updated

AndrewChubatiuk:
@cubic-ai-dev

vrutkovs (approved):

PR #2440, opened 2026-07-23
vmanomaly: support v1.30 configuration

fred-navruzov:
> vmanomaly: support v1.30 configuration + historical inconsistency fixes

## Why

vmanomaly v1.30 adds Temporal Envelope models (our new important recommended model), stronger online-model controls, causal autotune options, and separate reader fetch/processing limits. The Operator decodes `configRawYaml` strictly, so these fields and classes were rejected even when the same configuration was valid in vmanomaly.

Reader settings need an additional integration step: the Operator deliberately rebuilds `reader` from `spec.reader` and preserves only the raw query map. Adding fields only to the internal YAML struct would therefore accept them initially but discard them before writing the generated configuration.

## What changed

- update the default vmanomaly image from `v1.29.7` to `v1.30.0`;
- support `temporal_envelope` and `temporal_envelope_multivariate`, including seasonal presets, learned holidays, forecast offsets, confidence-boundary controls, dependency scoring, channel limits, `groupby`, and warmup settings;
- resolve the migrated `zscore`, `mad`, `std`, and `rolling_quantile` aliases to their v1.30 online implementations while retaining their deprecated full class paths;
- support `history_strength` on online z-score, MAD, and quantile models and `min_n_samples_seen` on standard deviation;
- support the v1.30 autotune contract: frozen model parameters, validation layout, exact online validation, complexity tie-breaking, F-beta weight, parallelism, and Optuna execution controls;
- add `spec.reader.fetchTimeout`, `processingTimeout`, `queryLastSeenMaxLookback`, and `seriesProcessingBatchSize`, and propagate them into generated vmanomaly YAML;
- close strict-decoding gaps for common model fields such as per-model TTL/data range, relative deviation, scalar-or-pair scale/deviation values, and `forecast_at`;
- preserve meaningful explicit zero/empty values (`seed: 0`, `dependency_rank: 0`, `random_state: 0`, `seasonalities: []`, and `optimize_complexity: false`) instead of dropping them through `omitempty` and unintentionally restoring vmanomaly defaults;
- regenerate the CRD overlays and API/environment reference documentation.

## Compatibility

- Existing model aliases and deprecated class paths remain accepted.
- Existing CR fields are unchanged; the reader additions are optional.
- `optimization_params.timeout` is represented as numeric seconds, matching the vmanomaly v1.30 API. The previous Operator duration-string representation produced a value vmanomaly's autotune schema does not accept.
- Users who pin `spec.image.tag` are unaffected by the default image bump. Unpinned VMAnomaly resources will roll to v1.30.0 with the Operator release containing this change.

## Example

```yaml
spec:
  reader:
    datasourceURL: http://vmsingle:8428
    samplingPeriod: 1m
    fetchTimeout: 15s
    processingTimeout: 45s
    queryLastSeenMaxLookback: 2h
    seriesProcessingBatchSize: 16
  configRawYaml: |
    models:
      envelope:
        class: temporal_envelope
        queries: [cpu]
        seasonalities: [hod_smooth, dow_smooth]
        min_n_samples_seen: 16
        loss_reactivity: 5
      tuned_joint_envelope:
        class: auto
        tuned_class_name: temporal_envelope_multivariate
        queries: [cpu]
        optimization_params:
          anomaly_percentage: 0.02
          exact: true
          optimize_complexity: true
          frozen_params:
            groupby: [cluster]
            seasonalities: [hod_smooth, dow_smooth]
```

## Validation

- `go test ./internal/controller/operator/factory/vmanomaly/config`
- `go test ./operator/v1` from the `api` module
- `make docs` (includes generation, formatting, vet, build, CRD rendering, and API/environment docs)
- `make lint`
- cross-checked representative generated values with the vmanomaly v1.30 Pydantic schemas, including explicit-zero and empty-seasonality cases
- `git diff --check`


AndrewChubatiuk (approved):

vrutkovs (approved):

internal/controller/operator/factory/vmanomaly/config/models.go
```diff
@@ -151,16 +166,61 @@ type autoTunedModel struct {
...
+	Trials                  *int          `yaml:"n_trials,omitempty"`
+	Timeout                 *float64      `yaml:"timeout,omitempty"`
+	Jobs                    *int          `yaml:"n_jobs,omitempty"`
+	ShowProgressBar         bool          `yaml:"show_progress_bar,omitempty"`
+	GCAfterTrial            bool          `yaml:"gc_after_trial,omitempty"`
+}
+
+func (p *autoTunedOptimizationParams) validate() error {
+	if p.AnomalyPercentage == nil {
+		return fmt.Errorf("anomaly_percentage is required")
```
vrutkovs, line 192: This breaks e2e tests

AndrewChubatiuk: @fred-navruzov is it now required for each model? or how should it work?

fred-navruzov: Actually, `anomaly_percentage` being required is not new in vmanomaly v1.30: it has been required for deployable `class: auto` models since [AutoTunedModel was introduced in v1.12](https://docs.victoriametrics.com/anomaly-detection/changelog/#v1120). The [current model documentation](https://docs.victoriametrics.com/anomaly-detection/components/models/#autotuned) also lists it as a minimum required setting.

vmoperator previously hid an omission: the field was a non-pointer `float64` without `omitempty`, so missing input became Go zero and was serialized as `anomaly_percentage: 0`. The backend therefore never saw a missing value, and the E2E fixture passed with an implicit zero. (`0.01` is a documentation example; the `0.05` default belongs to the separate v1.30 asynchronous autotune API.)

To support v1.30 without breaking existing resources, I suggest we:

1. Keep the pointer, but normalize `nil` to explicit `0` before validation/marshaling.
2. Keep the `[0, 0.5)` range validation.
3. Make the E2E fixture explicit with `optimization_params.anomaly_percentage: 0`.
4. Add regression coverage for omission -> emitted `0`, explicit `0`, and invalid boundary values.

This makes the legacy behavior deliberate while retaining backend-compatible validation. Requiring explicit user input later could be a separate migration/deprecation change.

@vrutkovs, @AndrewChubatiuk, wdyt?

AndrewChubatiuk: if this parameter is required, let's treat this change as a bugfix. updated e2e tests already

vrutkovs: I'm fine with that as long as it doesn't invalidate existing setups. Perhaps we should default it to 0.01 now instead of 0?

PR #2443, opened 2026-07-25
helm-coverter: fixed VMUser conversion

### LINKED ISSUE #2441: BUG REPORT (helm-converter) : username is mandatory for VMUser but vmauth won't start if username is defined with a bearer token
Hi all,

Issue : for converting vmauth values file I need to defined username key inside config section even if a bearer token is defined, BUT vmauth forbid the username key if bearer_token is defined with the message:
```
{"ts":"2026-07-24T07:53:35.691Z","level":"fatal","caller":"VictoriaMetrics/app/vmauth/auth_config.go:833","msg":"cannot load auth config: failed to parse -auth.config=\"/config/auth.yml\": failed to parse users from auth config: username and password cannot be specified if bearer_token is set"}
```

I had this issue because try to test migration from helm to operator managed vmauth so for my test I deploy a test vmauth component with Helm then try to apply the converted value to operator. But for having the converted value I MUST define username on each user definition even if it is a bearer token authentication.

From my point of view it will be better to have always the username because if I want to track slow queries or anormal queries from a user even if it is authenticated with a token it's easier to track username than tracking token in logs.

Don't hesitate to tell me if I have to open a ticket on victoriametrics Github about this username key 

EDIT 1 : I try one VMUser generated by helm-converter and it was rejected by the operator with the following message:
```
Error from server (Forbidden): error when creating "test-vmuser-op.yaml": admission webhook "vmusers.operator.victoriametrics.com" denied the request: one of spec.{username,bearerToken,jwt} must be defined for user, got: [username,bearerToken]
```

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/2441

### DIFF
--- a/internal/converter/converter.go
+++ b/internal/converter/converter.go
@@ -104,6 +104,7 @@ type VMAuthConfigValues struct {
 // operator types whose field names already match vmauth's native keys. Backend TLS
 // (tls_ca_file etc.) isn't covered: it doesn't map onto the operator's nested tlsConfig.
 type VMAuthConfigUser struct {
+	Name         string                                     `yaml:"name,omitempty" json:"name,omitempty"`
 	Username     string                                     `yaml:"username,omitempty" json:"username,omitempty"`
 	Password     string                                     `yaml:"password,omitempty" json:"password,omitempty"`
 	BearerToken  string                                     `yaml:"bearer_token,omitempty" json:"bearer_token,omitempty"`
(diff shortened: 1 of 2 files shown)


internal/converter/converter.go
```diff
@@ -1622,32 +1623,49 @@ func ConvertVMAuthUsers(vmauthName, namespace string, values *VMAuthHelmValues)
 	}
 	users := make([]*vmv1beta1.VMUser, 0, len(values.Config.Users))
 	for i, u := range values.Config.Users {
-		if u.Username == "" {
-			return nil, fmt.Errorf("config.users[%d]: username is required", i)
+		if u.Username == "" && u.BearerToken == "" {
+			return nil, fmt.Errorf("config.users[%d]: one of name, username or bearer_token is required", i)
```
vrutkovs:
```suggestion
			return nil, fmt.Errorf("config.users[%d]: one of username or bearer_token is required", i)
```

`Name` is not really required

internal/converter/converter.go
```diff
@@ -1622,32 +1623,49 @@ func ConvertVMAuthUsers(vmauthName, namespace string, values *VMAuthHelmValues)
...
+			objectName = u.Username
+		}
+		if objectName == "" {
+			objectName = fmt.Sprintf("user-%d", i)
+		}
+
 		targetRefs, err := convertVMAuthConfigUserTargetRefs(u)
 		if err != nil {
-			return nil, fmt.Errorf("config.users[%d] (username=%q): %w", i, u.Username, err)
+			return nil, fmt.Errorf("config.users[%d] (name=%q): %w", i, objectName, err)
```
vrutkovs, line 1789: We don't know whether its `name` or `username`. Perhaps we should store `objectParam` alongside `objectName`? Do we even want to make it so complicated?

internal/converter/converter_test.go
```diff
@@ -1256,14 +1259,51 @@ func TestConvertVMAuthUsers_Errors(t *testing.T) {
 	_, err := ConvertVMAuthUsers("test-name", "test-ns", &VMAuthHelmValues{
 		Config: &VMAuthConfigValues{Users: []VMAuthConfigUser{{Password: "x"}}},
 	})
-	assert.ErrorContains(t, err, "username is required")
+	assert.ErrorContains(t, err, "one of name, username or bearer_token is required")
```
vrutkovs:
```suggestion
	assert.ErrorContains(t, err, "one of username or bearer_token is required")
```

internal/converter/converter.go
AndrewChubatiuk: it's name, as we're using it while setting spec.name

vrutkovs (approved):

PR #2457, opened 2026-07-28
test: add upgrade tests for LTS releases 0.68.6 and 0.68.7

vrutkovs:


AndrewChubatiuk (approved):
is it worth keeping all 0.68.x releases here?

vrutkovs:
I'd rather keep them, if possible. Customers are unfortunately conservative and very likely to get stuck on 0.68.2 for a while - and them attempt to straight to the latest version since it has the feature they want

Issue #1396, opened 2025-06-02
rendered crd is different than in docs


cmergenthaler:
Hello,

while configuring our VM setup with CRs and following the [docs](https://docs.victoriametrics.com/operator/api/), we noticed that kubeconform fails even though operator is able to reconcile the deployed resources without problems. We investigated the problem and found out some fields are removed from rendered CRDs, resulting in a deviation between operator capabilities and the actual k8s CRDs. Is there a specific reason why some fields are being removed? 

**Example:**
```yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMSingle
spec:
  extraEnvs:
    - name: GOMEMLIMIT
      valueFrom:
        resourceFieldRef:
          resource: limits.memory
```

Is a valid CR from operator and [docs](https://docs.victoriametrics.com/operator/api/#vmsinglespec-extraenvs) perspective, it is working fine deploying the resource.

Still kubeconform fails validating the resource with the official CRD: `VMSingle is invalid: problem validating schema. Check JSON formatting: jsonschema: '/spec/extraEnvs/0' does not validate with https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/operator.victoriametrics.com/vmsingle_v1beta1.json#/properties/spec/properties/extraEnvs/items/additionalProperties: additionalProperties 'valueFrom' not allowed`

The `valueFrom` (along with others) is removed from the rendered CRD here:
https://github.com/VictoriaMetrics/operator/blob/690ef8991d3265172b9cc6e6cff0c569f1d5fbc4/config/crd/patches/operator.victoriametrics.com_vmsingles.yaml#L18-L19

These removals block us from shift-left CR validation.

f41gh7:
Hello, 

 Looks like it's an issue at kubeconform side. It must respect `x-kubernetes-preserve-unknown-fields` for raw data https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#rawextension

 The reason, why CRD pruned in such way is the size of result yaml file. Kubernetes has default 1.4MB limitation for created objects. And many deployment tools, such as helm stores configuration files ( CRD included) into Secret/Configmap. It makes impossible to install operator CRD with helm without fields pruning. It takes too much size (even compressed). 

 Related kubeconform issue:
https://github.com/yannh/kubeconform/issues/224

AndrewChubatiuk:
hey @cmergenthaler
[...]

PR #2460, opened 2026-07-29
fix extraEnvs with resourceFieldRef in crd schemas

monotek:
When using the CRDs to create a CRD schema to validate resources in CI via kubeconform we run into issuses when extraEnvs uses resourceFieldRef like:

```
    extraEnvs:
      - name: GOMEMLIMIT
        valueFrom:
          resourceFieldRef:
            containerName: vlstorage
            resource: requests.memory
```

kubeconform fails with "additional properties 'valueFrom' not allowed":

> Run ./scripts/cicd/kubeconform.sh kubernetes/namespaces/logging
Starting validation of kubernetes/namespaces/logging namespace
validating kubernetes/namespaces/logging/dev/de1
dist/kubernetes/namespaces/logging/dev/de1/VLCluster_vlcluster_6.yaml - VLCluster vlcluster is invalid: problem validating schema. Check JSON formatting: jsonschema validation failed with 'file:///home/runner/work/mops/mops/crd-schemas/vlcluster_operator.victoriametrics.com_v1.json#' - at '/spec/vlstorage/extraEnvs/0': additional properties 'valueFrom' not allowed
Summary: 8 resources found in 8 files - Valid: 7, Invalid: 1, Errors: 0, Skipped: 0
Error: Process completed with exit code 1.

The Go type behind `extraEnvs` is the stock `corev1.EnvVar` (supports `resourceFieldRef`, `fieldRef`, `secretKeyRef`, `configMapKeyRef`, `fileKeyRef` out of the box), so the Kubernetes API server always accepted these values and the operator worked correctly at runtime. The problem was purely in the generated CRD schema: a kustomize JSON6902 patch stripped the entire `valueFrom` sub-schema from `extraEnvs.items` and replaced it with a blanket `x-kubernetes-preserve-unknown-fields: true` on the item itself. That satisfies the API server (unknown fields are preserved unvalidated) but breaks strict schema-based validators like `kubeconform`, which see `valueFrom` as an undeclared/"additional" property.

This same copy-pasted patch existed for `extraEnvs` in **every** CRD that has it — not just VLCluster.

## Changes

- Removed the `extraEnvs`-stripping patch stanzas from all 12 affected files under `config/crd/patches/`:
  - `vlagents`, `vlclusters` (vlstorage/vlselect/vlinsert), `vlsingles`
  - `vmagents`, `vmalertmanagers`, `vmalerts`, `vmanomalies`, `vmauths`, `vmclusters` (vmstorage/vmselect/vminsert), `vmsingles`
  - `vtclusters` (storage/select/insert), `vtsingles`
- Regenerated `config/crd/overlay/crd.yaml` and `crd.descriptionless.yaml` via `controller-gen v0.21.0` + `kustomize` (matching the pinned tool versions in the Makefile), so the full `EnvVarSource` schema is now present for `extraEnvs.valueFrom` everywhere.

This is purely a CRD-schema fix — no Go/API types, controller logic, or reconcile behavior changed. `x-kubernetes-preserve-unknown-fields` is still used (unchanged) for the genuinely large/recursive fields (`affinity`, `containers`, probes, `volumes`, etc.) where it exists to keep CRD size down; `EnvVar` doesn't need that treatment since it's a small, non-recursive type and controller-gen now emits its full schema natively.

## Test plan

- [x] Regenerated CRDs with the pinned `controller-gen`/`kustomize` versions and confirmed the diff is scoped to the `extraEnvs.valueFrom` schema additions only (no unrelated churn).
- [x] Validated a VLCluster manifest with `vlstorage.extraEnvs[].valueFrom.resourceFieldRef` against the regenerated schema — passes.
- [x] Spot-checked all affected CRDs (`vmclusters`, `vtclusters`, `vmagents`, `vlagents`, etc.) to confirm `resourceFieldRef` (and the other `valueFrom` variants) are present and validated in each.

### DIFF
--- a/config/crd/overlay/crd.descriptionless.yaml
+++ b/config/crd/overlay/crd.descriptionless.yaml
@@ -264,10 +264,81 @@ spec:
                       type: string
                     value:
                       type: string
+                    valueFrom:
+                      properties:
+                        configMapKeyRef:
+                          properties:
+                            key:
+                              type: string
+                            name:
+                              default: ""
+                              type: string
+                            optional:
+                              type: boolean
+                          required:
+                          - key
+                          type: object
+                          x-kubernetes-map-type: atomic
+                        fieldRef:
+                          properties:
+                            apiVersion:
+                              type: string
+                            fieldPath:
+                              type: string
+                          required:
+                          - fieldPath
+                          type: object
+                          x-kubernetes-map-type: atomic
+                        fileKeyRef:
+                          properties:
+                            key:
+                              type: string
+                            optional:
+                              default: false
+                              type: boolean
+                            path:
+                              type: string
+                            volumeName:
+                              type: string
+                          required:
+                          - key
+                          - path
+                          - volumeName
+                          type: object
+                          x-kubernetes-map-type: atomic
+                        resourceFieldRef:
+                          properties:
+                            containerName:
+                              type: string
+                            divisor:
+                              anyOf:
+                              - type: integer
+                              - type: string
+                              pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                              x-kubernetes-int-or-string: true
+                            resource:
+                              type: string
+                          required:
+                          - resource
+                          type: object
+                          x-kubernetes-map-type: atomic
+                        secretKeyRef:
+                          properties:
+                            key:
+                              type: string
+                            name:
+                              default: ""
+                              type: string
+                            optional:
+                              type: boolean
+                          required:
+                          - key
+                          type: object
+                          x-kubernetes-map-type: atomic
+                      type: object
                   required:
                   - name
                   type: object
-                  x-kubernetes-preserve-unknown-fields: true
                 type: array
               extraEnvsFrom:
                 items:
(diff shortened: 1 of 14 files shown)


vrutkovs:
>kubeconform fails with "additional properties 'valueFrom' not allowed":

This is a [known kubeconform issue](https://github.com/yannh/kubeconform/issues/224), see #1396

monotek:
But it still could be fixed by completing the crd schema here.

vrutkovs:
Yes, but that would increase the CRD size and it would no longer fit in max secret / configmap size - that is crucial for Helm. This was the reason why this solution was rejected in the linked issue - and things only got worse since then.

Also, aside from kubeconform errors and missing admission validation, I don't see any effect on end-user

monotek:
But this crd is a multi resource yaml file. 
In etcd they should be split up, or not?
So in the end the single crds resources are far away from 1mb?

vrutkovs:
No, etcd will not be auto-spliting (and reassembling it) - default `max-request-size` is a hard ceiling here. This is even after we somehow smuggled this CRD in the configmap for Helm

monotek:
So one would need to split up the crd in multiple files to work around it?
Whats the reason the large file is kept?

vrutkovs:
etcd doesn't support any kind of reassembly - each object needs to be smaller than the max size. Several common approaches to tackle that here are:
* Instead of extending existing object we introduce a new object and refer to it. The tradeoff here are constant API breaks to accomodate new object support
* Remove unnecessary parts like descriptions. Tradeoff - poor documentation
* Use preserve-unknown-fields to minimize the size. Tradeoff - missing validation on API level, but it can be complimented with a custom admission webhook

monotek:
But the single CRDs are way below 1 MB.
Are we talking about the Helm issue, that all resources are going to be in a single secret?
That an issue of cause. In our case we always handle CRDs outside of Helm.

vrutkovs:
We are dangerously close to 1 MB:
```
$ du -cksh config/crd/bases/*.yaml | sort -hr | head
940K	config/crd/bases/operator.victoriametrics.com_vmclusters.yaml
912K	config/crd/bases/operator.victoriametrics.com_vlclusters.yaml
908K	config/crd/bases/operator.victoriametrics.com_vtclusters.yaml
```

This is the size with descriptions removed and `x-kubernetes-preserve-unknown-fields` in several places

monotek:
To bad :(
Thanks for the explanations :)

Issue #2487, opened 2026-08-09
VMCluster: serviceSpec.useAsDefault cannot change vmselect Service from headless to ClusterIP


Barakkol33:
## What happens

For components whose default Service is headless (vmselect, vmstorage, vmalertmanager), `serviceSpec.useAsDefault`: true silently keeps `clusterIP: None` no matter what is supplied in spec. There is no way to turn the main Service into a normal ClusterIP Service.

## Why

build.Service re-inherits the default's clusterIP after replacing the spec:

https://github.com/VictoriaMetrics/operator/blob/v0.66.1/internal/controller/operator/factory/build/service.go#L89-L91
```
if serviceOverrides.Spec.ClusterIP == "" && serviceOverrides.Spec.Type == svc.Spec.Type {
    serviceOverrides.Spec.ClusterIP = svc.Spec.ClusterIP
}
```
At that point svc.Spec.ClusterIP is already "None", set by the component callback:

https://github.com/VictoriaMetrics/operator/blob/v0.66.1/internal/controller/operator/factory/vmcluster/vmcluster.go#L188-L189
```
svc.Spec.ClusterIP = "None"
svc.Spec.PublishNotReadyAddresses = true
```
Omitting type in the override is not an escape either, because it is defaulted from the base Service three lines earlier, which then makes the guard's second condition true:

https://github.com/VictoriaMetrics/operator/blob/v0.66.1/internal/controller/operator/factory/build/service.go#L86-L88

## Reproduction

```
helm install victoria-metrics-operator vm/victoria-metrics-operator --version 0.67.2 
```

Then kubectl apply this file:
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
metadata:
  name: example
  namespace: infra
spec:
  retentionPeriod: "1"
  vmselect:
    replicaCount: 1
    serviceSpec:
      useAsDefault: true
      spec:
        type: ClusterIP
        ports:
          - name: http
            port: 8481
            targetPort: 8481
  vmstorage:
    replicaCount: 1
```

Result:
```
kubectl get svc vmselect-example -o jsonpath='{.spec.clusterIP}'
```
*Expected*: an allocated cluster IP.

*Actua*l: None.

Changing useAsDefault to false (and giving the override a distinct metadata.name) produces a second Service that does get a VIP — demonstrating the inconsistency.

## Suggested fix

[...]

PR #2491, opened 2026-08-12
VMCluster/VMAlertmanager: reject serviceSpec.type override with useAsDefault

mehrdadbn9:
## What

Follows up on #2487. As discussed in the issue, the default Service of `vmselect`, `vmstorage` and `vmalertmanager` is intentionally headless (`clusterIP: None`) because it also exposes cluster-native / gossip ports. Because of that, combining `serviceSpec.useAsDefault: true` with an explicit `spec.type` silently produces a service that is still headless - the user gets no error and no VIP, which is confusing.

Per the maintainer's suggestion in https://github.com/VictoriaMetrics/operator/issues/2487#issuecomment-5258765657 (prohibit overriding the service type for these components when `useAsDefault: true`), this PR makes the operator reject that combination with a clear error message instead of silently ignoring it.

## Changes

- `AdditionalServiceSpec.ValidateNoServiceTypeOverrideWithUseAsDefault(component)` - new helper that rejects an explicit `spec.type` when `useAsDefault` is set.
- Wired into `VMCluster.Validate()` for `vmselect` and `vmstorage`.
- Wired into `VMAlertmanager.Validate()` for `vmalertmanager`.
- `vminsert` is intentionally left allowed: its default Service is not headless, so an explicit type override there works as expected.

Users who need a different service type can still create a separate additional Service by setting `serviceSpec.useAsDefault: false` with a distinct `metadata.name`.

## Tests

Added unit test cases covering: reject for vmselect/vmstorage/vmalertmanager, allow when no explicit type is set, allow separate additional services, and allow type override for vminsert.

```
go build ./...          # OK
go vet ./operator/v1beta1/  # OK
go test ./operator/v1beta1/  # ok (full package)
```

Feedback welcome - happy to adjust the error message wording or the scope if you'd rather also cover `vminsert`.


api/operator/v1beta1/vmextra_types.go
```diff
@@ -396,6 +396,25 @@ func (asc *AdditionalServiceSpec) IsSomeAndThen(cb func(s *AdditionalServiceSpec
...
 
+// ValidateNoServiceTypeOverrideWithUseAsDefault rejects an explicit service type
+// combined with useAsDefault for components whose default service is headless
+// (vmselect, vmstorage, vmalertmanager). Those services must stay headless
+// (clusterIP: None) for cluster-native communication: an explicit type: ClusterIP
+// would be silently ignored (the headless clusterIP is inherited from the default
+// service), while type: LoadBalancer/NodePort would be honored and produce a
+// non-headless service that breaks cluster-native discovery. Reject the combination
+// loudly so users pick an intended setup: drop the type override, or create a
+// separate additional service (useAsDefault=false with a distinct metadata.name).
```
AndrewChubatiuk: please make this comment brief, it's a trivial check not need to have this multiline explanantion

api/operator/v1beta1/vmextra_types.go
```diff
@@ -396,6 +396,25 @@ func (asc *AdditionalServiceSpec) IsSomeAndThen(cb func(s *AdditionalServiceSpec
...
+func (asc *AdditionalServiceSpec) ValidateNoServiceTypeOverrideWithUseAsDefault(component string) error {
+	if asc == nil || !asc.UseAsDefault {
+		return nil
+	}
+	if asc.Spec.Type != "" {
+		return fmt.Errorf("serviceSpec.useAsDefault cannot be combined with an explicit spec.type=%q for %s: the default service is headless (clusterIP: None) and must stay headless for cluster-native communication (an explicit type is either silently ignored or produces a non-headless service that breaks it). Remove spec.type from the serviceSpec override, or create a separate service by setting serviceSpec.useAsDefault=false with a distinct metadata.name", asc.Spec.Type, component)
```
AndrewChubatiuk: same is here, additionally no need to pass component name here,

AndrewChubatiuk (changes requested):
Thanks for a PR! Overall it's good, left some comments. Same changes probably should be propagated to all select and storage components as they also unconditionally set ClusterIP type

mehrdadbn9:
Thanks @AndrewChubatiuk — addressed the feedback:

- Shortened the helper GoDoc/rationale (it was a trivial check, no need for the multiline explanation).
- Removed the `component` parameter from `ValidateNoServiceTypeOverrideWithUseAsDefault()`; the error message no longer names a specific component.
- Corrected the rationale: the default service must stay headless (`clusterIP: None`) for cluster-native communication; an explicit `spec.type` with `useAsDefault=true` is rejected (per #2487). (The earlier wording about "silently ignored" was inaccurate — `build.Service` does honor an explicit type override.)
- Applied to `vmselect`, `vmstorage` (VMCluster) and `vmalertmanager`.

On propagation to **all** select/storage components: I deliberately kept **vminsert** allowed, because its default Service is *not* headless (unlike vmselect/vmstorage/vmalertmanager), so an explicit type override there works as intended and shouldn't be rejected. If you'd still like vminsert covered (or the standalone CRDs like VMAgent/VMAlert/VMSingle), let me know and I'll extend it.

Force-pushed `84715de`. Full `api` module test suite passes.

vrutkovs (approved):

AndrewChubatiuk (approved):
LGTM!

PR #2522, opened 2026-08-20
fix: require k8s >= 1.30 for preStop Sleep lifecycle hook

vrutkovs:
PreStop hook feature gate is still Alpha in 1.29, so its best to avoid attempting to use native sleep


AndrewChubatiuk (approved):

Haleygo:
See follow up in https://github.com/VictoriaMetrics/operator/commit/fcec48a8f38a78cd0eb89ba2c54ceca58889e56f.

PR #2523, opened 2026-08-20
test(e2e): clear explicit ServiceSpec.Type when enabling UseAsDefault

vrutkovs:
In e2e tests we should clear up service type, otherwise UseAsDefault conflicts with previously set service type ClusterIP - a new validation was added to prevent that.

This fixes failure in e2e tests:
* `[FAIL] e2e vmcluster update should update exist cluster [It] by deleting deleting and renaming additional services [vm, cluster, vmcluster]`


AndrewChubatiuk:
let's rather update validation to 

```
func (asc *AdditionalServiceSpec) ValidateNoServiceTypeOverrideWithUseAsDefault() error {
	if asc == nil || !asc.UseAsDefault {
		return nil
	}
	if asc.Spec.Type != "" && asc.Spec.Type != corev1.ServiceTypeClusterIP {
		return fmt.Errorf("serviceSpec.useAsDefault cannot be combined with an explicit spec.type=%q: the default service is headless (clusterIP: None) and must stay headless. Remove spec.type, or create a separate service with serviceSpec.useAsDefault=false and a distinct metadata.name", asc.Spec.Type)
	}
	return nil
}
```

ClusterIP is also a valid value

vrutkovs:
Sure, lets have both? Or we're worried that this may hide other cases in the future?

AndrewChubatiuk (approved):

PR #2528, opened 2026-08-21
config-reloader: trigger reload after initial directory sync

### LINKED ISSUE #2527: vmalert: rules stay stale forever when the rules ConfigMap is updated during pod startup
### Describe the bug

  Since v0.72.0, VMAlert reads its rule files from an emptyDir (`rules-out`), populated by the `config-init` init container and synced by the `config-reloader` sidecar.

  If the ConfigMap is updated between the init container's copy and the sidecar starting, VMAlert loads the stale copy and is never reloaded:

  1. `config-init` copies snapshot A to `rules-out`.
  2. The kubelet refreshes the volume to snapshot B.
  3. `vmalert` starts and reads `rules-out` — still snapshot A.
  4. `config-reloader` starts; its initial sync copies snapshot B to `rules-out`, but the initial sync never sends a reload signal (`dirWatcher.start()` in `cmd/config-reloader/file_watch.go` only signals from the fsnotify event loop).

  If the ConfigMap never changes again, no fsnotify event ever fires and VMAlert serves snapshot A forever, while snapshot B sits on its own disk. There is no periodic fallback: `--resync-interval` only applies to the config-secret file watcher, not to rule directories. Before v0.72.0 this could not happen, because vmalert read the ConfigMap mount directly.

 This race is easy to hit when many VMRule objects are applied in bulk (e.g. by a GitOps sync)  while the VMAlert deployment performs its first rollout.

  ### Environment

  operator v0.73.1

  ### Proposed fix

  Send one reload signal after the initial sync in `dirWatcher.start()`. A reload is idempotent and goes through `--delay-interval`, so an unconditional signal at startup is cheap and makes the loaded state converge to the on-disk state regardless of startup ordering.

shunki-fujita:
Fixes #2527

### Problem

Since v0.72.0, VMAlert reads its rule files from an emptyDir populated by the `config-init` init container and synced by the `config-reloader` sidecar. If the rules ConfigMap is updated between the init container's copy and the sidecar starting, the sidecar's initial sync writes the new content to the target dir but never notifies the application. The application keeps serving the snapshot it read at startup, and if the ConfigMap never changes again, no fsnotify event ever fires — the stale state is permanent.

### Fix

- `dirWatcher.start()` sends one reload signal after the initial directory sync, provided at least one watched directory completed it. This also applies to watch-only directories (no `--target-dir`): the same startup race exists there — a change landing between the application's first read and the watch being established produces no event — just with a smaller window.
- Failed reload calls are retried with exponential backoff (1s, doubling up to 30s) until they succeed. Without this, the startup reload could be lost when the application's reload endpoint is not listening yet, reintroducing the same permanent staleness. This also covers transient failures of event-driven reloads.
- The directory content hash is committed to the cache only after a successful sync (suggested by @vrutkovs). A failed sync — at startup or on an event — leaves the hash uncommitted, so the next fsnotify event retries it naturally, without a dedicated retry loop. This also fixes a pre-existing gap where a failed event-driven sync was never retried, because the hash had already been cached.

### Tests

- reload retry: the reload call fails twice and succeeds on the third attempt
- failed initial sync: no startup reload fires, and the next event retries the sync and triggers the reload even for unchanged content, since the hash was never committed
- partial failure: a failing pair does not hold back the startup reload for a pair that synced successfully
- liveness: events for healthy directories are processed while another pair's sync keeps failing
- no watched dirs: no startup reload
- existing dir-watcher tests drain the startup signal, pinning the new behavior

### DIFF
--- a/cmd/config-reloader/file_watch.go
+++ b/cmd/config-reloader/file_watch.go
@@ -301,11 +301,35 @@ func (dw *dirWatcher) start(ctx context.Context, updates chan struct{}) {
 			logger.Infof("triggering reload after initial sync")
 			triggerReload()
 		}
-		retryInitialSync(ctx, pending, triggerReload)
+		// Pairs whose initial sync failed are retried with backoff, multiplexed
+		// with event processing so a failing pair never blocks the event loop.
+		backoff := initialSyncRetryBackoff
+		var retryC <-chan time.Time
+		if len(pending) > 0 {
+			retryC = time.After(backoff)
+		}
 		for {
 			select {
 			case <-ctx.Done():
 				return
+			case <-retryC:
+				var failed []dirPair
+				for _, p := range pending {
+					if err := p.sync(); err != nil {
+						logger.Errorf("cannot copy dir %s to target: %s", p.src, err)
+						failed = append(failed, p)
+					}
+				}
+				if len(failed) < len(pending) {
+					triggerReload()
+				}
+				pending = failed
+				if len(pending) > 0 {
+					backoff = min(backoff*2, 30*time.Second)
+					retryC = time.After(backoff)
+				} else {
+					retryC = nil
+				}
 			case event := <-dw.w.Events:
 				baseDir := filepath.Dir(event.Name)
 				logger.Infof("dir update: base dir: %s", baseDir)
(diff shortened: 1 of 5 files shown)


cmd/config-reloader/file_watch.go
```diff
@@ -275,20 +275,61 @@ func (dw *dirWatcher) start(ctx context.Context, updates chan struct{}) {
...
 		}
 	}
 	go func() {
 		defer dw.wg.Done()
+		if len(dw.pairs) > 0 {
+			// The application may have already read stale content before the
+			// watcher was established, or before the initial sync replaced it.
+			// Without further source changes, no fsnotify event is guaranteed,
+			// so trigger one reload after startup.
+			logger.Infof("triggering reload after initial sync")
```
vrutkovs:
```suggestion
			// Reload here to catch changes between container start and initial sync
			logger.Infof("triggering reload after initial sync")
```
shorter comments please

vrutkovs (changes requested):

cmd/config-reloader/file_watch.go
shunki-fujita: [660c19e](https://github.com/VictoriaMetrics/operator/pull/2528/commits/660c19e1a8e287f37f1aeddad13aa9074dd5db28)

cmd/config-reloader/file_watch.go
vrutkovs: My understanding is that we add the directory to the cache first time anyway, even if the sync succeeds? Perhaps we should change that assumption - and that would help us get rid of "infinite cycle to handle initial sync"?

We can update `updateCache` function to return `newHash` and set `filesContentHashPath[baseDir] = newHash` here

vrutkovs (changes requested):
Needs CHANGELOG entry too

cmd/config-reloader/file_watch.go
shunki-fujita: Agreed, that's a better design.
Committing the hash only after a successful sync makes the retry loop unnecessary, and it also covers a case the loop didn't (a failed sync in the event path was never retried).

cmd/config-reloader/file_watch.go
```diff
@@ -263,36 +265,51 @@ func (dw *dirWatcher) start(ctx context.Context, updates chan struct{}) {
...
+			return newHash, false, nil
 		}
-		filesContentHashPath[eventPath] = newHash
 		logger.Infof("base dir: %s hash not the same, update needed", eventPath)
-		return true, nil
+		return newHash, true, nil
 	}
 	for _, p := range dw.pairs {
-		if _, err := updateCache(p.src); err != nil {
+		newHash, _, err := updateCache(p.src)
```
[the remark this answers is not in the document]
shunki-fujita: This is a known trade-off: the retry loop was removed deliberately per the review above — a failed sync leaves the hash uncommitted, so any later source event retries it. The remaining gap (sync fails at startup and the source never changes again) requires an I/O error on a writable emptyDir that then heals without any source change, which we consider unlikely enough to not warrant the extra loop.

vrutkovs (approved):

Issue #2495, opened 2026-08-13
Runtime failure signals


AquaOctet:
## Change:
Pod Ready means that the Operator process is running. It does not mean that every controller is making progress.
Pod readiness is not a substitute for controller signals. We want to add the separate signals below:
- add the existing `operator_object_status{status="failed"}` metric to the dashboard
- add the existing `workqueue_longest_running_processor_seconds` metric to the dashboard and ship a stalled-reconcile alert
- add the existing `configreloader_k8s_watch_errors_total` and  `configreloader_last_reload_errors_total` metrics to the dashboard and alerts
- add `operator_cache_watch_errors_total{type,reason}` through controller-runtime `DefaultWatchErrorHandler`
- remove the stale operator_prometheus_converter_watch_events_total query from `dashboards/operator.json` in the VictoriaMetrics repository

> Do not alert only because reconcile count is zero. An idle controller is valid.
> >Note: that pod readiness is itself currently unreliable: the chart's `readinessProbe` targets `/health`, a no-op endpoint, so readiness today reflects only that the HTTP server is listening.

## Done when: 
The NetworkPolicy reproduction fires a stalled-reconcile or cache-watch alert while the pod remains Ready. The config-reloader permission test fires a sidecar alert. A failed CR appears in the object-status panel. Healthy idle controllers do not alert.

### Product result: 
Operators see confirmed reconcile and config-reload failures before users report missing changes.

### Failure Coverage:
<meta charset="utf-8" /><b style="font-weight: normal" id="docs-internal-guid-ee7553dc-7fff-d6d3-9d1d-d7d9ec1670f1"><div dir="ltr" style="margin-left: -51pt" align="left"></div></b>
Failure | Covered by | Required result
-- | -- | --
Package parts do not match | Tested release record; Full-package pre-release tests | The release record names the exact package. The same package passes the test.
API, RBAC, or cache cannot sync | Full-package pre-release tests; Upgrade dry-run check; Runtime failure signals | The test fails. Dry-run reports the exact error or timeout. Runtime monitoring alerts if it still happens.
A child watch is missing or stops | Full-package pre-release tests; Runtime failure signals | The test changes a child object and verifies reconcile and watch recovery. Runtime monitoring reports a watch error.
Upgrade restarts a managed workload unexpectedly | Full-package pre-release tests | The test fails on an unexpected restart or availability loss.
[...]

PR #2536, opened 2026-08-23
operator: added single operator_controller_errors_total metric with reason labels for all reconcile errors

AndrewChubatiuk:
related issue https://github.com/VictoriaMetrics/operator/issues/2495

Currently operator metrics track only certain reconcile errors also `operator_controller_reconcile_errors_total` despite it's name tracks only cancel context errors. Added `operator_controller_errors_total` with `controller`, `namespace`, `name` and `reason` labels as a replacement for all per-error type reconcile error counter and additionally introduced `reason="other"` to track other errors, which are not tracked at the moment

### DIFF
--- a/internal/controller/operator/controllers.go
+++ b/internal/controller/operator/controllers.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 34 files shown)


internal/controller/operator/controllers.go
```diff
@@ -144,63 +187,59 @@ func handleReconcileErrWithStatus[T client.Object, ST reconcile.StatusWithMetada
...
-				UID:             object.GetUID(),
-				ResourceVersion: object.GetResourceVersion(),
-			},
-		}
-		if err := rclient.Create(ctx, errEvent); err != nil {
-			logger.WithContext(ctx).Error(err, "failed to create error event at kubernetes API during reconciliation error")
-		}
+
+	// object is guaranteed to have identity here: the early return above already filtered out objects without one.
+	errEvent := &corev1.Event{
```
vrutkovs, line 222: Lets check that we have permissions to create events - adding a kubecontroller annotation here too would be useful

vrutkovs (approved):

Issue #1181, opened 2024-12-05
Add resources health checks to the argocd repo


f41gh7:
argocd allows to bundle health checks for custom resources in order to track it's health status.

https://argo-cd.readthedocs.io/en/latest/operator-manual/health/#way-2-contribute-a-custom-health-check
https://github.com/argoproj/argo-cd/tree/master/resource_customizations

 It'd be great to have track for our `status` resource field, that exposes actual state of the applications. 

An example of lua script:

```lua
local hs = {}
if obj.status ~= nil then
	if obj.status.updateStatus ~= "operational" then
		hs.status = "Progressing"
		hs.message = obj.status.reason
		return hs
	end
	hs.status = "Healthy"
	hs.message = ""
	return hs
end
hs.status = "Progressing"
hs.message = "actual status is unknown"
return hs
```

AndrewChubatiuk:
depends on https://github.com/VictoriaMetrics/operator/issues/1155

megative:
Hi there, had a couple of free hours, and noticed this issue.

Created a PR here: https://github.com/argoproj/argo-cd/pull/27451

megative:
seems like it stuck a bit, I'll update here as soon as the review is passed

AndrewChubatiuk:
thank you!

megative:
@vrutkovs @AndrewChubatiuk @f41gh7 hi guys, cloud you please take a look at this one? https://github.com/argoproj/argo-cd/pull/27451

The reviewers on ArgoCD side seems to be satisfied with changes but they want some clarification if that fits the project needs.

megative:
It's merged now!

AndrewChubatiuk:
hey @megative 
thanks a lot! will label this issue as `waiting for release` till it's available in ArgoCD releases

AndrewChubatiuk:
available in argocd v3.5.0-rc1

piodyb:
The health check from #1181 was merged upstream in argoproj/argo-cd#27451 and is now shipped in ArgoCD v3.5.0+ (bundled `resource_customizations/operator.victoriametrics.com/_/health.lua`).

[...]

PR #2539, opened 2026-08-24
operator: set status for child objects if no parent objects exist

AndrewChubatiuk:
Change should address comment https://github.com/VictoriaMetrics/operator/issues/1181#issuecomment-5394029146


internal/controller/operator/vmalertmanagerconfig_controller.go
```diff
@@ -133,7 +134,10 @@ func (r *VMAlertmanagerConfigReconciler) Reconcile(ctx context.Context, req ctrl
 			return nil
 		})
 	}
-	err = g.Wait()
+	if err = g.Wait(); err != nil {
```
vrutkovs: This would return just the parent status on `err`, ignoring child status, imo
```
err = g.Wait()
if syncErr := reconcile.SyncAggregatedChildStatus(ctx, r.Client, &instance); syncErr != nil && err == nil {
	err = syncErr
}
return
```
would include both parent and child errors correctly

vrutkovs (changes requested):

internal/controller/operator/factory/vmalertmanager/statefulset.go
```diff
@@ -464,6 +464,10 @@ func getAssetsCache(ctx context.Context, rclient client.Client, cr *vmv1beta1.VM
 // CreateOrUpdateConfig - check if secret with config exist,
 // if not create with predefined or user value.
 func CreateOrUpdateConfig(ctx context.Context, rclient client.Client, cr *vmv1beta1.VMAlertmanager, childCR *vmv1beta1.VMAlertmanagerConfig) error {
+	if cr.IsUnmanaged() {
```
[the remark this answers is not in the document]
AndrewChubatiuk: Confirmed and fixed: removed the IsUnmanaged() early return entirely — it was a pure addition by this PR with no pre-existing equivalent, and it skipped Secret creation even for normal alertmanagers that just don't select any VMAlertmanagerConfig (e.g. ConfigRawYaml-only setups). The existing StatusForChildObjects(parent, pos.configs.All()) call at the end of the normal flow already releases stale conditions correctly when the selection is empty, so nothing is lost. Same fix applied to vmanomaly/config.go, which had the identical pattern.

internal/controller/operator/factory/vmalert/rules.go
```diff
@@ -25,9 +25,9 @@ import (
 // Alerting rules are dropped when hasNotifiers is false, since vmalert would have nowhere to
 // send them; recording rules are unaffected.
 func CreateOrUpdateRuleConfigMaps(ctx context.Context, rclient client.Client, cr *vmv1beta1.VMAlert, childCR *vmv1beta1.VMRule, hasNotifiers bool) ([]string, error) {
-	// fast path
 	if cr.IsUnmanaged() {
-		return nil, nil
+		parentObject := fmt.Sprintf("%s.%s.vmalert", cr.Name, cr.Namespace)
+		return nil, reconcile.StatusForChildObjects(ctx, rclient, parentObject, []*vmv1beta1.VMRule(nil))
```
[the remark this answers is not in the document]
AndrewChubatiuk: Confirmed and fixed: guarded the fast-path StatusForChildObjects call with !build.IsControllerDisabled("VMRule"), returning success when VMRule is disabled (its CRD unavailable), matching the existing pattern used for the same disabled-CRD scenario elsewhere.

internal/controller/operator/vmagent_controller.go
```diff
@@ -105,6 +105,10 @@ func (r *VMAgentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 
 	RegisterObjectStat(&instance, r.name)
 	if !instance.DeletionTimestamp.IsZero() {
+		parentObject := fmt.Sprintf("%s.%s.vmagent", instance.Name, instance.Namespace)
+		if err = releaseScrapeChildStatuses(ctx, r.Client, parentObject); err != nil {
```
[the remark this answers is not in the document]
AndrewChubatiuk: Confirmed and fixed: moved the DeletionTimestamp branch ahead of the IsUnmanaged-gated RLock and now take agentSync's full write Lock around releaseScrapeChildStatuses + finalize.OnVMAgentDelete, so no in-flight collectVMAgentScrapes (which takes the same write lock) can re-apply a condition to a child in between. Applied the identical fix to vmsingle_controller.go, which shared the same race via vmsingleSync.

AndrewChubatiuk:
@cubic-ai-dev

internal/controller/operator/factory/reconcile/status.go
```diff
@@ -35,7 +35,7 @@ type objectWithStatus interface {
 }
 
 func childConditionType(parentObjectName string) string {
-	if len(strings.Split(parentObjectName, ".")) != 3 {
+	if len(strings.Split(parentObjectName, ".")) < 3 {
```
vrutkovs, line 39: Is this needed to support cluster-level objects? We should make it stricter then - 2 or 3 dots

internal/controller/operator/vmagent_controller.go
```diff
@@ -166,7 +172,7 @@ func collectVMAgentScrapes(l logr.Logger, ctx context.Context, rclient client.Cl
 		objects.Items = append(objects.Items, dst.Items...)
 	}); err != nil {
 		err = fmt.Errorf("cannot list VMAgents for %T: %w", instance, err)
-		return
+		return err
```
vrutkovs: unnecessary change - this would be a named return anyway

internal/controller/operator/vmalert_controller.go
```diff
@@ -94,6 +97,12 @@ func (r *VMAlertReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 
 	RegisterObjectStat(&instance, r.name)
```
vrutkovs, line 93: While we're changing this here lets make it consistent with other controllers:
* RegisterObjectStat
* DeletionTimestamp check
* IsUnmanaged check

Currently, I think there is a path where we don't release the lock

internal/controller/operator/vmscrapeconfig_controller.go
```diff
@@ -80,11 +81,12 @@ func (r *VMScrapeConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	if err = collectVMAgentScrapes(l, ctx, r.Client, r.BaseConf, &instance); err != nil {
-		return
+	err = collectVMAgentScrapes(l, ctx, r.Client, r.BaseConf, &instance)
+	if singleErr := collectVMSingleScrapes(l, ctx, r.Client, r.BaseConf, &instance); singleErr != nil && err == nil {
+		err = singleErr
 	}
-	if err = collectVMSingleScrapes(l, ctx, r.Client, r.BaseConf, &instance); err != nil {
-		return
+	if err == nil {
+		err = reconcile.SyncAggregatedChildStatus(ctx, r.Client, &instance)
```
vrutkovs: lets collect errors first and aggregate them later:
```
agentErr := collectVMAgentScrapes(l, ctx, r.Client, r.BaseConf, &instance)
singleErr := collectVMSingleScrapes(l, ctx, r.Client, r.BaseConf, &instance)
errs := []error{agentErr, singleErr}
if agentErr == nil && singleErr == nil {
	errs = append(errs, reconcile.SyncAggregatedChildStatus(ctx, r.Client, &instance))
}
err = utilerrors.NewAggregate(errs)
```
What do you think?  Since it repeats quite often lets move this to a util pkg

internal/controller/operator/factory/reconcile/status.go
AndrewChubatiuk: name can contain dot (4 dots in this case), this may crash operator

internal/controller/operator/vmalert_controller.go
AndrewChubatiuk: updated

internal/controller/operator/vmscrapeconfig_controller.go
AndrewChubatiuk: done

internal/controller/operator/vmalertmanagerconfig_controller.go
```diff
@@ -83,6 +85,7 @@ func (r *VMAlertmanagerConfigReconciler) Reconcile(ctx context.Context, req ctrl
 		return
 	}
 	if alertmanagerReconcileLimit.Throttle() {
+		err = reconcile.SyncAggregatedChildStatus(ctx, r.Client, &instance)
 		return
 	}
```
vrutkovs, line 91: Needs `if !instance.IsUnmanaged(nil) {`

internal/controller/operator/vmanomalyconfig_controller.go
vrutkovs, line 91: Needs `if !instance.IsUnmanaged(nil) {` section

internal/controller/operator/vmalertmanagerconfig_controller.go
AndrewChubatiuk: isUnmanaged is for parent objects, not for child as in this case

internal/controller/operator/vmanomalyconfig_controller.go
AndrewChubatiuk: IsUnmanaged is to identify if parent object can manage exact child object, this is not applicable here

vrutkovs (approved):

Haleygo:
Closing in favor of https://github.com/VictoriaMetrics/operator/pull/2650, see detailed reason in https://github.com/VictoriaMetrics/operator/issues/2649.

PR #2540, opened 2026-08-24
fix: scope RBAC namespaces per reconciler

vrutkovs:
Use each reconciler's watched namespaces for VMAgent and VMSingle RBAC creation and orphan cleanup. Prevent shared global config from creating or deleting RBAC in unrelated namespaces.

### DIFF
--- a/internal/controller/operator/factory/vmagent/rbac.go
+++ b/internal/controller/operator/factory/vmagent/rbac.go
@@ -121,24 +121,6 @@ func createK8sAPIAccess(ctx context.Context, rclient client.Client, cr, prevCR *
 	return nil
 }
 
-func rbacNamespaces(cr *vmv1beta1.VMAgent, watchedNamespaces []string) []string {
-	if len(watchedNamespaces) == 0 {
-		return nil
-	}
-
-	if cr.Spec.SelectAllByDefault ||
-		cr.Spec.ServiceScrapeNamespaceSelector != nil ||
-		cr.Spec.PodScrapeNamespaceSelector != nil ||
-		cr.Spec.ProbeNamespaceSelector != nil ||
-		cr.Spec.NodeScrapeNamespaceSelector != nil ||
-		cr.Spec.StaticScrapeNamespaceSelector != nil ||
-		cr.Spec.ScrapeConfigNamespaceSelector != nil {
-		return watchedNamespaces
-	}
-
-	return []string{cr.Namespace}
-}
-
 func ensureCRExist(ctx context.Context, rclient client.Client, cr, prevCR *vmv1beta1.VMAgent) error {
 	var prevClusterRole *rbacv1.ClusterRole
 	if prevCR != nil {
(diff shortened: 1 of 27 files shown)


AndrewChubatiuk (approved):

vrutkovs:
Lets not merge it yet though - I think there are few more bugs around this

internal/controller/operator/factory/vmagent/rbac.go
```diff
@@ -118,6 +118,24 @@ func createK8sAPIAccess(ctx context.Context, rclient client.Client, cr, prevCR *
 	return nil
 }
 
+func rbacNamespaces(cr *vmv1beta1.VMAgent, watchedNamespaces []string) []string {
```
AndrewChubatiuk: looks like a part of API package

vrutkovs:
Tested this in e2e tests - seems to help with resources cleanup

internal/controller/operator/factory/vmagent/rbac.go
AndrewChubatiuk: this function also makes sense for vmsingle

AndrewChubatiuk: also looks like similar functions makes sense for other parent objects

Haleygo:
This change broke a case where I'm uncertain what the expected result should be:
In multi-namespace mode, the operator grants discovery RBAC only in the CR's own namespace unless a `NamespaceSelector` or `selectAllByDefault` is set — but `vmagent` still selects scrape objects from every watched namespace regardless of selectors. As a result, a CR with only an object selector still gets scrape jobs pointing at other namespaces, with no `Role` backing them.

For example, run the operator with `WATCH_NAMESPACE=agent-ns,scrape-ns`, with a `VMAgent` in `agent-ns` that only has an object selector:

```yaml
spec:
  podScrapeSelector:
    matchLabels: {select: "me"}   # no podScrapeNamespaceSelector, no selectAllByDefault
```

and a `VMPodScrape` labelled `select: me` sitting in `scrape-ns`:

```yaml
kind: VMPodScrape
metadata:
  name: ps
  namespace: scrape-ns
  labels:
    select: me
spec:
  selector:
    matchLabels:
      app: test
  podMetricsEndpoints:
    - port: metrics
```

The `VMPodScrape` is still selected by `vmagent` and generated into the scrape config, but `vmagent` fails to scrape it because it cannot list resource `pods` in `scrape-ns`.

This case was working before this pull request. However, [our docs](https://docs.victoriametrics.com/operator/resources/vmagent/) state:

image

So I'm uncertain whether this was a bug that is now partially fixed by this pull request — in which case the complete fix should also prevent `vmagent` from selecting scrape objects outside its own namespace.
Or whether we should instead grant `vmagent` permissions across all namespaces listed in `WATCH_NAMESPACE`.

@vrutkovs @AndrewChubatiuk Do you have an answer?

vrutkovs:
I'd prefer an explicit `NamespaceSelector` / `selectAllByDefault` for vmagent for it to look in other namespaces - same as the docs require. Debugging the VMAgent behaviour would no longer depend on (unrelated) operator change.

This is something our users may have misused and now is a good opportunity to adjust vmagent to match docs fully.

PR #2541, opened 2026-08-24
operator: set Failed status to CR during attempt to decrease disk size

### LINKED ISSUE #2512: VMAgent reports `updateStatus: operational` when a requested PVC decrease was not applied
When a claim template requests a smaller storage size than the bound PVC, the operator correctly declines to act, because Kubernetes has no in-place shrink of a bound volume.
The bug is what the operator reports afterwards. It logs one `info` line and then presents the reconcile as fully successful:

- `status.updateStatus` is set to `operational`
- `status.reason` is explicitly cleared
- `status.observedGeneration` is advanced to the current `metadata.generation`
- a Normal event is emitted reading `reconcile of object finished successfully`

There is no condition, no reason, and no event describing the change that was not applied. The resource is knowingly diverged from its own spec while reporting healthy, and nothing queryable reflects that, so the state cannot be alerted on and no consumer can detect it.

We ran in this state for some period of time and only noticed because someone happened to read operator logs while investigating something unrelated.
Observed on **v0.68.4**. The code below is byte for byte identical at current head (**v0.74.1**), so this is not fixed in a newer release.
Environment: EKS 1.34, `ebs.csi.aws.com`, StorageClass with `allowVolumeExpansion: true`, `VMAgent` with `statefulMode: true`.

`modifyPVC` returns a `nil` error for this case, so the reconcile is indistinguishable from a clean one:

```go
// internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go:173
if direction < 0 {
	l.Info(fmt.Sprintf("cannot decrease PVC=%s size from=%s to=%s, please check VolumeClaimTemplate configuration", ...))
	return metaChanged, nil        // :177
}
```

The `waitForPVCReady` call that follows cannot catch it either, because it only polls while the volume is *smaller* than requested:

```go
// internal/controller/operator/factory/reconcile/pvc.go:73
actualSize := pvc.Status.Capacity[corev1.ResourceStorage]
if actualSize.Cmp(size) < 0 {
	return false, nil
}
```

With `actual = 1440G` and `size = 72G`, `Cmp` returns `> 0`, so the branch is skipped and it returns `true, nil`. It is a readiness check rather than an equality check, so an over-sized volume looks correct.
The operator knows it did not apply part of the spec, so that should be visible in the API and not only in logs. Any of the following would help, roughly in order of usefulness:

1. Set a `Degraded` condition describing the declined change. `StatusMetadata` already declares `Conditions []Condition` and documents `Available`, `Progressing` and `Degraded`, and helpers exist at `status.go:103` (`applyChildStatusCondition`) and `status.go:301` (`setConditionTo`).
2. Emit a Warning event on the object via `createGenericEventForObject` at `controllers.go:282`.
3. Populate `status.reason`.
4. Do not advance `observedGeneration` past a generation that was not fully applied.
5. Log at `warn` rather than `info`.

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/2512


internal/controller/operator/factory/reconcile/statefulset_pvc_expand_test.go
```diff
@@ -832,7 +832,7 @@ func Test_updateSTSPVC(t *testing.T) {
 		},
 	})
 
-	// PVC bigger than VCT (manual expansion case), should log info and skip
+	// PVC bigger than VCT (manual expansion case), should error and skip
```
vrutkovs, line 835: Lets add a test with multiple storages too - iiuc when we're unable to downscale one storage the rest will be blocked too (we should attempt to scale all of them and aggregate the resulting error)

vrutkovs: Also, would this error be retried on the next reconciliation?

vrutkovs (changes requested):

internal/controller/operator/factory/reconcile/statefulset_pvc_expand_test.go
vrutkovs: Pushed a fix for these comments, please review - I'd really love to include it in the next RC

vrutkovs (approved):

PR #2543, opened 2026-08-24
operator: fix missing RBAC permissions

AquaOctet:
## Description:
For the Operator, I made RBAC Single sourced, and its generated from +kubebuilder:rbac markers instead of being hand maintained, with CI Enforicing it. That surfaced and hopefully fixes 3 missing permissions... It's the precursor to eliminating operator/chart drift.

## Changes:
- Update `config/rbac/role.yaml`:
  - Makes it generated output from the `+kubebuilder:rbac` markers in the source,
    instead of a hand-maintained file, and adds CI to keep it that way.
  - The markers were inert: all 31 blocks sat directly above `func ... Reconcile(`,
    and `+kubebuilder:rbac` is only collected from detached comment blocks, so
    `controller-gen` emitted no RBAC at all.
  - Goes 16 rules to 26. Same 15 apiGroups, no permission lost. Only operational
    change is `pods` gaining `delete` and `patch`, which the code needs at
    `factory/reconcile/statefulset.go:353` and `factory/k8stools/client_utils.go:100`.
- Remove `extensions`
  - `extensions/v1beta1 Ingress` was removed in Kubernetes 1.22
  - the repo pins `k8s.io/api v0.36.3` - Kubernetes 1.36 - and the README supports
    only current versions
  - the operator uses `networkingv1.Ingress` exclusively; `extensions` appeared
    nowhere else in the code

## Fixes:
Bugs on the published kustomize manifests (Helm users unaffected — the chart grants a broader set):
- `pods` was read-only, so `vmalert` rule reload silently stopped working. The
  annotation patch that triggers it failed with `Forbidden` and the error was
  logged rather than returned, so reconciliation reported success while rules went stale.
- `ingresses` was read-only, so `vmauth` could not create an `Ingress` at all.
- `watch` was missing on `serviceaccounts` and the `rbac.authorization.k8s.io`
  types, which are always read through the informer cache — a denied WATCH hangs
  rather than returning 403.


internal/controller/operator/vlcluster_controller.go
```diff
@@ -60,6 +60,7 @@ func (r *VLClusterReconciler) Init(name string, rclient client.Client, l logr.Lo
 // +kubebuilder:rbac:groups=operator.victoriametrics.com,resources=vlclusters/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=operator.victoriametrics.com,resources=vlclusters/finalizers,verbs=update
 // +kubebuilder:rbac:groups="networking.k8s.io",resources=networkpolicies,verbs=*
+
```
AndrewChubatiuk, line 63: today found that here and in may other places hpa and vpa permissions are also missing

AquaOctet: Is this something we want to add in this PR?

vrutkovs: I'd rather keep it in a separate PR. We're going to need more end-to-end tests to check this anyway, so its easier to limit this PR to rework of existing permissions only

config/rbac/role.yaml
```diff
@@ -27,197 +24,299 @@ rules:
...
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - delete
+  - get
+  - list
+  - patch
+  - watch
```
vrutkovs, line 55: No "update"?

internal/controller/operator/factory/finalize/common.go
```diff
@@ -14,6 +14,12 @@ import (
 	vmv1beta1 "github.com/VictoriaMetrics/operator/api/operator/v1beta1"
 )
 
+// +kubebuilder:rbac:groups="",resources=configmaps/finalizers;persistentvolumeclaims/finalizers;secrets/finalizers;serviceaccounts/finalizers,verbs=*
+// +kubebuilder:rbac:groups="",resources=configmaps/status,verbs=get;list;watch
```
vrutkovs: Configmaps have no status

internal/controller/operator/factory/reconcile/statefulset.go
```diff
@@ -27,6 +27,8 @@ import (
 	"github.com/VictoriaMetrics/operator/internal/controller/operator/factory/logger"
 )
 
+// +kubebuilder:rbac:groups="",resources=pods/eviction,verbs=*
```
vrutkovs:
```suggestion
// +kubebuilder:rbac:groups="",resources=pods/eviction,verbs=create
```
There are no other verbs for eviction yet, but lets keep it scoped to our needs

.github/workflows/main.yaml
```diff
@@ -54,6 +54,10 @@ jobs:
         with:
           sarif_file: "trivy-results.sarif"
 
+      - name: Check rbac markers are not inert
```
vrutkovs: I'd rather move it after lint & build - its useless to do a static analysis of the code which doesn't compile

internal/controller/operator/vmagent_controller.go
```diff
@@ -69,18 +69,20 @@ func (r *VMAgentReconciler) Init(name string, rclient client.Client, l logr.Logg
 // +kubebuilder:rbac:groups=operator.victoriametrics.com,resources=vmagents/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=operator.victoriametrics.com,resources=vmagents/finalizers,verbs=*
 // +kubebuilder:rbac:groups="",resources=nodes,verbs=get;watch;list
-// +kubebuilder:rbac:groups="networking.k8s.io",resources=ingresses,verbs=get;watch;list
+// +kubebuilder:rbac:groups="networking.k8s.io",resources=ingresses,verbs=get;list;watch;create;update;patch;delete
```
vrutkovs:
```suggestion
// +kubebuilder:rbac:groups="networking.k8s.io",resources=ingresses,verbs=get;list;watch
```

Only VMAuth modifies ingresses

internal/controller/operator/vmsingle_controller.go
```diff
@@ -75,16 +75,18 @@ func (r *VMSingleReconciler) Scheme() *runtime.Scheme {
 // +kubebuilder:rbac:groups=operator.victoriametrics.com,resources=vmsingles/finalizers,verbs=*
 // +kubebuilder:rbac:groups="",resources=nodes,verbs=get;watch;list
 // +kubebuilder:rbac:groups="",resources=nodes/metrics,verbs=get;watch;list
-// +kubebuilder:rbac:groups="networking.k8s.io",resources=ingresses,verbs=get;watch;list
+// +kubebuilder:rbac:groups="networking.k8s.io",resources=ingresses,verbs=get;list;watch;create;update;patch;delete
```
vrutkovs:
```suggestion
// +kubebuilder:rbac:groups="networking.k8s.io",resources=ingresses,verbs=get;list;watch
```

internal/controller/operator/vmauth_controller.go
```diff
@@ -69,9 +69,11 @@ func (r *VMAuthReconciler) Scheme() *runtime.Scheme {
 // +kubebuilder:rbac:groups=operator.victoriametrics.com,resources=vmauths,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=operator.victoriametrics.com,resources=vmauths/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=httproutes,verbs=get;list;watch;create;update;patch;delete
```
vrutkovs, line 71:
```suggestion
// +kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=httproutes,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups="networking.k8s.io",resources=ingresses,verbs=get;list;watch;create;update;patch;delete
```

internal/manager/manager.go
```diff
@@ -62,6 +62,8 @@ import (
 	webhookv1beta1 "github.com/VictoriaMetrics/operator/internal/webhook/operator/v1beta1"
 )
 
+// +kubebuilder:rbac:groups=apiextensions.k8s.io,resources=customresourcedefinitions,verbs=get;list
```
vrutkovs: I don't think this is needed?

vrutkovs (commented):
Not sure why inert finder script fails on CI - it passes locally

.github/workflows/main.yaml
AquaOctet: Moved

config/rbac/role.yaml
AquaOctet: I didn't see that Pods called for update.  From what I was able to find, the writes are:
- Delete `factory/reconcile/statefulset.go:355`
- Patch `factory/k8stools/client_utils.go:100` 
- Create on `pods/eviction` at factory/reconcile/statefulset.go:366

I added it anyways for symmetry.  Let me know what you think

vrutkovs:
I've limited this PR to just reconcile blocks working again, I'll address disrepancies found by Cubic in the follow up PR

internal/controller/operator/factory/finalize/common.go
```diff
@@ -14,6 +14,11 @@ import (
 	vmv1beta1 "github.com/VictoriaMetrics/operator/api/operator/v1beta1"
 )
 
+// +kubebuilder:rbac:groups="",resources=configmaps/finalizers;persistentvolumeclaims/finalizers;secrets/finalizers;serviceaccounts/finalizers,verbs=*
+// +kubebuilder:rbac:groups=apps,resources=daemonsets/finalizers;deployments/finalizers;replicasets;statefulsets/finalizers,verbs=*
+// +kubebuilder:rbac:groups=networking.k8s.io,resources=ingresses/finalizers,verbs=*
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings/finalizers;clusterroles/finalizers,verbs=*
```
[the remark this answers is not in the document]
vrutkovs: There are quite a few changes we need to do here, but I'd rather make them in a separate PR (so that it could be reverted safely)

vrutkovs (approved):

PR #2545, opened 2026-08-26
vtsingle/vlsingle: add removePvcAfterDelete field

vrutkovs:
Similar to `VMCluster` add support for `removePvcAfterDelete` so that we could conditionally remove PVCs


AndrewChubatiuk:
since this option was dropped while VLSingle was added as a replacement for VLogs thought it's deprecated

vrutkovs:
I think its useful to have, in any case, it would useful to have this option across `V*Single`

AndrewChubatiuk (approved):

PR #2553, opened 2026-08-29
fix: vmsingle relabelConfig flag reads the asset key

### LINKED ISSUE #2552: VMSingle with relabelConfig crash-loops: asset key relabeling.yaml, flag points at relabel.yaml
A `VMSingle` with `spec.inlineRelabelConfig` (or `spec.relabelConfig`) never starts: the container crash-loops right after opening storage.

```
fatal   VictoriaMetrics/app/vminsert/relabel/relabel.go:35  cannot load relabelConfig:
error when reading -relabelConfig="/etc/vm/relabeling/relabel.yaml":
cannot read "/etc/vm/relabeling/relabel.yaml": open /etc/vm/relabeling/relabel.yaml: no such file or directory
```

The operator writes the relabeling asset under the key `relabeling.yaml`:

```go
const relabelingName = "relabeling.yaml"   // vmsingle.go:40
cm.Data[relabelingName] = string(data)     // buildRelabelingsAssets
```

but builds the flag with a different key:

```go
relabelKeys := []string{"relabel.yaml"}    // vmsingle.go:385
args = build.RelabelArgsTo(args, "relabelConfig", relabelKeys, relabelConfigs...)
```

so the mounted ConfigMap has `relabeling.yaml` while the flag points at `relabel.yaml`.

`VMAgent` gets this right — it passes the same constant (`globalRelabelingName`) to both the asset and the flag.

Fix: use the constant in the flag key as well.

```go
relabelKeys := []string{relabelingName}
```

Operator v0.74.0, VictoriaMetrics v1.148.0, single-node cluster. PR follows.

AdzerKI:
Fixes #2552.

`VMSingle` writes its relabeling asset under `relabeling.yaml`, but builds the container flag with `relabel.yaml`, so any `VMSingle` with `spec.relabelConfig` or `spec.inlineRelabelConfig` crash-loops on start:

```
fatal  cannot load relabelConfig: ... open /etc/vm/relabeling/relabel.yaml: no such file or directory
```

The flag now uses the same constant as the asset, the way `VMAgent` already does with `globalRelabelingName`.


AndrewChubatiuk (approved):
Thanks for contribution! Please sign commits

PR #2572, opened 2026-09-02
fix: preserve prerelease config-reloader version

vrutkovs:
This fixes rc0 behavior, where operator tagged as `v0.75.0-rc0` starts config-reloader with `v0.75.0` tag, which doesn't yet exist. With this fix it would now use `config-reloader-v0.75.0-rc0` tag


internal/config/config.go
```diff
@@ -22,12 +24,16 @@ const (
 	UnlimitedQuantity = "unlimited"
 )
 
+// versionRe matches a semantic version string (vX.Y.Z[-pre]) from buildinfo.Version
+var versionRe = regexp.MustCompile(`v\d+\.\d+\.\d+(?:-[0-9A-Za-z]+(?:[.-][0-9A-Za-z]+)*)?`)
```
AndrewChubatiuk, line 28: if semver is used why versionRe is needed?

vrutkovs: This is a regexp which extracts semver version (`v0.75.0-rc0`) for golang's buildinfo (`operator-20260902-v0.75.0-rc0`)

AndrewChubatiuk (approved):

PR #2585, opened 2026-09-04
fix: derive config-reloader port from http.listenAddr

### LINKED ISSUE #2584: config-reloader ContainerPort ignores configReloaderExtraArgs http.listenAddr (hostNetwork port conflict)
## Summary

The operator always sets `containerPort` to the constant `8435` (`ConfigReloaderDefaultPort` in `internal/controller/operator/factory/build/container.go`) when building the `config-reloader` sidecar. Changing the process listen address with `spec.configReloaderExtraArgs["http.listenAddr"]` updates container args only. **Neither `containerPort` nor the generated HTTP liveness/readiness probes respect the overridden port** — both stay locked to `8435`, so the kube-scheduler still sees a host port conflict even when the process listens elsewhere.

On nodes with `hostNetwork: true`, that blocks scheduling a second DaemonSet that also needs a config-reloader: Kubernetes reports that the node did not have free ports for the requested pod ports (`didn't have free ports…`), even when one process listens on `8435` and another on `8436` via `http.listenAddr`.

`CommonConfigReloaderParams` is shared across `VMAgent`, `VMAuth`, `VMSingle`, `VMAlertmanager`, and `VMAlert`. The bug lives in the common builder `build.ConfigReloaderContainer` and affects any component that uses `configReloaderExtraArgs` to change the listen address.

## Environment

- Operator: v0.74.0 (also reproduced in source of v0.74.1, v0.75.0-rc1, and current `master` as of 2026-09-04)
- Resources: two `VMAgent` DaemonSets with `hostNetwork: true` on the same node
- Sidecar: default VictoriaMetrics config-reloader

## Steps to reproduce

1. Deploy two `VMAgent` CRs as DaemonSets with `hostNetwork: true` (and otherwise valid scrape config).
2. On the second agent, override the reloader listen port, for example:

```yaml
spec:
  hostNetwork: true
  configReloaderExtraArgs:
    http.listenAddr: "127.0.0.1:8436"
```

3. Observe PodSpec of the second DaemonSet: `config-reloader` still has `containerPort: 8435` and probes targeting `8435`, while args include `--http.listenAddr=127.0.0.1:8436`.
4. Scheduler fails to place the second pod on a node that already runs the first agent’s reloader on host port 8435.

## Expected

`ContainerPort` and HTTP probe ports for `config-reloader` should follow the port from `configReloaderExtraArgs["http.listenAddr"]` when set (with fallback to `8435`).

## Actual

Process listens on the overridden address (`127.0.0.1:8436`). PodSpec still declares `containerPort: 8435` and probes `8435`. Scheduler rejects the pod on a node that already exposes host port 8435 for another agent’s reloader.

## Proposed fix

1. Parse the port from `ConfigReloaderExtraArgs["http.listenAddr"]` (support `:PORT`, `IP:PORT`, `[IPv6]:PORT`).
2. Use that port for `ContainerPort` and liveness/readiness probes in `addPortProbesToConfigReloaderContainer` (default `8435` if unset/invalid).
3. Align scrape endpoint builders that hardcode `8435` if they scrape the sidecar by numeric port.
4. Update `configReloaderMetricsURL` in `internal/controller/operator/factory/reconcile/config_reload.go`, which also hardcodes `build.ConfigReloaderDefaultPort` (`8435`). Controllers that call `WaitForConfigReloadHash` (e.g. `VMAuth`) poll `/metrics` on that URL; when the sidecar listen port is overridden, they keep polling `8435` and eventually time out. `configReloaderMetricsURL` should resolve the configured port the same way as `ContainerPort` and probes.
5. Add unit tests in `container_test.go` for default, overridden, IPv6, and invalid values.

## Related

- [VictoriaMetrics/operator#1581](https://github.com/VictoriaMetrics/operator/issues/1581) — IPv6 / `enableTCP6` / `http.listenAddr` interaction (different symptom; related listen-address handling).
- [VictoriaMetrics/operator#1308](https://github.com/VictoriaMetrics/operator/issues/1308) — hardcoded config-reloader probes with proxy-protocol (probe port behavior; not hostNetwork / `ContainerPort` scheduling).

## Related code

- `internal/controller/operator/factory/build/container.go`: `ConfigReloaderDefaultPort`, `ConfigReloaderContainer`, `addPortProbesToConfigReloaderContainer`

cesium147:
Parse ConfigReloaderExtraArgs["http.listenAddr"] into ContainerPort, probes, named scrape targets, and WaitForConfigReloadHash so hostNetwork DaemonSets can use distinct reloader ports without scheduler conflicts.

Fixes #2584


vrutkovs:
Please don't merge this yet - it looks unnecessarily complicated and relies on static container name, which may be undesirable

internal/controller/operator/factory/build/container.go
```diff
@@ -326,19 +328,39 @@ func AppendArgsForInsertPorts(args []string, ip *vmv1beta1.InsertPorts) []string
...
-var configReloaderContainerProbe = corev1.ProbeHandler{
-	HTTPGet: &corev1.HTTPGetAction{
-		Path:   "/health",
-		Scheme: "HTTP",
-		Port:   intstr.FromInt(ConfigReloaderDefaultPort),
-	},
+// configReloaderHTTPPort returns the HTTP listen port from ConfigReloaderExtraArgs["http.listenAddr"]
+// (forms :PORT, IP:PORT, [IPv6]:PORT). Missing or invalid values fall back to ConfigReloaderDefaultPort.
+func configReloaderHTTPPort(extraArgs map[string]string) int32 {
+	addr, ok := extraArgs["http.listenAddr"]
```
AndrewChubatiuk: http.listenAddr -> httpListenAddr

internal/controller/operator/factory/reconcile/config_reload_test.go
```diff
@@ -50,13 +94,34 @@ func readyPod(name, namespace, ip string, labels map[string]string) *corev1.Pod
...
 
-func withConfigReloaderMetricsURL(t *testing.T, ts *httptest.Server) {
+type roundTripFunc func(*http.Request) (*http.Response, error)
+
+func (f roundTripFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	return f(r)
+}
+
+// withConfigReloaderMetricsResponse stubs the sidecar /metrics body in-process. httptest is
+// unreliable here when loopback is broken, so the wait tests drive an injected Transport.
```
vrutkovs: I'm not sure what "loopback is broken" means - could you provide an example? There are other places where we read metrics (i.e. vmdistributed) - do we need to make the same change there too?

I like the improvement in test UX, but the roundtrip change looks excessive

cesium147: You are right. The transport stub was local test-environment noise and does
not belong in this PR. Sorry for adding it.

I restored the existing httptest pattern and removed configReloadNewHTTPClient,
roundTripFunc, and the loopback comment. The remaining change only resolves
the configured reloader port from the PodSpec so WaitForConfigReloadHash uses
the same port as the container and probes.

PR #2589, opened 2026-09-05
vmagent, vmsingle, vmalert: fix config-reloader crash

### LINKED ISSUE #2583: BUG: vmalert config-reloader fails to start when spec.configMaps is set (operator v0.74.1)
## Describe the bug

The operator builds a `config-reloader` sidecar spec that its own reloader binary rejects.

When a `VMAlert` sets `spec.configMaps`, the sidecar crashes on start:

```
--target-dir count (1) must match watched directory count (2) from --watched-dir/--rules-dir
```

The pod never becomes ready. The rollout is stuck.

## To Reproduce

1. Run operator v0.74.1 with `config-reloader-v0.74.1`.
2. Create a ConfigMap `my-templates` with a `.tmpl` file.
3. Create a `VMAlert` with:

```yaml
spec:
  configMaps:
    - my-templates
  extraArgs:
    rule.templates: /etc/vm/configs/**/*.tmpl
```

4. The `config-reloader` container fails. Exit code 2.

Generated args:

```
--target-dir=/etc/vmalert/rules-out/rules-src-0
--watched-dir=/etc/vm/configs/my-templates
--watched-dir=/etc/vmalert/rules-src-0
```

Two watched dirs, one target dir.
## Root cause

Two separate problems, both in v0.74.1.

**1. Unpaired `--watched-dir`**

`internal/controller/operator/factory/build/container.go:418-422` adds a bare `--watched-dir` for every entry in `spec.configMaps`. No `--target-dir` goes with it.

`AddWatchTargetDir` at `container.go:525-536` does it correctly. Its own comment says it appends the pair "atomically".

`cmd/config-reloader/main.go:100` then calls `validateTargetDirCount`, which fails at `main.go:400-403`.

So the operator and its reloader disagree inside a single release.

**2. `sort.Strings` destroys the pairing**

`internal/controller/operator/factory/vmalert/vmalert.go:355` calls `sort.Strings(crc.Args)`.

The flags are matched by position. Sorting puts `/etc/vm/configs/...` before `/etc/vmalert/rules-src-0`. Even with a matching count, the templates would be written into the rules output directory.

The strict check added in reloader v0.73.0 is what stops this. On reloader v0.72.0 the flag exists but there is no validation, so the corruption would be silent.

## Why the init container is fine

`container.go:416` takes the `--only-init-config` branch. That branch does not emit the `crMounts` watched dirs at all. So only the sidecar is affected.

## Related

#1370 asked for `spec.configMaps` to be watched by the reloader. The bare `--watched-dir` comes from that. The validation added later in the reloader makes the two incompatible.

## Workaround

Move the ConfigMap out of `spec.configMaps` into `spec.volumes` plus `spec.volumeMounts` at the same path:

```yaml
spec:
  configMaps: []
  volumes:
    - name: my-templates
      configMap:
        name: my-templates
  volumeMounts:
    - name: my-templates
      mountPath: /etc/vm/configs/my-templates
      readOnly: true
```

The vmalert container still sees the files. The reloader gets no extra watched dir. Templates are static, so no reload is needed.

## Suggested fix

Use `AddWatchTargetDir` for `crMounts` too, or drop the bare `--watched-dir` loop. Also replace the positional pairing with something order independent, so `sort.Strings` cannot break it.

AndrewChubatiuk:
fixes #2583

adds empty watch-dir args for collected objects, if extra configuration, which requires config-reloader is defined


internal/controller/operator/factory/build/container.go
```diff
@@ -387,7 +387,19 @@ type reloadable interface {
...
-func ConfigReloaderContainer(isInit bool, cr reloadable, mounts []corev1.VolumeMount, ss *corev1.SecretKeySelector) corev1.Container {
+// ConfigReloaderOpts controls how ConfigReloaderContainer builds the reloader's container.
+type ConfigReloaderOpts struct {
+	CR                reloadable
+	Mounts            []corev1.VolumeMount
+	SecretKeySelector *corev1.SecretKeySelector
+	// IsInit builds the one-shot init-container variant (--only-init-config).
+	IsInit bool
+	// PadTargetDirs pairs every --watched-dir emitted for mounts with an empty --target-dir.
+	PadTargetDirs bool
```
Haleygo: Is `PadTargetDirs` field needed?
In `ConfigReloaderContainer()`, we already skip `mount` when `IsInit` is `true`. So if we remove the `PadTargetDirs` option and rely only on the `IsInit` check as:
```
	if opts.IsInit {
		args = append(args, "--only-init-config")
	} else {
		for _, m := range mounts {
			if m.Name != outVolumeName {
				args = append(args, fmt.Sprintf("--watched-dir=%s", m.MountPath), "--target-dir=")
			}
		}
	}
```
It will add additional empty `--target-dir=` pairs for `vmagent/vmsingle` volumes, but seems harmless, right?

docs/CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ aliases:
...
 * BUGFIX:  [vmanomaly](https://docs.victoriametrics.com/operator/resources/vmanomaly/): pass previously skipped spec.extraEnvsFrom to anomaly pods. See [#2567](https://github.com/VictoriaMetrics/operator/issues/2567).
+* BUGFIX: [vmagent](https://docs.victoriametrics.com/operator/resources/vmagent/), [vmsingle](https://docs.victoriametrics.com/operator/resources/vmsingle/), [vmalert](https://docs.victoriametrics.com/operator/resources/vmalert/): fix `config-reloader` sidecar crashing with `--target-dir count must match watched directory count` when `spec.configMaps` was combined with another feature that also makes the reloader watch a directory (extra scrape config secrets for `VMAgent`/`VMSingle`, or rule files for `VMAlert`). See [#2583](https://github.com/VictoriaMetrics/operator/issues/2583).
```
```suggestion
* BUGFIX: [vmagent](https://docs.victoriametrics.com/operator/resources/vmagent/), [vmsingle](https://docs.victoriametrics.com/operator/resources/vmsingle/), [vmalert](https://docs.victoriametrics.com/operator/resources/vmalert/): fix `config-reloader` sidecar crashing with `--target-dir count must match watched directory count` error. Previously, the bug could be triggered by `VMAlert` with `spec.configMaps` defined and at least one `VMRule` selected, or by `VMAgent` and `VMSingle` whose scrape config was large enough while also using `spec.configMaps`, relabeling configs or stream aggregation configs. This bug was introduced in [v0.72.0](https://github.com/VictoriaMetrics/operator/releases#release-v0.72.0). See [#2583](https://github.com/VictoriaMetrics/operator/issues/2583).
```

internal/controller/operator/factory/build/container.go
Haleygo: Is `PadTargetDirs` field needed?
In `ConfigReloaderContainer()`, we already skip `mount` when `IsInit` is `true`. So if we remove the `PadTargetDirs` option and rely only on the `IsInit` check as:
```
	if opts.IsInit {
		args = append(args, "--only-init-config")
	} else {
		for _, m := range mounts {
			if m.Name != outVolumeName {
				args = append(args, fmt.Sprintf("--watched-dir=%s", m.MountPath), "--target-dir=")
			}
		}
	}
```
It will add additional empty `--target-dir=` pairs for `vmagent/vmsingle` volumes, but seems harmless, right?

Haleygo:
Also, the comment is not true now:
https://github.com/VictoriaMetrics/operator/blob/37bf90e4037e4c1a9d293410bab5026cfe4266a9/internal/controller/operator/factory/vmalert/rules.go#L213

Now, with no groups, no configmap will be created too. 
It should be fine, only triggers vmalert deployment spec change for 0->1 group number transition, would be nice to fix the comment or have an empty configmap for 0 groups like previous if it doesn't complicate the code too much.

internal/controller/operator/factory/build/container.go
AndrewChubatiuk: right, but will rollout existing pods

AndrewChubatiuk: updated code

AndrewChubatiuk: updated

Haleygo:
@AndrewChubatiuk I polished the `--watched-dir/ --target-dir` argument generation logic a bit to avoid possible future sorting issue, please let me know if you see any issue.

It also starts ignoring the `watched-dir`, `target-dir`, and `rules-dir` args passed in `configReloaderExtraArgs`, as they could break the generated configs, and I don’t see a valid use case for users to add these via `configReloaderExtraArgs`.

Haleygo (approved):

AndrewChubatiuk:
@Haleygo 
most likely order of args in e2e should be updated as well

test/e2e/upgrade/utils.go
```diff
@@ -330,9 +331,37 @@ func getRolloutDiff(ctx context.Context, k8sClient client.Client, expectedPodSpe
...
 	return cmp.Diff(expectedPodSpec, got)
 }
 
+var configReloaderContainerNames = []string{"config-reloader", "config-init"}
+
+// normalizeConfigReloaderArgs makes config-reloader args comparable across operator versions.
+func normalizeConfigReloaderArgs(spec *corev1.PodSpec) {
+	if spec == nil {
+		return
+	}
```
AndrewChubatiuk: probably better is just remove previous versions from upgrade tests, at least for now we do like this, when rollout is expected

Haleygo: You mean just drop those resources from https://github.com/VictoriaMetrics/operator/blob/db6b96386f8519bc0b8463cb4b737f12de054f90/test/e2e/upgrade/upgrade_test.go#L677?

AndrewChubatiuk: yes, upgrade tests are needed to check that no rollout is expected, but if it's already documented there's nothing to test

PR #2599, opened 2026-09-10
feat(reconcile): propagate volumeAttributesClassName PVC updates

### LINKED ISSUE #2598: volumeAttributesClassName in volumeClaimTemplate is not applied to existing PVCs
### What happened

I set `volumeAttributesClassName` on `vlstorage.storage.volumeClaimTemplate.spec` in a VLCluster. The operator noticed the change and recreated the StatefulSet, and the new StatefulSet carries the field. The three existing PVCs were left as they were, so the volumes kept their original IOPS and throughput.

Operator v0.65.0, Kubernetes 1.36, aws-ebs-csi-driver v1.44.0.

### What I expected

`updateSTSPVC` already lists the live PVCs and reconciles them against the template, so I expected the class name to be carried over the way a size change is.

### Why it does not

`modifyPVC` in `internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go` returns early when the storage request is unchanged, and `Spec.Resources` is the only spec field it ever writes.

### Suggested change

Copy the field when it is set and differs:

```go
vacChanged := newObj.Spec.VolumeAttributesClassName != nil &&
    !ptr.Equal(existingObj.Spec.VolumeAttributesClassName, newObj.Spec.VolumeAttributesClassName)
```

and let that count towards the early return. The nil check is needed because clearing the field is rejected once a class has taken effect.

Users who do not set the field are unaffected, since both sides stay nil and the comparison is false.

I am happy to open a PR if you would take it.

vrutkovs:
Propagate `volumeClaimTemplate.spec.volumeAttributesClassName` changes to existing `PersistentVolumeClaim`s

Fixes #2598


internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
```diff
@@ -154,33 +154,41 @@ func modifyPVC(ctx context.Context, rclient client.Client, existingObj, newObj,
...
+
+	changed := metaChanged || vacChanged
+	if !changed && direction == 0 {
 		return false, nil
 	}
+	l := logger.WithContext(ctx)
+	if vacChanged {
+		l.Info(fmt.Sprintf("updating volumeAttributesClassName for pvc=%s from=%s to=%s",
+			newObj.Name, ptr.Deref(existingObj.Spec.VolumeAttributesClassName, "<none>"), ptr.Deref(newObj.Spec.VolumeAttributesClassName, "<none>")))
+		existingObj.Spec.VolumeAttributesClassName = newObj.Spec.VolumeAttributesClassName
```
DawerRafi, line 177: I believe we would need a `nil` check. Since the API forbids removing a VAC once set. 

ref: https://github.com/kubernetes/kubernetes/blob/8a1ebaa4eb27b6ba1cbfab00fd927cd7ea230606/pkg/apis/core/validation/validation.go#L2633

vrutkovs: Good idea, updated

internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
```diff
@@ -154,33 +154,47 @@ func modifyPVC(ctx context.Context, rclient client.Client, existingObj, newObj,
...
-	if !metaChanged && direction == 0 {
+
+	changed := metaChanged || vacChanged
+	if !changed && direction == 0 {
 		return false, nil
 	}
+	l := logger.WithContext(ctx)
+	if vacChanged {
+		// k8s forbids clearing volumeAttributesClassName once set, reject early
+		if existingObj.Spec.VolumeAttributesClassName != nil && newObj.Spec.VolumeAttributesClassName == nil {
```
Haleygo: From https://kubernetes.io/docs/reference/kubernetes-api/core/persistent-volume-claim-v1/, it only forbids removing VAC once one is successfully applied, but 
>If the claim enters an Infeasible error state, this field can be reset to its previous value (including nil) to cancel the modification.

So we should check on `existingObj.Status.CurrentVolumeAttributesClassName` instead `existingObj.Spec.VolumeAttributesClassName`.

vrutkovs: Updated the PR to implement that

internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
```diff
@@ -154,33 +154,49 @@ func modifyPVC(ctx context.Context, rclient client.Client, existingObj, newObj,
...
+	changed := metaChanged || vacChanged
+	if !changed && direction == 0 {
 		return false, nil
 	}
+	l := logger.WithContext(ctx)
+	if vacChanged {
+		// k8s permits reset only after CSI marks volume attribute change infeasible.
+		if existingObj.Status.CurrentVolumeAttributesClassName != nil &&
+			newObj.Spec.VolumeAttributesClassName == nil &&
+			(existingObj.Status.ModifyVolumeStatus == nil || existingObj.Status.ModifyVolumeStatus.Status != corev1.PersistentVolumeClaimModifyVolumeInfeasible) {
```
Haleygo: I don't think the checks around `ModifyVolumeStatus` are needed, I see the rule is just:
> You can't set `spec.volumeAttributesClassName` to `nil` while `status.currentVolumeAttributesClassName` is set.

And the `If the claim enters an `Infeasible` error state, this field can be reset to its previous value (including `nil`) to cancel the modification` case is already covered by the rule above.

So I'd keep the check simple, as the upstream does: https://github.com/kubernetes/kubernetes/blob/8a1ebaa4eb27b6ba1cbfab00fd927cd7ea230606/pkg/apis/core/validation/validation.go#L2634-L2635

Haleygo:
While checking the upstream check at https://github.com/kubernetes/kubernetes/blob/8a1ebaa4eb27b6ba1cbfab00fd927cd7ea230606/pkg/apis/core/validation/validation.go#L2634, it appears the status-based check was introduced in v1.34; prior to that, the validation was performed against the spec, see https://github.com/kubernetes/kubernetes/commit/f7109ed281a6a8538cd98c7369593c2737af76d3#diff-c713e8919642d873fdf48fe8fb6d43e5cb2f53fd601066ff53580ea655948f0d.
It would be better to add a Kubernetes version check to decide which behavior to forbid in this case.

Haleygo:
I pushed https://github.com/VictoriaMetrics/operator/pull/2599/changes/3b87774b42d476c7567bf300607568d7f3146445 to align volumeAttributesClassName removal with k8s validation, please let me know if you see any issue.
Also, this change looks more like a bugfix than a feature to me.

vrutkovs (commented):
Rebased and updated the test, LGTM

Haleygo (approved):

PR #2621, opened 2026-09-14
fixed dnsConfig in CRD

### LINKED ISSUE #2618: CRD schemas: dnsConfig nodes ship an invalid root-level items key (breaks KubeVela/CUE and other strict JSON Schema consumers)
## Summary

Every `dnsConfig` node in the released CRD manifests ships an **invalid JSON Schema**: an object-typed schema carrying a stray root-level `items` key:

```yaml
dnsConfig:
  description: Specifies the DNS parameters of a pod...
  items:                                  # <-- invalid: items on a type:object node
    x-kubernetes-preserve-unknown-fields: true
  properties:
    nameservers: {type: array, items: {type: string}, ...}
    options:     {type: array, items: {...}}
    searches:    {type: array, items: {type: string}}
  type: object
```

The Kubernetes apiserver tolerates the dead key, so the CRDs install fine — but any strict JSON Schema consumer breaks. Notably **KubeVela**, which parses all CRD schemas with CUE, fails hard:

```
E0914 06:22:47 traitdefinition_controller.go:80] "Could not refresh packageDiscover"
err="parse json schema of k8s crds error: constraint not allowed because type array is excluded (and 16 more errors)"
```

`items` only applies to arrays; on a node where arrays are excluded (`type: object`) it is contradictory, hence the CUE error. This kills `packageDiscover` for **all** traitdefinitions on the cluster, not just VM-related ones.

## Root cause

The bogus `items` is not generated — it is added by the release-time kustomize patches in `config/crd/patches/`. In `operator.victoriametrics.com_vlagents.yaml` (and per-CRD variants), every *array* field is loosened correctly by patching **inside** `items`, e.g.:

```yaml
- op: add
  path: /.../properties/containers/items/x-kubernetes-preserve-unknown-fields
  value: true
```

but the dnsConfig op treats the object field like an array and adds `items` **at its root**:

```yaml
- op: add
  path: /.../properties/dnsConfig/items     # <-- should not exist; dnsConfig is type: object
  value:
    x-kubernetes-preserve-unknown-fields: true
```

After the #2489 refactor the op survives in `config/crd/patches/common-array-fields-with-claimtemplates.yaml` (line 10) plus 3 more in `operator.victoriametrics.com_vlclusters.yaml`, and is still on current master.

Evidence that controller-gen output itself is clean: the **embedded** `spec.dnsConfig` inside `vmdistributed`/`vldistributed` (whose paths are not covered by these patches) has always shipped as a valid `{type: object, properties: {...}}` in every release. Only the patch-covered standalone CRDs are affected — and the bogus value `{x-kubernetes-preserve-unknown-fields: true}` matches the patch value verbatim.

## Affected versions

Checked release assets (`crd.yaml`) — the malformed nodes are present in **every** release from the dnsConfig feature introduction (PR #461, first shipped ~v0.31.0, 2022-04) through **v0.75.0-rc1** and current master:

| release | affected dnsConfig nodes |
|---|---|
| v0.31.0 – v0.39.4 | 8 |
| v0.69.0 | 14 |
| v0.70.0 – v0.73.1 | 18 |
| v0.74.x (full CRD set) | 18 |
| v0.75.0-rc1 | still present (schema now also carries root x-kubernetes-preserve-unknown-fields, but the stray items remains) |

The helm chart `victoria-metrics-k8s-stack` (e.g. 0.77.0, operator subchart crd.yaml.bz2 applied via its pre-install hook job) ships the same affected CRD set, so chart users get it too.

## Workaround

Remove the dead key after every install/upgrade (idempotent):

```bash
base="/spec/versions/0/schema/openAPIV3Schema/properties/spec/properties"
for entry in \
  "vlagents:$base/dnsConfig/items" \
  "vlclusters:$base/vlinsert/properties/dnsConfig/items" \
  "vlclusters:$base/vlselect/properties/dnsConfig/items" \
  "vlclusters:$base/vlstorage/properties/dnsConfig/items" \
  "vlsingles:$base/dnsConfig/items" \
  "vmagents:$base/dnsConfig/items" \
  "vmalertmanagers:$base/dnsConfig/items" \
  "vmalerts:$base/dnsConfig/items" \
  "vmanomalies:$base/dnsConfig/items" \
  "vmauths:$base/dnsConfig/items" \
  "vmclusters:$base/vminsert/properties/dnsConfig/items" \
  "vmclusters:$base/vmselect/properties/dnsConfig/items" \
  "vmclusters:$base/vmstorage/properties/dnsConfig/items" \
  "vmsingles:$base/dnsConfig/items" \
  "vtclusters:$base/insert/properties/dnsConfig/items" \
  "vtclusters:$base/select/properties/dnsConfig/items" \
  "vtclusters:$base/storage/properties/dnsConfig/items" \
  "vtsingles:$base/dnsConfig/items" ; do
  crd="${entry%%:*}.operator.victoriametrics.com"; path="${entry#*:}"
  kubectl patch crd "$crd" --type=json \
    -p "[{\"op\":\"remove\",\"path\":\"$path\"}]" >/dev/null 2>&1 \
    && echo "patched $crd" || echo "skip $crd (absent or already patched)"
done
```

Removing the root-level `items` is safe: the real per-field schemas (`nameservers`/`options`/`searches`, each with proper `items`) stay in place, and the apiserver ignores the dead key anyway.

## Suggested fix

Simply drop the bogus ops (`.../dnsConfig/items`) — the fully generated `properties` are valid and complete, so dnsConfig needs no loosening. Alternatively, if loosening dnsConfig is actually intended, apply it the way the other object fields do:

```yaml
- op: add
  path: /.../properties/dnsConfig/x-kubernetes-preserve-unknown-fields
  value: true
- op: remove
  path: /.../properties/dnsConfig/properties
```

Happy to turn this into a PR if the maintainers prefer — just let me know which direction (drop vs. proper preserve-unknown) is wanted.

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/2618


docs/CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ aliases:
...
 * BUGFIX:  [vmanomaly](https://docs.victoriametrics.com/operator/resources/vmanomaly/): fix propagation of `spec.extraEnvsFrom` to anomaly pods, previously it was omitted. See [#2567](https://github.com/VictoriaMetrics/operator/issues/2567).
 * BUGFIX:  [vmalertmanager](https://docs.victoriametrics.com/operator/resources/vmalertmanager/): fix propagation of `spec.extraEnvsFrom` to vmalertmanager pod, previously it was omitted. See [#2582](https://github.com/VictoriaMetrics/operator/issues/2582).
+* BUGFIX: [vmoperator](https://docs.victoriametrics.com/operator/): remove an invalid root-level `items` key from every CRD's `dnsConfig` schema node. Kubernetes ignored it, but strict JSON Schema consumers (e.g. KubeVela) rejected the whole schema. See [#2618](https://github.com/VictoriaMetrics/operator/issues/2618).
```
```suggestion
* BUGFIX: [vmoperator](https://docs.victoriametrics.com/operator/): remove the invalid `items` key under `dnsConfig` in the `VMAgent`, `VMAlertmanager`, `VMAnomaly`, `VMCluster`, `VLAgent`, `VLCluster`, `VTAgent` and `VTCluster` CRDs. Previously, while Kubernetes itself ignored this key, it could fail validation in strict JSON Schema consumers such as KubeVela. See [#2618](https://github.com/VictoriaMetrics/operator/issues/2618).
```

Haleygo (approved):

PR #2623, opened 2026-09-15
vmrule: removed default annotations

### LINKED ISSUE #2622: BUG: VMRule CRD injects empty `record: ""` / `alert: ""` into every rule
### Describe the bug
`Rule.Record` and `Rule.Alert` in `api/operator/v1beta1/vmrule_types.go` carry `+kubebuilder:default=""`. 

The API server injects the empty counterpart field into every stored rule: alerting rules get `record: ""`, recording rules get `alert: ""`.

The applied manifest never contains these fields, so the live object never matches the source of truth. Diff-based GitOps tooling (Argo CD and equivalents) reports every application owning a VMRule as permanently OutOfSync.

### Why the defaults are no longer needed
- #2258 added `+listType=map` with `+listMapKey=record` / `+listMapKey=alert` to `RuleGroup.Rules`, plus `+kubebuilder:default=""` on both fields. The defaults were required only because the API server rejects a list map key that has neither a default nor a required marker.
- #2379 removed the list-map markers from `Rules` again. The defaults were left behind.

With the list-map keys gone, nothing requires the defaults — they are residue, and their only remaining observable effect is the spurious diff.

### Expected behavior
`record` and `alert` are `omitempty` optional fields. An applied VMRule should round-trip unchanged: no field the user did not set should appear on the stored object.

### Version
First affected release is **v0.73.0** (the markers are absent in v0.72.0) in v0.73.1, v0.74.0, v0.74.1 and v0.75.0-rc1, and on master.

### Proposed fix
Drop `+kubebuilder:default=""` from `Rule.Record` and `Rule.Alert`, regen bundle. The doc comment on `RuleGroup.Rules` ("Rules are merged by record fields") is also stale after #2379 and can go in the same change.

### Current workaround
Strip the defaults from the CRD bundle before applying, or add an `ignoreDifferences` entry in Argo CD for the field.

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/operator/issues/2622


docs/CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ aliases:
...
 * BUGFIX:  [vmalertmanager](https://docs.victoriametrics.com/operator/resources/vmalertmanager/): fix propagation of `spec.extraEnvsFrom` to vmalertmanager pod, previously it was omitted. See [#2582](https://github.com/VictoriaMetrics/operator/issues/2582).
 * BUGFIX: [vmoperator](https://docs.victoriametrics.com/operator/): fix HTTPRoute `extraRules` inheriting fields from previous rules. See [#2605](https://github.com/VictoriaMetrics/operator/issues/2605).
+* BUGFIX: [vmrule](https://docs.victoriametrics.com/operator/resources/vmrule/): drop a leftover `+kubebuilder:default=""` on `record`/`alert`, which made the apiserver inject an empty counterpart field into every stored `VMRule`, causing diff-based GitOps tooling (e.g. Argo CD) to report it as permanently `OutOfSync`. The default was needed only for a `+listMapKey` requirement added in [#2258](https://github.com/VictoriaMetrics/operator/pull/2258) and removed in [#2379](https://github.com/VictoriaMetrics/operator/pull/2379). See [#2622](https://github.com/VictoriaMetrics/operator/issues/2622).
```
```suggestion
* BUGFIX: [VMRule CRD](https://docs.victoriametrics.com/operator/api/#v1beta1-vmrule): drop the leftover default value `""` from the `record` and `alert` fields. This default caused Kubernetes to add an empty `alert: ""` or `record: ""` to every rule, so GitOps tools such as ArgoCD always detected a diff and kept `VMRule` objects `OutOfSync`. This default value was introduced in [v0.73.0](https://github.com/VictoriaMetrics/operator/releases#release-v0.73.0), which required every rule to have both `record` and `alert` fields set. [v0.73.1](https://github.com/VictoriaMetrics/operator/releases/tag/v0.73.1) stopped identifying rules this way, since a group may contain multiple rules with the same name. The `VMRule` CRD must be updated to get this fix. See [#2622](https://github.com/VictoriaMetrics/operator/issues/2622).
```

Haleygo (approved):

PR #2625, opened 2026-09-15
operator: report the reason when a pvc resize is declined

### LINKED ISSUE #2512: VMAgent reports `updateStatus: operational` when a requested PVC decrease was not applied

Haleygo:
Fix https://github.com/VictoriaMetrics/operator/issues/2512.
Follow-up to https://github.com/VictoriaMetrics/operator/pull/2541:
1. The previous change only handled the size decrease case. A declined expansion on a non-expandable `StorageClass` still surfaced as `context deadline exceeded`, since the resize was skipped while the `waitForPVCReady` check kept waiting until it timed out;
2. Follow the "don't return an error to the caller, since there is no point in requeueing and reconciling this when the `StorageClass` is non-expandable" rule for decline cases, report the decline reason under `status.reason` and emit a Kubernetes event, but do not return error which leads to direct requeue;
3. Do not skip metadata updates when a resize is declined.


internal/controller/operator/controllers.go
```diff
@@ -84,6 +84,7 @@ const (
 	reasonCancelContext = "cancel_context"
 	reasonConflict      = "conflict"
 	reasonOther         = "other"
+	reasonDeclined      = "declined"
```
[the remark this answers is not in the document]
Haleygo: Fixed.

internal/controller/operator/controllers.go
```diff
@@ -245,6 +248,11 @@ func handleReconcileErr(ctx context.Context, rclient client.Client, object clien
 	if err := rclient.Create(ctx, errEvent); err != nil {
 		logger.WithContext(ctx).Error(err, "failed to create error event at kubernetes API during reconciliation error")
 	}
+	if reconcile.IsDeclined(err) {
```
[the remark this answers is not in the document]
Haleygo: Fixed.

internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
```diff
@@ -178,43 +182,60 @@ func modifyPVC(ctx context.Context, rclient client.Client, existingObj, newObj,
...
 		if !expandable {
-			// don't return error to caller, since there is no point to requeue and reconcile this when sc is unexpandable
-			sc := ptr.Deref(newObj.Spec.StorageClassName, "default")
-			l.Info(fmt.Sprintf("storage class=%s for PVC=%s doesn't support live resizing", sc, newObj.Name))
-			return metaChanged, nil
+			// report the StorageClass isStorageClassExpandable actually checked
+			sc := pvcStorageClassName(existingObj)
+			if sc == "" {
+				sc = "default"
+			}
```
AndrewChubatiuk, line 205: wdyt about putting this inside pvcStorageClassName?

internal/controller/operator/factory/reconcile/statefulset_pvc_expand_test.go
```diff
@@ -152,12 +152,13 @@ func Test_recreateOrUpdateSTS(t *testing.T) {
...
 	type opts struct {
-		sts      *appsv1.StatefulSet
-		prevVCTs []corev1.PersistentVolumeClaim
-		wantErr  bool
-		preRun   func(c client.Client)
-		expected []corev1.PersistentVolumeClaim
-		actions  []k8stools.ClientAction
+		sts         *appsv1.StatefulSet
+		prevVCTs    []corev1.PersistentVolumeClaim
+		wantErr     bool
```
AndrewChubatiuk: let's get rid of bool property

internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
Haleygo: The `default` here is used to render the error below, but the default StorageClass name is not actually `default` in many cases. For other caller, the name should not be set to `default`.

https://github.com/VictoriaMetrics/operator/blob/359e87c4cfa440b4c033f0cf329565a60afb8c25/internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go#L255-L265

internal/controller/operator/factory/reconcile/statefulset_pvc_expand_test.go
Haleygo: Removed.

internal/controller/operator/factory/reconcile/statefulset_pvc_expand.go
AndrewChubatiuk: `if className != DefaultStorageClassName` instead of `if len(className) > 0` should be enough

AndrewChubatiuk (approved):

